PR #48, opened 2026-05-23
feat(video): wire i2v/r2v + model/duration/count picker; fix image --model no-op

svasakorn:
## Summary

Wires the `ui_automation` video + image controls from the Phase B roadmap. Every path drives the live Flow editor and was verified end-to-end against a Pro/Ultra account; the captured generate route is noted per feature.

**Video (`gflow video …`)**

- `t2v`: `--model` picker (`omni-flash` | `veo-lite` | `veo-fast` | `veo-quality` | `veo-lite-lp`), `--duration` (`4`/`6`/`8`, plus `10` for `omni-flash` only), `--count` (1–4). → `batchAsyncGenerateVideoText`
- `i2v`: start frame + optional `--end-image`. → `batchAsyncGenerateVideoStartImage` / `…StartAndEndImage`
- `r2v` (new): reference-to-video, repeatable `--ref`. Model-aware reference cap (omni_flash ≤7, veo_3_1_* ≤3). → `batchAsyncGenerateVideoReferenceImages`

**Image fix**

- `gflow image t2i/i2i --model` now actually selects the model. It was a no-op under `ui_automation`: the wire field was set but the model picker was never clicked, so Flow used its UI default.

**Selector fixes**

- The output-count selector `[id*=-trigger-1]` collided with the `-trigger-10` duration tab; the aspect selector matched a non-existent `aria-controls*=9_16`; the video-mode tab match was ambiguous. All now use exact `[id$=-trigger-X]` suffixes + aria-label text.

## Key approach notes (would value your opinion)

- **Image inputs bind through the editor's media dialog** (frame slot / add-media → "Upload media" → file chooser → "Add to Prompt"). `set_input_files` on the generic hidden `<input type=file>` only adds the image to the *library* — Flow then ignores it and fires the plain Text route. Binding through the dialog is what makes I2V/R2V actually condition on the image.
- **The editor is forced to English** via a `--lang=en-US` Chromium launch arg. The frame-slot and dialog labels are localized (e.g. TH "เริ่ม"/"สิ้นสุด") with no locale-free anchor, and `locale="en-US"` alone only sets `Accept-Language` while Flow still serves `/fx/<locale>/`. Open to a locale-agnostic alternative if you prefer.
- A small request-body diagnostic logs `referenceCount` + mediaId prefixes (never tokens/credits) so the binding is provable from the structured log.

## Test plan

- [x] `uv run pytest -q` — 449 api tests pass (new video DTO / transport / selector tests included)
[...]

Issue #15, opened 2026-05-16
fix(i2v): uploadImage 401s with valid session — hybrid transport auth mismatch


ffroliva:
## Summary

`gflow video i2v <image> <prompt>` consistently returns `AuthExpiredError` (HTTP 401 on `POST /v1/flow/uploadImage`), even with a freshly verified session that the same CLI invocation captured seconds earlier. `t2i` and batch image generation work fine on the same session in the same window.

## Reproduction

1. Fresh login: `gflow auth login --profile denon82` (auto-picked `chrome` strategy on Windows with Chrome installed; verified success via `auth_login_success_verified` event).
2. Generate a source image: `gflow image t2i "an airship" -n 1 --out tmp/e2e_single --profile denon82` → exit 0, PNG saved.
3. Try i2v on that image: `gflow video i2v tmp/e2e_single/<uuid>_1.png "the airship sails through clouds" --profile denon82 -o tmp/e2e_video` → exit 3, `AuthExpiredError`.

Observed twice — once after the original batch run, again 18s after a verified fresh login.

Example Problem Details (correlation_id `c0a6d4e3-8e4c-4d43-a710-9da7a01ff018`):

```json
{
  "error_class": "AuthExpiredError",
  "problem": {
    "type": "https://gflow-cli.dev/errors/auth-expired",
    "status": 401,
    "detail": "HTTP 401",
    "route": "https://aisandbox-pa.googleapis.com/v1/flow/uploadImage",
    "remediation_hint": "Run `gflow auth login --profile <name>` to refresh the session."
  },
  "cli_command": "video i2v",
  "cli_version": "0.6.0a5"
}
```

## Diagnosis (hypothesis)

i2v has a **hybrid transport architecture** the bug fix in 1621748d didn't touch:

| Step | Transport | Notes |
|---|---|---|
| t2i image generation | `UiAutomationTransport` (UI clicks in Chromium) | works |
| i2v image **upload** | `FlowApiClient.upload_image` → `_post_json` to `/v1/flow/uploadImage` | 401 |
| i2v video generation | (not reached) | unknown |

Google's REST endpoints under `aisandbox-pa.googleapis.com` typically require more than just session cookies — almost certainly a **`SAPISIDHASH` Authorization header** computed from the `SAPISID` cookie. Browser-driven UI calls get this for free (Chromium's network stack computes & attaches it). A direct `page.request.post(...)` may or may not, depending on context origin and how the request is made.

[...]

PR #50, opened 2026-05-24
feat(image): i2i local-ref attach via media dialog (closes #39) + __aenter__ teardown guard

### LINKED ISSUE #39: [BUG] image i2i fails with AuthExpiredError: HTTP 401 (missing UI-automation for uploads)
# [BUG/FEATURE] Robust `i2v` and `i2i` support via "API-via-Browser" Bypass

**Description:**
Character consistency in `v0.7.0` is currently blocked because `i2v` and `i2i` require an image upload. The current transport logic falls back to a legacy HTTP POST which fails with `401 Unauthorized`. 

Empirical research using `browser-harness` has identified a highly robust "API-via-Browser" bypass that eliminates the need for complex UI clicking and solves the auth issue.

**Technical Discovery:**
- **Hidden Gateway:** Google Flow maintains a persistent hidden file input: `input[type="file"][accept="image/*"]`.
- **Direct Upload:** Verified that calling `page.set_input_files(selector, path)` on this input automatically triggers the `uploadImage` XHR request without any button clicking.
- **UUID Extraction:** The image UUID (needed for generation) is returned in the JSON response at `media.name`.
- **Auth Solution:** By using `page.evaluate(fetch, ...)` inside the browser session, we can reuse the application's internal `Authorization` (Bearer) tokens and session state.

**Proposed Implementation (Task 0 Fix):**

1. **Implement `upload_image` in `ui_automation.py`:**
   - Use `page.set_input_files('input[type="file"][accept="image/*"]', path)`.
   - Register a `page.on("response")` listener to capture the result from `https://aisandbox-pa.googleapis.com/v1/flow/uploadImage`.
   - Return the extracted `media[0].name`.

2. **Implement `i2v` (Image-to-Video) in `ui_automation_video.py`:**
   - Instead of navigating dropdowns, capture the `Authorization` header from the upload request.
   - Use `page.evaluate()` to call the `batchAsyncGenerateVideoStartAndEndImage` endpoint using `fetch()`.
   - This ensures all cookies and tokens are valid, bypassing the 401 error.

**Impact:**
- Restores character consistency for automated pipelines (Stickman/Kitten archetypes).
- significantly faster than UI-automation based clicking.
- Robust against UI layout changes.

**Verification:**
- Verified manually using a standalone script that successfully triggered the upload and captured the UUID.
- Requires integration into the `UiAutomationTransport` protocol.

svasakorn:
## Summary

Two focused follow-ups on top of #48 (now merged):

- **`gflow image i2i --ref <local-file>`** now binds the reference through the editor's media dialog — the same `add_2` dialog + the inherited `_attach_references` that #48 added for R2V — instead of the REST `uploadImage` endpoint, which 401s. **Closes #39** (same root as #15). Bare-UUID `--ref` still flows through `refs` unchanged; the CLI splits local paths (UI-attached) from pre-uploaded UUIDs.
- **`FlowApiClient.__aenter__` teardown guard** — if any step after the Playwright driver starts raises (persistent-context launch, bootstrap nav, or `transport.setup`), the browser is now torn down. Python does not call `__aexit__` when `__aenter__` raises, so an unguarded failure orphaned the chrome process; the orphan held the profile's user-data-dir lock and the next run spiralled into rapid `about:blank` tabs + `TargetClosedError`. Context-close + driver-stop are now shared by the guard and `__aexit__` via `_close_browser_resources`.

## Test plan

- [x] `gflow image i2i --ref <png>` live-verified via the real CLI: `uploadImage` 200, `imageInputs[]` populated, image returned (single ref and `--count 2`, fresh session).
- [x] CLI-layer tests (`test_cli_image.py`: local paths → `ref_paths`, UUIDs → `refs`, no REST upload) + transport-layer test (`test_ui_automation.py`: `_generate_images_locked` awaits `_attach_references` with the paths when `ref_paths` is set, and not for plain t2i).
- [x] Regression test for the teardown guard (`test_concurrency.py`: `__aenter__` failure still closes context + stops the driver).
- [x] `uv run ruff check src tests` + `uv run ruff format --check src tests` + `uv run pyright src` clean; full `uv run pytest -m "not e2e and not live"` green (806 passed).
- [x] Commits DCO-signed.

## Notes

Built on the `_attach_references` helper merged in #48. While here I noticed 3 pre-existing `pyright` errors in `tests/` (`test_concurrency.py`, `test_ui_automation.py` helper return types) — they aren't CI-gated (the workflow runs `pyright src`, not `tests`), so I left them out to keep this PR focused. Happy to send a separate `fix(tests)` cleanup if you'd like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `FlowApiClient.__aenter__` now tears down a partially-launched browser if any
+  step after the Playwright driver starts raises (e.g. the persistent-context
+  launch, the bootstrap navigation, or `transport.setup`). Python does not call
+  `__aexit__` when `__aenter__` raises, so an unguarded failure orphaned the
+  chrome process, which then held the profile's user-data-dir lock — the next
+  run could not acquire it and spiralled into rapid `about:blank` tabs +
+  `TargetClosedError`. Context close + driver stop are now shared by
+  `__aenter__`'s guard and `__aexit__` via `_close_browser_resources`.
 - `gflow image i2i --ref <local-file>` now binds the reference through the
   editor's media dialog instead of the REST `uploadImage` endpoint (which 401s —
   same root as #15/#39). Local-path refs ride a new `GenerateImageRequest.ref_paths`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `gflow image i2i --ref <local-file>` now binds the reference through the
+  editor's media dialog instead of the REST `uploadImage` endpoint (which 401s —
+  same root as #15/#39). Local-path refs ride a new `GenerateImageRequest.ref_paths`
+  field and are attached via the inherited R2V `_attach_references` (the image-mode
+  add-media dialog is the same `add_2` surface). Live-verified end-to-end: upload
+  200, `imageInputs[]` populated, image returned. Bare-UUID `--ref` still flows
+  through `refs` unchanged.
 - `gflow image t2i/i2i --model` now actually selects the requested model. It was
   a no-op under `ui_automation` (the wire field was set but the model picker was
   never clicked, so Flow used its UI default). Adds `_select_image_model`.
--- a/src/gflow_cli/api/client.py
+++ b/src/gflow_cli/api/client.py
@@ -166,6 +166,26 @@ class FlowApiClient:
         # opening a second Playwright process against the same profile dir
         # (which would conflict on the Chromium lockfile — spec § 5.4.4).
         self._pw = await async_playwright().start()
+        # Partial-setup leak guard: __aexit__ is NOT invoked when __aenter__
+        # raises, so a launched persistent context (and its chrome process)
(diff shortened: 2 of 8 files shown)


ffroliva:
Thanks for this @svasakorn — really appreciate the careful pairing of the i2i local-ref fix with the `__aenter__` teardown guard. The reuse of the R2V `_attach_references` helper from #48 is exactly the right call, and the partial-setup leak guard catches a class of "next run spirals into about:blank" failures that have bitten us before.

Merging now into `develop`. I'll run the live e2e on my end for this one (`gflow image i2i --ref <local-file>`, single + `--count 2`) and capture the verification ledger.

One small ask for future PRs: when you ship a credit-spending pathway, please add an e2e test alongside the unit tests — there's a pattern to follow in `tests/e2e/test_transports_e2e.py` (Criterion C2 covers t2i; an i2i variant would slot in next to it, gated by `GFLOW_CLI_E2E_PROFILE` like the others). That way the next person touching the editor flow has an automated tripwire instead of relying on manual verification. Happy to pair on the scaffolding if useful.

Thanks again!

svasakorn:
Thank you — and noted on the e2e coverage. Agreed that's the right automated tripwire for credit-spending paths; on future PRs that touch a generate flow I'll add an e2e test alongside the unit tests, following the `tests/e2e/test_transports_e2e.py` C2 pattern (an i2i variant next to the t2i one, gated by `GFLOW_CLI_E2E_PROFILE` like the others). Appreciate you running the live verification on your end, and the offer to pair on the scaffolding. Thanks again!

Issue #56, opened 2026-05-24
i2i ref attach (#50) fails with TimeoutError on ffroliva account; image-mode add-media flow likely differs from video R2V


ffroliva:
## Summary

`gflow image i2i --ref <local-file>` fails with an unhandled `TimeoutError` on the `ffroliva` Chrome-strategy profile (Pro account, en-US locale) post-merge of #50. The ref-attach path inside `_attach_references` clicks `Add Media` successfully but no `filechooser` event ever fires.

#50 (now reverted in a follow-up PR) was live-verified by @svasakorn on their account and CI was green for the unit tests. So this looks account/UI-variant-specific, not a regression in test coverage.

## Environment

- Profile: `ffroliva`, Chrome-strategy, en-US locale, logged into Flow (Pro)
- Branch at time of failure: `develop` @ `32c2331` (#50 merged)
- CLI: `gflow image i2i "make it cinematic with golden hour lighting" --ref test_assets/image_00.png --profile ffroliva --count 1`
- Reference image: local PNG, exists, valid
- correlation_id: `cf4ae6a4-ff5a-4f41-9c82-0d2403ec4d6f`

## Reproduce

```sh
.venv/Scripts/python.exe -m gflow_cli image i2i \
  "make it cinematic with golden hour lighting" \
  --ref test_assets/image_00.png \
  --profile ffroliva \
  --count 1
```

Project created (`477f1f4d-294a-4ba8-a9a6-3fc8e15b523e`), editor entered, image mode selected, count set, then 34s of silence, then `TimeoutError`.

## Structlog timeline (up to failure)

```
09:24:23.5Z  ui_automation.entered_editor
09:24:24.5Z  ui_automation_video.selector_matched probe=mode_switch_trigger
09:24:25.4Z  ui_automation_video.selector_matched probe=image_mode_tab
09:24:26.9Z  ui_automation.image_mode_entered
09:24:28.9Z  ui_automation.aspect_ratio_set value=9:16
09:24:29.0Z  ui_automation.count_setter_completed desired=1 success=true
(silent 34s gap — no further events)
09:25:02.8Z  error_unhandled exception_class=TimeoutError stack_hash=e6172106...
```

## Root-cause hypothesis

The failing call site is `_attach_references` (added in #48 for R2V at `src/gflow_cli/api/transports/ui_automation_video.py:674`, reused by #50 at `src/gflow_cli/api/transports/ui_automation.py:1510`). The helper sequence is:

[...]

Issue #24, opened 2026-05-19
Make UiAutomationTransport selectors language-agnostic (structural-first)


ffroliva:
## Problem

Google Flow renders its UI in the **account's language**, not the requested
locale. The video-generation design spike confirmed Flow ignores
`locale="en-US"` + `?hl=en` — the test account's Flow renders in Portuguese
(`Vídeo`, `Imagem`, `Criar`, ...). See
`docs/superpowers/specs/2026-05-18-ui-automation-video-generation-design.md`
§10.4.

Several `UiAutomationTransport` selectors (the working **t2i /
`generate_images`** path) match on **localized button text**, so they only
work for accounts whose Flow language happens to be in our hard-coded list.

## Current state — `src/gflow_cli/api/transports/ui_automation.py`

**Language-agnostic already (good — keep):**

- `PROMPT_INPUT_SELECTORS` — leads with the `data-slate-editor` structural selector.
- `SUBMIT_BUTTON_SELECTORS` — leads with the `arrow_forward` icon ligature.
- `GEN_SETTINGS_BUTTON_SELECTORS` — all `crop_*` icon ligatures.
- Aspect / count tabs — `_ASPECT_TAB` (numeric ratios `16:9`, `9:16`, ...) and
  `_COUNT_TAB` (`1x`, `x2`, ...). Numbers/symbols are not translated.

**Fragile:**

- `ONBOARDING_SELECTORS` (~line 158) — **purely localized button text**
  (`Agree` / `Aceitar` / `I agree` / `Concordo` / `Accept` /
  `Create with Flow` / `Criar com o Flow` / `Get Started` / `Começar`), with
  **no structural anchor at all**. An account whose Flow renders in an unlisted
  language (German, Japanese, ...) cannot pass the onboarding gate.
- `NEW_PROJECT_SELECTORS` (~line 142) and the `SUBMIT_BUTTON_SELECTORS` tail —
  localized-text fallbacks (`Novo projeto` / `Nuevo proyecto` /
  `Nouveau projet`; `aria-label*="Create"`). These work today only because the
  icon selector leads — maintenance debt and a silent-failure risk.

## Why it works today

t2i is **not broken** — the icon-ligature selectors lead and match, and the
test accounts happen to be in covered languages. The risk is silent: onboarding
fails only for an uncovered language, with no obvious error.

## Proposed fix (structural-first)

Flow is built on **Radix UI** + **Material Symbols**. Both expose developer
strings that are never translated:

[...]

PR #60, opened 2026-05-24
fix(image): re-introduce i2i ref-attach (#50) with locale-agnostic media-dialog selectors (#56)

svasakorn:
## Summary

Re-introduces the i2i local-file ref-attach from #50 (reverted in #57) and fixes the #56 root cause by making the media-dialog upload selectors **locale-agnostic**. Branched off `develop` @ `5345c1b` as you suggested.

## Root cause (#56)

It turned out to be a **locale leak**, not an intermediate-menu / nesting difference. Your `ffroliva` dump ([#issuecomment-4528075496](https://github.com/ffroliva/gflow-cli/issues/56#issuecomment-4528075496)) shows the **same** popover with the **same** "Upload media" button at the **same** level as mine — just rendered in pt-BR (`Enviar mídia`). The old `UPLOAD_MEDIA_BUTTON = button:has-text('Upload media')` didn't match the pt-BR label, so the click missed → `expect_file_chooser` never fired → the ~34s silent timeout.

Flow renders the dialog in the **Chrome _profile_ language**, which `--lang=en-US` cannot override on an already-configured profile (matches your "server-side locale overrides the launch arg" observation; same class as #51 / #24).

## Changes

`src/gflow_cli/api/transports/ui_automation_video.py`:

- **`UPLOAD_MEDIA_BUTTON`** → the locale-free `upload` icon ligature, `:text-is('upload')` (exact — so it doesn't also grab the `Uploads` tab whose icon is `drive_folder_upload`), scoped to `[role='dialog'][data-state='open']`. This is the selector you recommended, with the exact-match refinement.
- **Fallback tier** — the original English-text selector (`has-text('Upload media')`), tried only if the icon misses. Covers the narrow case where Google changes the icon ligature but keeps the English label. Short per-tier timeouts so two misses can't recreate a long silent hang.
- **'Add to Prompt'** has **no icon** (confirmed via DOM), so it's selected structurally as the lone iconless button in the open dialog (`.filter(has_not=<i.google-symbols>)`) instead of by localized text.
- **Fail-loud guard** — if neither tier opens a chooser, screenshot + a clear `RuntimeError`. The message explicitly tells the operator to set the **Chrome _profile_ language** to English (`chrome://settings/languages`) and that the **Google _account_ language is not enough** — that distinction is genuinely easy to get wrong.

`tests/e2e/test_transports_e2e.py`:
- Adds the C2-style **i2i e2e variant** (`test_e2e_i2i_local_ref_attach`), `GFLOW_CLI_E2E_PROFILE`-gated, ui-automation transport only (the REST transports can't drive the add-media dialog). Generates a tiny PNG inline — no asset dependency.

The shared `_upload_via_open_dialog` is also used by I2V/R2V, so this hardens those upload paths too.

## Verification

Real CLI, English-rendering profile, this branch:

```
11:22:55.614  image_mode_entered
11:23:00.372  count_setter_completed
11:23:13.986  image_uploaded  status=200   <- upload fired (~13.6s, not the 34s hang)
11:23:15.670  reference_attached
11:23:16.318  prompt_submitted
11:23:50.935  batch_response_captured status=200  -> image returned
```

**Honest caveat:** I can only verify on an English-rendering profile, where *both* the icon and text selectors would match — so this proves the end-to-end path works, but not the pt-BR path specifically. The icon selector is locale-independent by construction and your `ffroliva` dump confirms the same `upload` ligature there. Could you run the gated e2e (or your `capture_image_add_media_dom.py` from #59) on `ffroliva` to confirm before closing #56? I deliberately wrote "Refs", not "Closes".

## Not fixed here (focused PR — flagging per your preference)

`FRAME_SLOT_BY_LABEL` (I2V Start/End slots, `div[aria-haspopup='dialog']:has-text('{label}')`) is still **text-based** and will hit the same locale issue on a non-English profile. It's the I2V frame-slot path, separate from this upload fix, and belongs with #24 — happy to follow up separately rather than widen this PR.

## Test plan

- [x] `check_repo_hygiene.py` — clean (255 files)
- [x] `ruff check src tests` — clean
- [x] `ruff format --check src tests` — clean (103 files)
- [x] `pytest -q -m "not e2e and not live"` — 806 passed, 6 skipped (single local interpreter)
- [x] Real CLI `gflow image i2i --ref <png>` on an EN profile — image returned (timeline above)
- [ ] CI matrix 3.11 / 3.12 / 3.13 (fork PR → `action_required`; needs your approve-and-run)
- [ ] e2e i2i variant on a logged-in profile (`GFLOW_CLI_E2E_PROFILE=...`)

DCO-signed (both commits). Refs #56, #24.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   run could not acquire it and spiralled into rapid `about:blank` tabs +
   `TargetClosedError`. Context close + driver stop are now shared by
   `__aenter__`'s guard and `__aexit__` via `_close_browser_resources`.
+- `gflow image i2i --ref <local-file>` now binds the reference through the
+  editor's media dialog instead of the REST `uploadImage` endpoint (which 401s —
+  same root as #15/#39). Local-path refs ride a new `GenerateImageRequest.ref_paths`
+  field and are attached via the inherited R2V `_attach_references` (the image-mode
+  add-media dialog is the same `add_2` surface). Bare-UUID `--ref` still flows
+  through `refs` unchanged. Re-introduces #50 (reverted in #57 for the account/
+  locale variant tracked in #56); the media-dialog selectors are now
+  locale-agnostic (see the next entry).
 - `gflow image t2i/i2i --model` now actually selects the requested model. It was
   a no-op under `ui_automation` (the wire field was set but the model picker was
   never clicked, so Flow used its UI default). Adds `_select_image_model`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   through `refs` unchanged. Re-introduces #50 (reverted in #57 for the account/
   locale variant tracked in #56); the media-dialog selectors are now
   locale-agnostic (see the next entry).
+- The media-dialog upload selectors are now **locale-agnostic** (issue #56/#24).
+  `UPLOAD_MEDIA_BUTTON` matched localized text (`has-text('Upload media')`), so on
+  a non-English Chrome profile (Flow follows the *Chrome profile* language, which
+  the `--lang=en-US` arg cannot override) the click missed and the file chooser
+  never opened — a silent ~34s hang. It now anchors on the locale-free `upload`
+  icon ligature (`:text-is('upload')`, exact, so it doesn't grab the `Uploads`
+  tab), with the original English-text selector kept as a graceful **fallback
+  tier** (matches if Google ever changes the icon but keeps the English label);
+  'Add to Prompt' (which has no icon) is selected structurally as the only
+  iconless button in the open dialog. If neither tier opens a chooser,
+  `_upload_via_open_dialog` raises a clear error + writes a screenshot (no silent
+  hang) and points the operator at the Chrome-profile-language workaround. Fixes
+  I2I/I2V/R2V upload alike.
 - `gflow image t2i/i2i --model` now actually selects the requested model. It was
   a no-op under `ui_automation` (the wire field was set but the model picker was
   never clicked, so Flow used its UI default). Adds `_select_image_model`.
--- a/src/gflow_cli/api/image.py
+++ b/src/gflow_cli/api/image.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from dataclasses import dataclass
 from enum import StrEnum
+from pathlib import Path
 from types import MappingProxyType
 from typing import Any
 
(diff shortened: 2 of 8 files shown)


ffroliva:
## Verification on ffroliva (pt-BR) ✅ — production code passes. But the test has a bug that needs a 1-line fix before merge.

### Verification ledger (5 layers)

| Layer | Result | Evidence |
|---|---|---|
| 1. Exit code | `0` | pytest `PASSED` after 58s |
| 2. File count | N/A | Test asserts on the API contract (`media_name`, `fife_url`), doesn't download — fine, the structlog covers the upload step |
| 3. Pillow dims | N/A | (no file downloaded) |
| 4. structlog invariants | ✅ all green | `image_mode_entered` → `count_setter_completed` → **`image_uploaded status=200 target=ref0`** (upload fired — NOT the 34s silent hang from before) → `reference_attached index=0` → `prompt_submitted` (via locale-stable `arrow_forward` icon) → `batch_response_seen/captured status=200` |
| 5. Gallery confirm | pending repo owner | Flow project `becce409-a99b-4006-80bb-5ac516a562e1` |

**The locale-agnostic `upload` icon ligature selector works on a pt-BR account.** Mission accomplished on the production code — your fix direction was correct, and the exact-match `:text-is('upload')` refinement to avoid the `Uploads` tab confusion was a nice touch.

### Test transport bug — please amend before merge

```python
# tests/e2e/test_transports_e2e.py:169
async with _make_client("evaluate_fetch", profile) as client:  # ← REST
    image = await client.generate_image(req=req)
```

…but the docstring three lines up explicitly says:
> UI-automation transport ONLY — the REST transports (bearer/sapisidhash) cannot drive the add-media dialog, so they never invoke `_attach_references`.

So the test as committed routes through `evaluate_fetch` (REST), which silently drops `ref_paths` (REST has no media-dialog code path). I had to manually patch this to `"ui_automation"` on a local checkout to actually exercise your fix — that's the run whose evidence is above.

**This means your own EN-profile verification was a false positive** (sorry to be the bearer): your test sent a REST request with the prompt, Flow generated a text-only image, the `media_name + https://` assertion passed → green. The ref was never attached, the locale-agnostic selectors were never tested. On `ffroliva` we got a different false signal — `AuthExpiredError: evaluate_fetch HTTP 401` because the REST session token is stale — but again, never reached the actual fix code.

Only after patching to `"ui_automation"` did the test do what its docstring promises.

### One-line amend

```diff
- async with _make_client("evaluate_fetch", profile) as client:
+ async with _make_client("ui_automation", profile) as client:
```

(Could optionally inline a `_make_client` call with `transport=None` to use the registered default — same effect since `_DEFAULT_TRANSPORT = "ui_automation"` — but the explicit name is clearer.)

### Optional follow-up (separate PR / your call)

The assertion `image.media_name and image.fife_url.startswith("https://")` would pass even if `ref_paths` were silently ignored (i.e. it doesn't detect the very class of bug we just lived through). A stronger assertion would be useful — e.g. inspect the captured wire request payload for an `imageInputs[]` entry, or assert the generated image dimensions match `Aspect.PORTRAIT` (text-only Narwhal vs ref-conditioned Narwhal usually differ in framing). Not blocking — just flagging.

### State of PRs from this session

- **#56** (this issue): reopened post-revert, still tracking; will close on merge of #60
- **#57** (the revert) — merged
- **#59** (capture script `scripts/dev/capture_image_add_media_dom.py`) — open, CI green, separate; not needed for #60 but might pair well as a future tripwire

Thanks again for the speed on this — the focused scope, the `:text-is('upload')` refinement, the iconless structural selector for "Add to Prompt", the fail-loud RuntimeError with the **Chrome profile vs Google account language** hint, and the honest scoping of `FRAME_SLOT_BY_LABEL` to #24 are all great. Ready to merge as soon as the test transport is fixed.

svasakorn:
Thanks for merging #60! Heads-up: my test-transport fix (`evaluate_fetch` → `ui_automation`) landed a moment after the squash, so it missed the merge — the e2e on `develop` still routes through REST (the false positive you flagged). Small follow-up in #61. It also adds the `reference_attached` assertion so the test can't silently pass on a text-only image again.

PR #71, opened 2026-05-25
feat(image): native count selector, image batch command, --same-project

### LINKED ISSUE #14: feat(batch): use Flow's native count selector for -n N; add --same-project for multi-prompt batches
### Part 1 — Fix `-n N` to use Flow's UI count selector (bug)

**Today:** `gflow image t2i "prompt" -n 4` fans out 4 parallel single-image submissions into 4 separate projects. The `asyncio.Lock` in 1621748d made this safe, but the design is wrong.

**Should be:** Call `generate_images(count=N)` once. The transport sets the `x{N}` count button in Flow's UI (`1x / x2 / x3 / x4`), submits once, downloads N images from one project. Max `-n 4` (Flow's UI cap). CLI rejects `-n > 4` with a clear error.

Result: 1 project, N images, 1 submission round-trip. Strictly faster and matches user intent.

### Part 2 — Add `--same-project` for multi-prompt batches (feature)

Today there's no first-class multi-prompt CLI (each prompt = separate `gflow` invocation). When that lands (e.g. `gflow image batch manifest.tsv`), users will want a choice:

- **Default**: each prompt → its own new project.
- **`--same-project`**: all prompts share one project, submitted sequentially with random 3-7s jitter. On per-item failure, log and continue.

Combines cleanly with Part 1: `gflow image batch m.tsv -n 4 --same-project` = one project, each manifest row produces 4 images via the native selector, 3-7s jitter between rows.

## Implementation sketch

- **Part 1**: refactor `generate_images_batch` to delegate to `generate_images(count=N)` when given one prompt; remove the fan-out plus lock for that path (lock may still be needed for cross-profile concurrency — investigate). Add count-selector UI interaction. CLI validation: `-n` in `[1, 4]`.
- **Part 2**: new `--same-project` flag on the multi-prompt commands (when those land). New transport path `_run_batch_same_project(prompts, jitter_range)`.

## Open

- Flag name `--same-project` confirmed (bikeshed: `--group`, `--one-project`).
- Jitter range: 3-7s hardcoded for v1, configurable later.

## Out of scope

- Per-prompt `-n` (different counts per prompt) — future, behind a JSON manifest.
- Parallel-in-same-project — Flow's UI almost certainly can't handle it.

ffroliva:
Part 1 — fix -n N wiring (issue #14):
- `generate_images_batch` now makes ONE transport call with `count` set on
  the request instead of fanning out N parallel single-image submissions.
  The UI transport clicks x1/x2/x3/x4 and returns all N images in one
  round-trip. Simpler, faster, matches user intent.
- Extracted `_drive_images_generation` (returns list) so both single-image
  and multi-image paths share one token-mint + transport-call site.

Part 2 — `gflow image batch` subcommand (issue #14):
- Accepts JSON or TSV manifest files (dispatched by extension).
- TSV: `prompt[\tcount[\taspect_ratio[\tmodel]]]`; optional columns fall
  back to CLI defaults.
- JSON: `[{"text": "...", "count": 2, ...}]`; uses existing
  `parse_batch_item_dict` for per-entry validation.
- `MAX_BATCH_PROMPTS = 5` constant — easy to change or configure.
- `--same-project` flag: all prompts share one Flow project with 3-7s
  random jitter between submissions (vs. default: each prompt creates its
  own project).
- `run_manifest_image_batch` orchestrator lives in `image_batch.py`.

Tests: 35 new tests covering TSV parsing, JSON parsing, manifest file
dispatcher, and `run_manifest_image_batch` runner behaviour.

Closes #14

### DIFF
--- a/src/gflow_cli/image_batch.py
+++ b/src/gflow_cli/image_batch.py
@@ -494,6 +494,7 @@ def _tsv_parse_model(
     return model
 
 
+
 def parse_tsv_manifest(
     text: str,
     *,
@@ -703,6 +704,7 @@ async def _download_results(
             )
         )
     return outcomes
+
 
 
 async def run_manifest_image_batch(
--- a/src/gflow_cli/image_batch.py
+++ b/src/gflow_cli/image_batch.py
@@ -494,7 +494,6 @@ def _tsv_parse_model(
     return model
 
 
-
 def parse_tsv_manifest(
     text: str,
     *,
@@ -704,7 +703,6 @@ async def _download_results(
             )
         )
     return outcomes
-
 
 
 async def run_manifest_image_batch(


ffroliva:
Superseded — the count selector, `image batch` command, and `--same-project` shipped to develop through the data-layer / stay-mounted-batch work. This branch now produces a zero-line diff against develop (`changedFiles=0`), so its two commits are no-ops. Closing as already-merged. Reopen if a regression shows a gap.

PR #73, opened 2026-05-25
release: v0.9.1 — wheel-build fix (PyPI rescue for v0.9.x)

ffroliva:
## Why

The v0.9.0 release workflow [run #26415129126](https://github.com/ffroliva/gflow-cli/actions/runs/26415129126) failed at the PyPI publish step:

> 400 Invalid distribution file. ZIP archive not accepted: Duplicate filename in local headers.

Root cause: `pyproject.toml` had both `[tool.hatch.build.targets.wheel.force-include]` and `[tool.hatch.build.targets.sdist.force-include]` blocks pointing at `src/gflow_cli/data/migrations`, on top of the already-comprehensive `packages = ["src/gflow_cli"]` directive — hatchling included the migrations directory twice (both `__init__.py` and `0001_initial.sql`).

## Fix

Remove the redundant force-include blocks. Hatchling's default package inclusion already covers `.sql` files when they live inside `packages`.

## Verification

Local rebuild after the fix:
- Wheel total entries: 56, unique: 56, duplicates: `[]`
- `gflow_cli/data/migrations/0001_initial.sql` still included
- Impeccable Routine: hygiene · ruff · pyright src · 31/31 scoped tests — all green

## Release notes

`v0.9.1` carries the same product content as `v0.9.0` plus this single build fix. The v0.9.0 tag stays on GitHub for historical accuracy but never reached PyPI; v0.9.1 is the first PyPI release of the v0.9.x line.

Refs the v0.9.0 release sequence — recovery patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.9.1] — 2026-05-25
+
+### Fixed
+
+- **Wheel build no longer emits duplicate ZIP entries.** PyPI rejected the
+  `v0.9.0` upload with HTTP 400 *"Duplicate filename in local headers"*
+  because `pyproject.toml` had `[tool.hatch.build.targets.wheel.force-include]`
+  AND `[tool.hatch.build.targets.sdist.force-include]` blocks pointing at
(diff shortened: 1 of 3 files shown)


ffroliva:
Closing in favor of force-rewriting v0.9.0 with the wheel-build fix (per maintainer direction). v0.9.1 will not be cut.

PR #70, opened 2026-05-25
feat(selectors): structural-first cascade for onboarding + I2V frame slots (issue #24 phase 2)

ffroliva:
## Summary

- Restructures `ONBOARDING_SELECTORS` into a structural-first cascade: 5 locale-free ARIA/ID anchors (`button#L2AGLb` + ARIA-label variants covering Google's Funding Choices SDK in any language) followed by ~35 `:has-text()` entries spanning 12 locales (EN, PT, DE, ES, FR, IT, NL, JA, ZH, KO, PL, RU + TR, ID).
- Flips `_attach_frame` (I2V/R2V frame slots) from English-text-first to structural-first: tries `FRAME_SLOTS_STRUCT` (`swap_horiz` icon + positional index, locale-free) before falling back to `FRAME_SLOT_BY_LABEL`, removing the hard dependency on English UI labels.
- Updates `KNOWN_ISSUES.md` issue #24 entry with Phase 2 progress and revised workaround guidance.

## Changes

| File | Change |
|---|---|
| `src/gflow_cli/api/transports/ui_automation.py` | Split `ONBOARDING_SELECTORS` into `_ONBOARDING_STRUCTURAL_SELECTORS` + `_ONBOARDING_TEXT_SELECTORS`; updated `--lang=en-US` comment |
| `src/gflow_cli/api/transports/ui_automation_video.py` | `_attach_frame` structural-first selection with text-label fallback |
| `tests/api/transports/test_ui_automation.py` | Cascade-ordering invariant tests in `TestBypassOnboarding`; structural-tier click test |
| `tests/api/transports/test_ui_automation_video.py` | New `TestAttachFrameSlotSelection` (4 tests: structural path, text fallback, double-miss error, image-missing error) |
| `KNOWN_ISSUES.md` | Issue #24 entry updated with Phase 2 progress |

## Test plan

- [x] `uv run python -m pytest tests/api/transports/test_ui_automation.py tests/api/transports/test_ui_automation_video.py -v` — 148 passed
- [x] `uv run pyright src/gflow_cli/api/transports/ui_automation.py src/gflow_cli/api/transports/ui_automation_video.py` — 0 errors
- [x] `uv run ruff check` — all checks passed
- [x] `uv run ruff format` — 3 files reformatted
- [ ] Live e2e on non-English Chrome profile — deferred (required to remove `--lang=en-US`)

## Resolves / Closes

[...]

PR #124, opened 2026-05-30
fix(transports): exit Agent composer mode before mode switch so the gen panel mounts

svasakorn:
## Problem

On accounts whose Flow editor shows the new **"Agent"** composer toggle next to the prompt box, create-project generation intermittently fails. When Agent mode is active, Flow removes the entire media-generation panel from the DOM — the `crop_*` settings trigger, the Image/Video mode tablist, and the count/model controls all disappear. `_switch_to_image_mode` / `_switch_to_video_mode` then can't find their trigger and raise:

```
RuntimeError: mode-switch dropdown trigger not found on the Flow editor.
```

This affects both image (`generate_images`, stay-mounted batch) and video paths.

## Fix

Add `VideoGenerationMixin._exit_agent_mode(page)`, called at the top of **both** `_switch_to_image_mode` and `_switch_to_video_mode`:

- If the locale-stable `crop_*` settings trigger is already present → media mode → **no-op** (the common path; the toggle is never touched).
- If the panel is absent → Agent mode → click the pill, then **confirm the panel actually re-mounted**. If it does not, log a warning and return `False` rather than claim a false "exited" — the caller's own trigger probe then fails loudly. Best-effort: never raises.

Panel presence is a small `_media_panel_present()` helper, reused by the e2e.

## Selector — scoped, locale-safe, no ARIA

```python
COMPOSER_AGENT_TOGGLE_SELECTOR = (
    "div:has(div[role='textbox'][data-slate-editor='true'])"
    ":has(button:has(i:text('arrow_forward'))) button:has(span.content)"
)
```

- **Scoped (must-fix #1):** the pill is matched only inside the composer that holds **both** the Slate prompt box **and** the `arrow_forward` submit — so `.first` can't grab an unrelated `span.content` button added elsewhere in a future Flow build. The composer's ancestor chain has no stable id/role/data-* (all styled-component hashes), so the prompt box + submit icon are the structural anchors.
- **Locale-safe (issue #24):** no visible-text match and the literal "Agent" never appears. The only `:text()` is `arrow_forward`, a Material Symbols icon ligature (locale-invariant — same technique as the existing `crop_*` anchors).
- **No ARIA:** mode is detected from the *absence* of the `crop_*` trigger, not an `aria-pressed` state.

## Tests

- **Unit** — `TestExitAgentMode` (7): no-op when panel present, click + confirm re-mount, **returns False when the click does not re-mount** (no false success), no-op when toggle absent, never-raises on probe error, a selector guard (locale-safe + aria-free + scoped), and a **stdlib-parser scope guard** proving a decoy `span.content` button *outside* the composer is excluded. `TestModeSwitchExitsAgentFirst` (2): both mode switches call `_exit_agent_mode` first.
- **E2E** — `test_e2e_agent_mode_recovered_before_mode_switch` (`e2e_auth`, **zero credits**): reproduces Agent mode on the real DOM, **asserts the scoped selector resolves to exactly one element** (`count() == 1` — the browser-only check the unit test can't do), then proves `_switch_to_image_mode` re-mounts the panel + emits `exited_agent_mode`. Verified locally on a live profile (1 passed, ~18 s).

## Review items

- **#1 selector uniqueness** — addressed both ways the review offered: scoped the selector to the composer **and** added a unit scope-guard + a live `count() == 1` assert.
- **#2 structlog prefix** — events now use `ui_automation_video.` to match the module.
- **#3 defense-in-depth** — composer-scoping hardens against stray clicks.
- **#4 video-path wiring** — `TestModeSwitchExitsAgentFirst` pins the call-site for both paths.
- **#5 non-EN evidence** — scoping the claim honestly: the selector is locale-safe **by construction** (zero localized text — pinned by unit test). A genuine non-EN *live* run isn't possible here: per the `flow-locale-driven-by-chrome-profile-lang` finding the UI follows the Chrome **profile** language, and `GFLOW_CLI_LOCALE=de-DE` on this EN profile still renders `htmlLang=en` / `?hl=en` — a true non-EN run needs a German-created profile I don't have.

## Validation

- [x] hygiene · doc-links · `ruff check` · `ruff format --check` · `pyright src` (0 errors) — all clean
- [x] `pytest --cov=gflow_cli --cov-fail-under=80` — **1056 passed, 85.9% coverage**
- [x] Live e2e verified on a real authenticated profile (scoped selector `count() == 1`)
- [x] CHANGELOG under `[Unreleased] → Fixed`
- [x] DCO `Signed-off-by` present

_Updated at head `e757e09`._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Create-project generation failing when Flow's "Agent" composer mode is active.**
+  Flow's newer editor adds an Agent toggle next to the prompt box; when it is on,
+  the media-generation panel (the `crop_*` settings trigger, Image/Video mode
+  tablist, and count/model controls) is removed from the DOM, so the UI-automation
+  transport raised "mode-switch dropdown trigger not found". `_switch_to_image_mode`
+  and `_switch_to_video_mode` now call `_exit_agent_mode()` first, which re-mounts
+  the panel by clicking the toggle off. Detection is locale-invariant and uses no
+  UI text and no `aria-` attribute (`button:has(span.content)` plus the absence of
+  the locale-stable `crop_*` trigger), so it works in every Flow UI language.
+
 ## [0.10.0] — 2026-05-29
 
 ### Fixed
--- a/src/gflow_cli/api/transports/ui_automation.py
+++ b/src/gflow_cli/api/transports/ui_automation.py
@@ -832,6 +832,10 @@ class UiAutomationTransport(VideoGenerationMixin):
         The dropdown is closed afterwards (via :kbd:`Escape`) so the caller's
         :meth:`_configure_generation_settings` can open it fresh.
         """
+        # New Flow UI: if the composer is in Agent mode the generation panel is
+        # absent — switch back to media mode first so the trigger probe below
+        # can find the crop_* dropdown.
+        await VideoGenerationMixin._exit_agent_mode(page)
         trigger = await VideoGenerationMixin._probe_selector_cascade(
             page,
             "mode_switch_trigger",
--- a/src/gflow_cli/api/transports/ui_automation_video.py
+++ b/src/gflow_cli/api/transports/ui_automation_video.py
@@ -80,6 +80,44 @@ VIDEO_TAB_IN_MENU_SELECTORS = (
     "[role='menu'] [role='tab']:has-text('Vídeo')",
     "[role='menu'] [role='tab']:has-text('Video')",
 )
+
+# Composer "Agent" mode toggle. Flow's newer editor puts a pill toggle next to
+# the prompt box: a ``<button>`` whose only label is a ``<span class="content">``.
+# When Agent mode is on, the whole media-generation panel is REMOVED from the
+# DOM — the aspect/settings button (the locale-stable ``crop_*`` icon trigger
+# keyed on by MODE_SWITCH_TRIGGER_SELECTORS / GEN_SETTINGS_BUTTON_SELECTORS), the
+# Image/Video tablist, and the count/model controls all disappear, so
+# ``_switch_to_image_mode`` / ``_switch_to_video_mode`` raise "mode-switch
+# dropdown trigger not found". Clicking the pill returns to media mode.
+#
+# This selector is deliberately STRUCTURAL — no localized text and no ARIA:
+#  * No localized text: the pill's "Agent" label is translated in some Flow
+#    locales, so matching it by visible text would regress non-English users
+#    (issue #24: locale-agnostic selectors — a recurring source of PR pushback in
+#    this module). The only ``:text(...)`` here is ``arrow_forward``, a Material
+#    Symbols icon ligature, which is locale-invariant — the same technique the
+#    module already uses for ``crop_*`` / SUBMIT_BUTTON_SELECTORS anchors.
+#  * No ARIA: aria-* anchors have also been pushed back on in past reviews, and
+#    one is not needed here — Agent mode is detected from the *absence* of the
+#    ``crop_*`` media trigger, so the toggle only has to be located, not have its
+#    state read.
+#
+# SCOPED to the generation composer (PR #124 review must-fix): the pill is
+# matched only inside the element that holds BOTH the Slate prompt box AND the
+# ``arrow_forward`` submit button. Page-wide there is exactly one
+# ``button:has(span.content)`` today (live-verified count == 1), but ``.first``
+# on the bare global selector would silently grab the wrong element if a future
+# Flow build added another ``span.content`` button (header/sidebar) ordered
+# before the pill. Scoping to the prompt+submit composer keeps the match correct
+# regardless of unrelated additions elsewhere. The composer's own ancestor chain
+# carries no stable id/role/data-* attribute (all styled-component hashes), so
+# the prompt box and submit icon ARE the stable structural anchors. Uniqueness is
+# pinned by a structural unit test (decoy outside the composer) and asserted live
+# (count == 1) in the e2e.
+COMPOSER_AGENT_TOGGLE_SELECTOR = (
+    "div:has(div[role='textbox'][data-slate-editor='true'])"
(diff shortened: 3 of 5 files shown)


ffroliva (changes requested):
## Council review — Verdict: 🟡 YELLOW (soft block)

Thanks for this fix — the structure is sound: the `_exit_agent_mode` pre-step is called symmetrically from both `_switch_to_image_mode` and `_switch_to_video_mode`, the no-op fast path keeps it cheap on the common path, it's best-effort/never-fatal (a real re-mount failure still surfaces via the downstream trigger probe), the CHANGELOG entry is accurate, DCO is present, and CI is green on 3.11/3.12/3.13. A 6-dimension council (correctness / quality / security / tests / memory / UI-verify) rated 5 dimensions GREEN.

One must-fix holds it at YELLOW:

### Must-fix
1. **Selector-uniqueness is asserted, not verified** — `COMPOSER_AGENT_TOGGLE_SELECTOR = "button:has(span.content)"` (`src/gflow_cli/api/transports/ui_automation_video.py:103`). The comment claims `span.content` is "the only `span.content` in the editor — verified unique", but `.first` blind-picks match #0 and **no test pins this**. The no-op `crop_*`-absence guard bounds *when* the click fires, not *which* button it hits. If a future Flow build adds another `button:has(span.content)` ordered before the Agent pill, the click lands on the wrong element and re-mount silently fails — and CI never catches it (the e2e is opt-in `e2e_auth`). Please either (a) add a unit/DOM-fixture assertion that the selector resolves to exactly one element on a captured Agent-mode DOM snapshot, or (b) scope the selector to the composer container near the prompt box.

### Nice-to-have
2. **structlog event prefix** — `log.info("ui_automation.exited_agent_mode")` / `log.debug("ui_automation.agent_toggle_probe_failed")` use the `ui_automation.` prefix, but the method lives in `ui_automation_video.py` where every other event uses the `ui_automation_video.` prefix. Consider aligning for grep consistency.
3. **Defense-in-depth (security/UI)** — scoping the selector to a composer-local ancestor would also harden against accidental clicks on unrelated controls.
4. **Video-path wiring is unit-untested** — a unit test confirms `_exit_agent_mode`'s branches, but only the live e2e exercises the *image-path* call site. A cheap mock test asserting `_switch_to_video_mode` calls `_exit_agent_mode` first would close the gap.
5. **Single-locale live evidence** — "1 passed in 17.59s" was almost certainly EN. The selector *should* be locale-safe (no text/no ARIA), but per our `pr-must-verify-on-affected-surface` history (PR #70 verified the wrong surface), one zero-credit `e2e_auth` run on a non-EN profile (e.g. de-DE) would confirm it, or scope the locale claim to "structural-selector, untested on non-EN".

_Reviewed at head `22eaae9`. Council dimensions: D1 GREEN · D2 GREEN · D3 GREEN · D4 GREEN · D5 GREEN · D6 YELLOW._

svasakorn:
Thanks for the council review — all five items addressed, pushed at head `e757e09`.

**#1 — Selector uniqueness (must-fix).** Did both options you offered. The selector is now **scoped to the composer** — matched only inside the element holding both the Slate prompt box and the `arrow_forward` submit:

```python
COMPOSER_AGENT_TOGGLE_SELECTOR = (
    "div:has(div[role='textbox'][data-slate-editor='true'])"
    ":has(button:has(i:text('arrow_forward'))) button:has(span.content)"
)
```

so `.first` can no longer grab an unrelated `span.content` button. Uniqueness is now **pinned**, not asserted: a stdlib-parser unit test (`test_scope_excludes_a_decoy_span_content_button_outside_composer`) proves a decoy `span.content` button in the page header is excluded while the composer pill is matched, and the e2e asserts `count() == 1` on the real DOM (the browser-only check the unit can't do).

**#2 — structlog prefix.** Renamed to `ui_automation_video.exited_agent_mode` / `.exit_agent_mode_no_panel` / `.agent_toggle_probe_failed` to match the module.

**#3 — defense-in-depth.** The composer-scoping also hardens against accidental clicks on unrelated controls, as you noted.

**#4 — video-path wiring.** Added `TestModeSwitchExitsAgentFirst` — cheap mock tests pinning that **both** `_switch_to_image_mode` and `_switch_to_video_mode` call `_exit_agent_mode` before the trigger probe. (Caught a real bug writing these: the helpers are referenced on `VideoGenerationMixin`, so the patch target had to be the base class, not the subclass.)

**#5 — single-locale evidence.** Scoping the claim honestly rather than overstating it. The selector is locale-safe **by construction** — zero localized text (the only `:text()` is the `arrow_forward` icon ligature), pinned by a unit assertion. I could not produce a genuine non-EN *live* run: per the locale-driven-by-Chrome-profile-language behaviour (root cause of #56), the UI follows the Chrome **profile** language, and I verified `GFLOW_CLI_LOCALE=de-DE` on this EN-created profile still renders `htmlLang=en` / `?hl=en` / pill label "Agent" — so a de-DE run would be EN with a German `Accept-Language`, not real non-EN evidence. A true non-EN run needs a German-created profile I don't have; flagging that rather than claiming coverage I can't back.

**Verification ledger (head `e757e09`):**
- hygiene · doc-links · `ruff check` · `ruff format --check` · `pyright src` (0 errors) — clean
- `pytest --cov=gflow_cli --cov-fail-under=80` — **1056 passed, 85.9%**
- Live e2e on a real profile — `1 passed`, scoped selector `count() == 1`, panel re-mounted, `exited_agent_mode` emitted
- CHANGELOG under `[Unreleased] → Fixed` · DCO present

ffroliva (approved):
Re-reviewed at e757e09 — must-fix resolved: selector scoped to the prompt+submit composer, uniqueness pinned by a structural decoy unit test + live count==1, plus a defensive panel-remount re-check and both call-site wiring tests. Council verdict now GREEN. Approving.

ffroliva:
Merged into `develop` (merge commit `1d84239`). Thank you for this fix, @svasakorn / THEIA! 🙏 A genuinely high-quality external contribution — you diagnosed a real Flow-UI regression and your response to the council review was exemplary: composer-scoped selector, uniqueness pinned by a decoy unit test *and* a live `count == 1`, a defensive panel-remount re-check so a failed toggle never reports false success, and call-site ordering covered for both image and video paths. Locale discipline (issue #24) respected throughout. Much appreciated. 🎉

PR #138, opened 2026-06-01
fix(transports): handle Flow's Agent chat-panel shape in _exit_agent_mode

svasakorn:
## Problem

Follow-up to the earlier Agent-pill fix (`_exit_agent_mode`). Flow now surfaces Agent mode in **two non-deterministic shapes** on project open:

1. the in-composer **pill** (already handled), and
2. a docked **chat side-panel** ("Untitled session") on the right.

While the chat panel is up, the in-composer pill is **absent from the DOM**, so the pill-only recovery found nothing to click and create-project generation still failed with:

```
RuntimeError: mode-switch dropdown trigger not found on the Flow editor.
```

Which shape appears is non-deterministic per open — the "sometimes it happens, sometimes it doesn't" report.

## Fix

`_exit_agent_mode` now drives a small **bounded loop** handling both shapes in one pass: while the media panel is absent, dismiss whichever Agent affordance is present — the **chat panel first** (its X reveals the pill), then the **pill** — and re-check `crop_*`. It is keyed on the **outcome** (`crop_*` is back), never on assuming which control exists, so it covers pill-only, panel-only, panel-then-pill, and neither (older UI) without special-casing. The pill is clicked **at most once** (it's a binary toggle — clicking twice would re-enter Agent mode).

## New selector — structural, locale-safe, no ARIA

```python
AGENT_CHAT_PANEL_CLOSE_SELECTOR = (
    "div:has(button:has(i:text-is('edit_square'))) button:has(i:text-is('close'))"
)
```

- Anchored on the panel header's New-session (`edit_square`) + close (`close`) Material Symbols ligatures — locale-invariant, no localized text, no `aria-` (same discipline as the pill selector).
- `:text-is` is **EXACT** so it does **not** also match the sidebar's `left_panel_close` ligature (`:text('close')` would over-match).
- Click timeouts are named constants (`_AGENT_CLICK_TIMEOUT_MS` / `_AGENT_SETTLE_MS`), not magic numbers; clicks are `force=True` (the panel's entry animation otherwise stalls Playwright's actionability check).

## Tests

- **Unit** — `TestExitAgentMode` reworked onto a stateful page mock that models Flow's real transitions (chat-close reveals the pill; pill click re-mounts the panel): media-noop, pill, **chat-then-pill**, **pill-clicked-once (no flip-flop)**, no-affordance, plus a locale-safe/aria-free guard for the new chat-close selector. `TestModeSwitchExitsAgentFirst` unchanged.
- **E2E** — adds `test_e2e_agent_chat_panel_recovered_before_mode_switch` (`e2e_auth`, zero credits) reproducing the chat-panel shape on the real DOM and asserting recovery.

## Live verification

Drove the **real code path** (`setup → _enter_editor` [creates a fresh project] → `_switch_to_image_mode` → `_exit_agent_mode`) across profiles **0001–0006, 3 project opens each = 18 runs**. The randomly-appearing Agent shape was caught and recovered every time:

| Profile | Shape (random, not forced) | Recovered |
|---|---|---|
| 0001 | media ×3 | ✅✅✅ |
| 0002 | media ×3 | ✅✅✅ |
| 0003 | **PILL** ×3 | ✅✅✅ |
| 0004 | **CHAT-PANEL** ×3 | ✅✅✅ |
| 0005 | media ×3 | ✅✅✅ |
| 0006 | **CHAT-PANEL** ×3 | ✅✅✅ |

All three shapes (media / pill / chat-panel) occurred organically; **18/18** recovered.

## Validation

- [x] hygiene · doc-links · `ruff check` · `ruff format --check` · `pyright src` (0 errors) — all clean
- [x] `pytest --cov=gflow_cli --cov-fail-under=80` — **1151 passed, 87.25% coverage**
- [x] Live-verified across 6 profiles / 18 project opens (table above)
- [x] CHANGELOG under `[Unreleased] → Fixed`
- [x] DCO `Signed-off-by` present

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Create-project generation failing when Flow opens the Agent _chat panel_.**
+  A follow-up to the earlier Agent-pill fix: Flow now also surfaces Agent mode as
+  a docked chat side-panel ("Untitled session") on some project opens, and while
+  it is up the in-composer Agent pill is absent from the DOM — so the pill-only
+  recovery could not find anything to click and generation still failed with
+  "mode-switch dropdown trigger not found". `_exit_agent_mode` now handles both
+  Agent shapes in one pass: it dismisses the chat panel (locale-stable, aria-free
+  structural close anchor) which reveals the pill, then turns the pill off,
+  looping until the media panel re-mounts. Keyed on the outcome (`crop_*` is
+  back), so it covers pill-only, panel-only, and panel-then-pill without assuming
+  which control is present.
+
 ### Added
 
 - **`gflow scene` command group (Add Clip / Scenes).** Compose ordered,
--- a/src/gflow_cli/api/transports/ui_automation_video.py
+++ b/src/gflow_cli/api/transports/ui_automation_video.py
@@ -128,6 +128,29 @@ COMPOSER_AGENT_TOGGLE_SELECTOR = (
     "div:has(div[role='textbox'][data-slate-editor='true'])"
     ":has(button:has(i:text('arrow_forward'))) button:has(span.content)"
 )
+
+# Agent CHAT side-panel close (X). Flow's even-newer editor sometimes promotes
+# Agent mode from the in-composer pill (above) to a full chat panel docked on the
+# right ("Untitled session", "What would you like to do?") — it appears on some
+# project opens and not others. While that panel is up, the in-composer pill is
+# NOT in the DOM at all, so the pill selector matches nothing; the panel must be
+# dismissed first (its X), after which the pill reappears (usually still active)
+# and the normal pill path takes over. The panel header carries a New-session
+# button (``edit_square`` icon) next to its close (``close`` icon); we anchor on
+# that pairing so we hit the panel's X and not some other ``close`` icon on the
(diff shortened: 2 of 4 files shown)


ffroliva:
Merged — thank you for this, @svasakorn! 🙏

This is a high-quality contribution. I ran it through our multi-dimensional council review (`/gflow:pr-council-review`) and it came back **GREEN across correctness, code quality, security, tests, and UI/live-verification**:

- **Root-cause fix, not a patch** — generalizing `_exit_agent_mode` into a bounded, *outcome-keyed* loop (keyed on the `crop_*` re-mount, not on guessing which control exists) cleanly handles all of pill-only / panel-only / panel-then-pill, and never aborts paid generation on a DOM probe error.
- **Locale-stable selectors** — Material-Symbols ligatures + `:text-is` exact-match (correctly avoiding the `left_panel_close` over-match). This is exactly our hard rule for Flow's localized UI.
- **Genuinely live-verified on the affected surface** — 18/18 recovery across 6 profiles with all three Agent shapes occurring organically, plus a zero-credit `e2e_auth` test asserting the `exited_agent_mode` event. That's the bar we hold transport changes to, and you cleared it.

The only follow-ups are minor, maintainer-side polish (a memory-doc update on our side + three tiny consistency nits like qualifying one `i:` selector with `.google-symbols`). I've tracked them in a separate issue (#139) so they don't hold up your fix. Nothing for you to do here — really appreciate the careful work and the thorough verification. 🎉

PR #168, opened 2026-06-09
Add browser cookie profile verification and related tests

3mora2:
## Summary
Add verify_flow_profile() to probe a Chrome profile's session using browser_cookie3 and httpx. The new routine ensures the profile directory is inside GFLOW_CLI_HOME, locates and loads Chrome cookies (using Local State when present), calls the session API and maps responses via evaluate_session_response; failures are treated fail-closed and logged as VERIFICATION_ERROR. Added get_cookies_path() helper to resolve common Chrome cookie locations. Also add browser-cookie3 to project dependencies and ignore uv.lock in .gitignore.

## Validation

- [X] Focused tests added or updated for behavior changes
- [X] Documentation updated or explicitly marked not applicable
- [X] `uv run python scripts/ci/check_doc_links.py`
- [X] `uv run ruff check src tests`
- [X] `uv run pyright src`
- [X] Relevant pytest command:

## Contribution Checklist

- [X] This PR targets `develop`, unless it is a release or emergency fix
- [X] My commits use my real Git identity or GitHub noreply email
- [X] External contribution commits include `Signed-off-by:` (`git commit -s`)
- [X] I did not include secrets, cookies, account tokens, signed URLs, or private captured data
- [X] I reviewed any AI-assisted changes before submitting


ffroliva (dismissed):
*(Duplicate — please see the formatted review below. This review has been dismissed.)*

ffroliva (changes requested):
Thanks for contributing this, Ammar! The idea of a lightweight `browser_cookie3`+`httpx` path for session verification is sound and fits `gflow-cli` well. Three issues need fixing before merge.

---

## Must-fix

### 1. Broken test — patches the wrong target

**File:** `tests/auth/test_verification.py` (`test_authenticated_profile_httpx`)

The test patches `gflow_cli.auth.strategies.async_playwright`, but `verify_flow_profile` never calls Playwright — it uses `browser_cookie3` and `httpx` (both imported **inside** the function body). The patch has no effect.

In practice, `get_cookies_path(profile_dir)` raises `FileNotFoundError` on the empty tmp directory, the `except Exception` block catches it, and the function returns `VERIFICATION_ERROR`. The assertion `status.outcome is FlowSessionOutcome.AUTHENTICATED` will always fail.

Because `browser_cookie3` and `httpx` are imported inside the function, you need to inject them via `sys.modules`:

```python
import sys
from unittest.mock import AsyncMock, MagicMock, patch

async def test_authenticated_profile_httpx(self, gflow_home: Path) -> None:
    profile = gflow_home / "profile_default"
    profile.mkdir()

    fake_bc3 = MagicMock()
    fake_bc3.chrome.return_value = [MagicMock(name="SAPISID")]

    fake_resp = MagicMock(status_code=200, text=AUTHENTICATED_BODY)
    fake_client = MagicMock()
    fake_client.__aenter__ = AsyncMock(return_value=fake_client)
    fake_client.__aexit__ = AsyncMock(return_value=False)
    fake_client.get = AsyncMock(return_value=fake_resp)

    fake_httpx = MagicMock()
    fake_httpx.AsyncClient.return_value = fake_client

    with (
        patch("gflow_cli.auth.verification.get_settings") as mock_settings,
        patch("gflow_cli.auth.verification.get_cookies_path", return_value=Path("/fake/Cookies")),
        patch.dict(sys.modules, {"browser_cookie3": fake_bc3, "httpx": fake_httpx}),
    ):
        mock_settings.return_value.home = gflow_home
        status = await verify_flow_profile(profile)

    assert status.outcome is FlowSessionOutcome.AUTHENTICATED
    assert status.user_email == "test.user@example.com"
```

Also add at minimum: a test for the profile-outside-`GFLOW_CLI_HOME` case (should raise `SecurityError`) and one for the exception path (`browser_cookie3.chrome` raises `PermissionError` → returns `VERIFICATION_ERROR`).

---

### 2. Hardcoded Arabic locale in HTTP headers

**File:** `src/gflow_cli/auth/verification.py` (headers dict inside `verify_flow_profile`)

```python
# Current — Arabic locale hardcoded (appears to be your personal browser locale):
"accept-language": "ar,en-US;q=0.9,en;q=0.8",
"referer": "https://labs.google/fx/ar/tools/flow",

# Should match verify_flow_session and the existing transports:
"accept-language": "en-US,en;q=0.9",
"referer": "https://labs.google/fx/tools/flow",
```

Shipping your personal locale means all users will send Arabic-preference headers to the Flow API, which may trigger locale-specific responses or behaviour differences from the server.

---

### 3. `uv.lock` added to `.gitignore` while also being committed

**File:** `.gitignore`

The PR adds `uv.lock` to `.gitignore` but also commits the updated `uv.lock`. For a CLI tool like `gflow-cli`, the lock file is committed intentionally so installs are reproducible. Adding it to `.gitignore` after committing it creates confusion — git won't untrack an already-committed file, but future contributors will be told not to commit lock-file updates.

Please remove `uv.lock` from `.gitignore`. If you want to avoid including `uv.lock` churn in your personal commits, the right approach is to just not `git add` it — the `.gitignore` entry won't help once the file is already tracked.

---

## Nice-to-have (not blockers)

- `get_cookies_path()` in `paths.py` raises a bare `FileNotFoundError` with no message. `FileNotFoundError(f"No Chrome Cookies file found under {pdir}")` would make debugging much easier.
- `auth/__init__.py:status()` already has the same three cookie-path candidates inlined. Now that `get_cookies_path()` exists, `status()` could call it to eliminate the duplication.
- The `_cookies` variable name (underscore prefix conventionally signals "throwaway" in Python). Renaming it `chrome_cookies` better communicates that it's actively used.
- A `CHANGELOG.md` entry under `[Unreleased]` for the new `verify_flow_profile` function and `browser-cookie3` dependency would be appreciated.

---

Happy to help if you have questions about the `sys.modules` patching approach. Once the three must-fix items are addressed this is ready for another round.

ffroliva (changes requested):
## Council review — 🟡 Changes requested

Thanks for this, @3mora2 — the `verify_flow_profile` implementation itself is clean, well-tested (23 tests covering all four outcomes, retry, httpx, and the security boundary), and **security-solid** (the council confirmed `detail` is leak-proof, no cookie/token/email reaches any logger, fail-closed AUTHENTICATED). `uv.lock` is fully in sync with the new `browser-cookie3` dep and its transitive closure. No correctness or security defects.

The block is that the feature is **half-landed**. Three things to fix before merge (verified against current head `92536c2` — the recent `develop` merge synced files but didn't change these):

### Must-fix

1. **`verify_flow_profile` / `get_chrome_cookies` have no production caller.** `git grep verify_flow_profile -- src/**/*.py` returns only the definition; `src/gflow_cli/auth/real_chrome.py:207` still calls the old Playwright `verify_flow_session`. So the PR adds a runtime `browser-cookie3` dependency + a cookie-decryption/network surface that nothing actually invokes. Please either **wire it into `RealChromeStrategy`** (the intended fast-path) or **hold the dependency** until it's wired.

2. **CHANGELOG oversells it.** The entry says "verify sessions **without launching a heavy Playwright browser**", but because of (1) the heavy path is still the only one that runs. Either make (1) true, or reword to note the API is added-but-not-yet-wired.

3. **When you wire it in (1), add the chrome-strategy marker guard** on the Playwright fallback. Per the project's real-browser-auth rule, any path that drives Flow must require a chrome-marker profile and fail fast via `channel_for_profile()`. The fallback currently defaults `channel="chrome"` without that check, so a marker-less profile would launch bundled Chromium instead of failing with the `--browser chrome` hint (mirror `real_chrome.py`).

### Nice-to-have (not blocking)

- **Don't fork the security-critical probe** — `verify_flow_profile` copies the `SecurityError` boundary + fail-closed `except` + durability-log block verbatim from `verify_flow_session`. Parameterize the cookie/transport step of the existing function instead, so the boundary check isn't maintained in two places.
- **Move cookie extraction out of `verification.py`** (e.g. `auth/cookies.py`) — decryption is a separate responsibility from "is this profile signed in?". Note `get_cookies_path` correctly went into `paths.py`.
- **Narrow** `_get_chrome_cookies3`'s `except Exception` → `browser_cookie3.BrowserCookieError` so a missing file / import error isn't masked as "decryption failed".
- **Filter the cookie jar to the Flow domain** before handing it to httpx (currently sends the full Chrome cookie jar to labs.google).
- Consider an upper bound on `browser-cookie3` (it touches DPAPI/keychain) and `follow_redirects=False` on the cookie-bearing probe (defense-in-depth; host is Google-controlled).

Happy to re-run the council once (1)–(3) are addressed.

ffroliva:
## On test coverage — yes, this needs a live e2e (and we already have the pattern to copy)

The 23 unit tests are excellent, but they **mock** `browser_cookie3`, `httpx`, and the session response — so they prove the logic is sound but can't catch the three things that actually carry risk on this surface:

1. whether `browser_cookie3` really decrypts a live gflow Chrome cookie store (Windows DPAPI / AES-GCM — the exact failure your Playwright fallback exists to handle),
2. whether the real `/fx/api/auth/session` body still matches the pinned `AUTHENTICATED_BODY` fixture,
3. whether it reads the **gflow** profile dir, not the OS-default Chrome profile.

A mocked suite passing while the real wire is broken has bitten this repo before, so a live e2e is part of "done" here. **Good news: it costs zero credits** (cookie read + one `GET` — no generation), so the only requirement is a logged-in profile.

### Follow the existing pattern — don't invent one

We already have a near-identical e2e for the *old* path: **`tests/e2e/test_auth_verification_e2e.py`** tests `verify_flow_session` against the live endpoint. Add your `verify_flow_profile` cases to that same file, mirroring its conventions exactly:

- **Markers:** `pytestmark = [pytest.mark.e2e, pytest.mark.e2e_auth]` — `e2e_auth` is our zero-credit, browser-only tier (see the marker table in `pyproject.toml` and `docs/E2E_TESTING.md`).
- **Opt-in:** runs only under `-m e2e` with `GFLOW_CLI_E2E_PROFILE=<logged-in-profile>` set; it's excluded from the default CI run by `addopts` in `pyproject.toml`, so it won't break CI for people without a profile.
- **Fixtures (from `tests/e2e/conftest.py`):** use `e2e_profile_dir` for the authenticated case and `e2e_nosession_profile` for the empty-profile case — do **not** roll your own profile resolution (the fixture deliberately bypasses `GFLOW_CLI_HOME` isolation and handles the Windows Chrome-profile-lock teardown).
- **No `@pytest.mark.asyncio`** — `asyncio_mode = \"auto\"` is set; just write `async def`.

### The assertions that make it worth running

Copy the **linked-invariant** style the existing test uses (verify says AUTHENTICATED *and* the same profile actually works), plus cover the fallback that's unique to your path:

1. `e2e_profile_dir` → `verify_flow_profile(...)` returns `AUTHENTICATED` with a non-empty `user_email`, **and** the same profile passes `FlowApiClient.health_check()` — proves the cookie-decrypt + real session probe agree with the real client.
2. `e2e_nosession_profile` → returns `NO_SESSION` (empty profile → `200 {}`).
3. **Fallback path** (the one thing the old test can't cover): force a `browser_cookie3` decrypt failure → assert it falls through to the Playwright path rather than silently erroring. This is the highest-value assertion in the whole PR.

### Sequencing — to avoid rework

Land this e2e **in the same revision that wires `verify_flow_profile` into production** (must-fix #1 in my review). Writing it before the wiring means testing a path nothing calls, and you'd likely rewrite it once the call site is settled. So: wire it in → add the `e2e_auth` cases above → reword the CHANGELOG → done.

Reference: `docs/E2E_TESTING.md` for the full layer/marker reference, and `tests/e2e/test_auth_verification_e2e.py` as your template.

ffroliva (changes requested):
## Delta re-review — current head `1127f0a`

The previous wiring, CHANGELOG, Chrome-marker guard, locale/header, `uv.lock`, and test-addition requests are resolved. The existing E2E comment still applies only to **recording an executed zero-credit `e2e_auth` result**; the test cases themselves are now present.

Three newly verified blockers remain:

1. **Windows DPAPI failure bypasses the Playwright fallback** — `browser-cookie3==0.20.1` can raise `RuntimeError("Failed to decrypt the cipher text with DPAPI")`, but `auth/cookies.py` only converts `BrowserCookieError` into the `PermissionError` caught by `get_chrome_cookie_snapshot`. Handle the DPAPI runtime failure as a fallback-eligible decryption error and add a regression test.

2. **Cookie scope metadata is flattened** — `_name_value_cookies()` converts cookies into `dict[str, str]`, losing domain, path, Secure, and duplicate-name semantics before handing them to httpx. Preserve an appropriately scoped cookie jar (or otherwise retain those attributes) so the probe cannot send cookies Chrome would exclude for this URL.

3. **The new HTTP fast path lost transient retries** — `verify_flow_session` retries network failures and 429/503/504, while `verify_flow_profile` performs one `client.get()`. Reuse equivalent retry behavior so a temporary response does not reject an otherwise successful login.

GitHub also currently reports the PR as `DIRTY`; please resolve the merge conflicts with `develop`.

No need to revisit the already-resolved items above.

ffroliva:
## Council re-review @ `5d59c4c` → fix pushed (`a58cae0`)

Thanks @3mora2 — re-ran the multi-dimensional council (correctness / quality / security / tests / memory / auth-strategy). **All three of the previous round's must-fixes are genuinely resolved:**
- ✅ `verify_flow_profile` is wired into production (`real_chrome.py` calls it; the old `verify_flow_session` has no production callers left)
- ✅ CHANGELOG no longer oversells — accurately describes the fast path + marker-gated fallback
- ✅ Chrome-strategy marker guard is in place on the cookie-decrypt fallback
- ✅ D3 security: GREEN (cookie values never logged/persisted/leaked; TLS verify on; `follow_redirects=False`; scope is only auth/test/dep files — no hidden workflow/script changes)
- ✅ D4 tests: GREEN (strategy mocks migrated; `e2e_auth` zero-credit cases added per the prior guidance; retry/DPAPI/SSO-only all covered, sleeps mocked)

**Two new regressions surfaced** (flagged independently by the correctness and auth-strategy reviewers) in the marker lifecycle the wiring introduced — I've pushed a fix straight to the branch (`a58cae0`, maintainer edit) so this isn't another round-trip:

1. **A previously-verified chrome profile was downgraded by a transient failure.** The marker was unlinked on *any* non-authenticated outcome, including `VERIFICATION_ERROR` (a transient network blip after the retries). That deletes the marker on a profile that was fine → `channel_for_profile` stops returning `chrome` → `FlowApiClient` launches bundled Chromium on a real-Chrome profile, the exact thing the marker exists to prevent.
2. **Interrupt safety.** `except Exception` missed `KeyboardInterrupt`/`asyncio.CancelledError` (both `BaseException`), so Ctrl+C during verification left the freshly-written marker on a never-verified profile.

**Fix:** snapshot whether the marker pre-existed, roll back only the *speculative* write, and do it in a `finally` so interrupts are covered. A legitimately pre-existing marker now survives a transient failure. Added `test_real_chrome_preserves_preexisting_marker_on_transient_failure`. Local gates green (ruff, `pyright src` 0 errors, auth suite 49 passed).

### Remaining nice-to-haves (non-blocking, your call)
- `cookies.py` `except RuntimeError` is broader than the DPAPI case — consider gating on `'DPAPI' in str(exc)` so unrelated `RuntimeError`s don't silently route to the fallback.
- `verify_flow_profile`'s httpx client has no explicit timeout while `verify_flow_session` uses 15 s — pass `timeout=15.0` for the parity the docstring claims.
- `get_chrome_cookies` (public) appears to have no callers — drop it, or keep only `get_chrome_cookie_snapshot`.

Once CI is green I'll dismiss the changes-requested review and this is mergeable.

ffroliva (approved):
Re-reviewed at 9b93cfc3. All council must-fixes resolved, the two marker-lifecycle regressions fixed (a58cae0), and all three nice-to-haves addressed (DPAPI-gated catch, httpx timeout=15.0 parity, dropped unused get_chrome_cookies). CI green; e2e_auth suite (3 cases incl. live cookie-decrypt + Playwright fallback) verified green against a live authenticated profile. LGTM.

PR #219, opened 2026-06-29
fix: resolve Chrome profile and T2I extra-image generation bugs

C1ph3r404:
## Summary

1. Fix Playwright chrome channel validation in `browser_manager.py`:
   - `is_chrome_available()` was returning True for system Chromium, causing Playwright to request `channel="chrome"` which expects real Google Chrome.
   - Replaced with a strict check against Playwright's actual expected Chrome paths so it falls back to bundled Chromium correctly.

2. Fix image generation settings selector in `ui_automation.py`:
   - `GEN_SETTINGS_BUTTON_SELECTORS` lacked the `aria-haspopup='menu'` attribute and missed the `crop_original` icon.
   - Without it, the selector failed to find the panel after switching to image mode, leaving Flow's default generation count (2) active while the CLI only expected 1.

3. Warn on silent extra image generation in `client.py`:
   - If Flow generates more images than requested (e.g. 2 instead of 1 due to the settings panel miss), `_drive_image_generation` silently discarded the extras.
   - A warning is now logged showing the extra media IDs, alerting users to the extra credit consumption while preserving the single-image API contract.

### DIFF
--- a/src/gflow_cli/api/client.py
+++ b/src/gflow_cli/api/client.py
@@ -1310,13 +1310,34 @@ class FlowApiClient:
         constructing a :class:`GenerateImageRequest` with ``count>1`` and then
         invoking the single-image API still receive exactly one image (no
         silent discard).
+
+        When the transport returns more images than the requested count=1
+        (typically because the generation-settings panel was not found and
+        Flow used its own default, billing the account for the extra
+        generations), a warning is logged so the caller can investigate.
         """
+        import structlog as _structlog
+
+        _log = _structlog.get_logger(__name__)
         req_one: GenerateImageRequest = _dc_replace(req, count=1)
         images = await self._drive_images_generation(
             project_id=project_id,
             req=req_one,
             recaptcha_action=recaptcha_action,
         )
+        if len(images) > 1:
+            _log.warning(
+                "client.generate_image_extra_returned",
+                requested=1,
+                returned=len(images),
+                extra_media_ids=[img.media_name for img in images[1:]],
+                hint=(
+                    "Flow generated more images than requested — the generation-settings "
+                    "panel selector may have missed and Flow used its own default count. "
+                    "The extra image(s) were billed to your account but not saved. "
+                    "Use `gflow image t2i -n 2` to request and save multiple images explicitly."
+                ),
+            )
         return images[0]
 
     async def generate_image(
--- a/src/gflow_cli/api/transports/ui_automation.py
+++ b/src/gflow_cli/api/transports/ui_automation.py
@@ -386,15 +386,22 @@ WELCOME_SCREEN_SELECTORS = (
     "button:has-text('See what's new')",
 )
 
-
-# Generation settings trigger — the button shows the current ratio icon.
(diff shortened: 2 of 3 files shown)


ffroliva (changes requested):
Thank you for this @C1ph3r404 — both root causes are real and your fixes are correct. We ran a multi-dimensional council review and the three changes hold up:

- **`channel_for_profile()` gating on a Playwright-specific Chrome check** correctly fixes the `Chromium distribution 'chrome' is not found at /opt/google/chrome/chrome` failure on Chromium-only hosts — `is_chrome_available()` was too permissive there.
- **Requiring `button[aria-haspopup='menu']`** on the gen-settings selector fixes the mis-click that skipped the count panel and let Flow bill its own default count.
- The **`generate_image_extra_returned` warning** is a good observability add.

**Council verdict: 🟡 YELLOW** — correctness is sound; a few things to address before it can land:

1. **CHANGELOG** — the `[Unreleased]` section is empty; add a `### Fixed` block for the three bugfixes.
2. **DRY the selector** — `GEN_SETTINGS_BUTTON_SELECTORS` is now byte-identical to `MODE_SWITCH_TRIGGER_SELECTORS` (same button). Alias it (`GEN_SETTINGS_BUTTON_SELECTORS = MODE_SWITCH_TRIGGER_SELECTORS`) so the two call sites stay a single source of truth.
3. **Narrow the Linux probe** — `_is_playwright_chrome_channel_available()` should check only `/opt/google/chrome/chrome` on Linux. The extra `/usr/bin/google-chrome` / `/usr/bin/google-chrome-stable` entries would report the channel as available on hosts where Playwright's `channel="chrome"` still can't resolve it — re-triggering the very error this PR fixes.
4. **Tests** — add unit coverage for (a) the `generate_image_extra_returned` warning firing when 2 images are returned, and (b) `_is_playwright_chrome_channel_available()` returning False on a Chromium-only host.

Two notes for context:
- This PR is **behind `develop`** (it predates the merged #225/#224), so it currently shows `mergeable: unknown` and needs a rebase.
- It is **not** the fix for #222 — that turned out to be a `--password-store=basic` login/generation asymmetry, resolved separately in #225. This PR fixes a distinct Chromium-only channel bug + the T2I over-generation bug.

**We've co-landed your work** in #226 (rebased on current `develop`, with the four items above applied) and **live-verified** the T2I path on a real-Chrome host — the gen-settings panel is now found, count locks to 1, and no extra image is generated. You're credited as co-author on that commit. We took this route because the fix needs verification on a host with Google Chrome proper, which a Chromium-only environment can't do. Closing this in favor of #226 once it merges — but huge thanks for the diagnosis and the fix. 🙏

ffroliva:
Co-landed as #226 (now merged into `develop`) — rebased on current develop, all four council must-fixes applied, and live-verified on a real-Chrome host (gen-settings panel found, count locks to 1, no extra image generated). You're credited as co-author on the landing commit. Thank you again for the diagnosis and the fix, @C1ph3r404 — closing this in favor of #226. 🙏

Issue #222, opened 2026-06-29
macOS: generation runs on bundled Playwright Chromium despite .gflow_browser_strategy=chrome →   401 AuthExpiredError on createProject


gunalak:
Environment
  - gflow-cli 0.22.0 (uv tool install gflow-cli)
  - macOS (Apple Silicon)
  - System Google Chrome installed; Google AI account with Flow access + credits (Flow web UI works fine)

  Steps to reproduce
  1. gflow auth login --profile <name> --browser chrome — signed in fully; the Flow editor and my projects
  loaded; closed the browser. Verified OK:
  [info] auth_flow_session_verified  source=chrome strategy=chrome user_email=<redacted>
  [OK] Flow session verified (<redacted>).
  Session saved. Profile dir: .../profile_<name>

  2. gflow image t2i "a hot air balloon over Tokyo at sunrise" — opens a browser that is not logged in,
  then fails:
  [info]  browser.engine_selected  engine=playwright
  [info]  ui_automation.setup_shared_page
    Creating project...
  [error] error_raised  error_class=AuthExpiredError
          problem={... 'status': 401, 'route': 'https://labs.google/fx/api/trpc/project.createProject'}
  Authentication expired: HTTP 401

  Expected: generation reuses the authenticated Chrome session (system Chrome channel) and createProject
  succeeds.

  Actual: generation launches a logged-out browser on the playwright engine → 401.
  
  Diagnostics
  - gflow auth status:
  Profile '<name>' is configured.
    exists: True
    cookies_present: True
    cookies_path: .../profile_<name>/Default/Cookies
    browser_engine: playwright
  - Marker file is correct:

  $ cat .../profile_<name>/.gflow_browser_strategy
  chrome

  Hypothesis: On macOS the generation step uses bundled Playwright Chromium instead of the system Chrome
  channel, even though .gflow_browser_strategy=chrome. Bundled Chromium can't decrypt cookies written by
  real Chrome (macOS Keychain "Chrome Safe Storage"), so it gets a logged-out context → 401. auth status
  reporting browser_engine: playwright despite the chrome marker suggests the marker isn't being honored
  at generation time in 0.22.0.

  Questions
  1. Is the marker → system-Chrome-channel selection for the ui_automation transport present in 0.22.0, or
  only on develop? 
  2. Is there a macOS-specific config/env to force channel=chrome for generation?
  3. Happy to test a develop build or capture more logs / run with a debug flag — just let me know what's
  useful.

[...]

PR #223, opened 2026-06-29
fix(browser): diagnose + fail loud on chrome→bundled strategy downgrade (Refs #222)

ffroliva:
## Summary

Ships the **diagnosability + fail-loud** half of #222 (macOS: generation runs on bundled / undecryptable Chromium despite the `chrome` strategy → HTTP 401). This PR does **not** claim to fix the macOS 401 — see *Seeding attempt (removed)* — it makes the failure observable and loud, and records the localized root cause.

## Changes

**Diagnostics**
- `browser_manager.chrome_strategy_requested()` — detects `marker == "chrome"`, distinguishing a legit no-marker bundled launch from a silent downgrade.
- `browser_manager.resolved_chrome_binary()` — public, exception-free Chrome-path resolver.
- `FlowApiClient._log_and_guard_launch()` — logs `client.persistent_context_launch` (channel / `chrome_executable` / `cookies_db_path` / `password_store_basic` / `launch_args` / platform) and **raises `ConfigurationError` on macOS** when the chrome strategy silently downgrades to bundled Chromium; **warns** (`client.chrome_strategy_downgraded`) on other platforms.
- `FlowApiClient._log_context_cookie_state()` — logs `client.context_cookie_state` (presence / count / expiry of the Flow session cookie — **never values**, safe to paste publicly): the load-vs-send discriminator. Pure observability, never raises.

**Docs**
- `KNOWN_ISSUES.md`: macOS generation-401 entry (root cause, the shipped diagnostics, why seeding was disproven). `CHANGELOG.md`: `[Unreleased]` fail-loud + diagnostics notes.

## Seeding attempt (removed)

An earlier revision seeded the session cookie into the context by re-reading it via `browser_cookie3` at launch. The reporter's macOS run **disproved** it: while the generation context holds the profile lock, `browser_cookie3` *also* fails (`PermissionError`), so there is nothing to inject — still 401. That commit was dropped; #222 stays **open** for a real fix. (Trail: issue #222 comments.)

## Review

`/code-review` (4 parallel finders: correctness / cleanup-altitude / conventions / tests) — **no correctness bugs, no dangling refs** to the removed seeding code. Applied: accurate fail-loud message (it previously said "system Chrome was not found" while a real `chrome_executable` path was logged), plus coverage for the launch-event fields, the off-macOS warning, and the never-raise probe contract.

## Test plan
- [x] `ruff check` / `ruff format --check` — clean
- [x] `pytest tests/api/test_client_launch_kwargs.py tests/test_browser_manager.py` — 69 passed
- [x] `check_doc_links.py` / `check_repo_hygiene.py` — green
- [ ] macOS reporter to attach `client.context_cookie_state` logs on a fresh run (#222)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Extra-image observability**: `generate_image()` now logs a
   `client.generate_image_extra_returned` warning when the transport returns more images
   than the requested `count=1`, so silent over-generation is surfaced to the user (#219).
+- **macOS generation 401 — fail loud (#222)**: a `chrome`-strategy profile that is
+  silently downgraded to bundled Chromium at generation launch now raises instead of
+  running logged-out, so the macOS auth failure surfaces at its cause rather than as an
+  opaque later `401`.
+
+### Diagnostics
+
+- **Persistent-context cookie state (#222)**: generation launch now logs the resolved
+  `cookies_db_path` (`client.persistent_context_launch`) and the launched context's own
+  cookie state (`client.context_cookie_state`: `flow_session_cookie_present` / count /
+  expiry — **never values**, safe to paste publicly). This splits a cookie-**load**
+  failure (the macOS persistent context cannot decrypt the profile's cookies) from a
+  server-side rejection. Root cause of the macOS `401` is localized but not yet fixed —
+  see [KNOWN_ISSUES.md](KNOWN_ISSUES.md) and
+  [#222](https://github.com/ffroliva/gflow-cli/issues/222).
 
 ## [0.22.0] — 2026-06-28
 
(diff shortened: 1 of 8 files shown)


ffroliva:
Rebased onto `develop` (now contains #225, the `--password-store=basic` fix). The two #222 fixes are **complementary and now coexist cleanly** — git auto-merged because they touch different regions:

- **#225** (`_persistent_context_kwargs`): stops suppressing `--password-store=basic` so generation is symmetric with login (fixes the basic-vs-keychain decrypt mismatch).
- **#223** (this PR): `_log_and_guard_launch` + `chrome_strategy_requested`/`resolved_chrome_binary` — logs the resolved channel/executable/cookies-db at launch and fails loud on a silent chrome→bundled downgrade (the *other* macOS failure mode).

Together: #225 fixes the password-store mechanism, and #223's diagnostics will confirm on the reporter's Mac **which** mechanism actually fired (channel=null → downgrade guard; channel=chrome → password-store path). 

Verified locally on the rebased branch: `tests/api/test_client_launch_kwargs.py` + `tests/test_browser_manager.py` → 58 passed; ruff + `pyright src/gflow_cli/api/client.py src/gflow_cli/browser_manager.py` clean. `Refs #222` — leave open until the reporter confirms on macOS.

PR #225, opened 2026-06-29
fix(api): keep --password-store=basic at generation to fix macOS 401 (#222)

ffroliva:
## Summary

Fixes the macOS `401 AuthExpiredError` on `project.createProject` reported in #222, where `gflow image t2i` opened a logged-out browser despite a verified `auth login`.

**Root cause — a `--password-store` asymmetry between login and generation:**
- `auth login` (`auth/real_chrome.py:69`) launches Chrome **with** `--password-store=basic`, sealing the profile's cookies with Chrome's *basic* store (not the macOS Keychain).
- Generation (`api/client.py:_persistent_context_kwargs`) listed `--password-store=basic` in `ignore_default_args`, **stripping** the flag Playwright adds by default — so that Chrome used the macOS Keychain ("Chrome Safe Storage") to read the cookies.
- basic-sealed cookies read under the Keychain → decrypt failure → logged-out → `401`. macOS/Linux-only; Windows uses DPAPI and ignores the flag (why it never reproduced on Windows, and why `auth login`/`auth status` — which don't hit this path — succeeded).

**Fix:** stop suppressing `--password-store=basic` at generation so it is symmetric with login.

## ⚠️ Verification status — INCOMPLETE (needs macOS)
- [x] Unit: launch-kwargs pin test updated + regression assertion added (`tests/api/test_client_launch_kwargs.py`) — 2 passed
- [x] `ruff check` + `ruff format --check` clean; `pyright src/gflow_cli/api/client.py` 0 errors
- [ ] **Live `gflow image t2i` on macOS confirming the 401 is gone** — NOT DONE
- [ ] Confirm `auth login` (still basic) + generation (now basic) round-trip on macOS

This **cannot** be verified on CI (headless Linux) or on Windows — #222 is a macOS-Keychain-specific failure. Held as **draft** pending a maintainer macOS run. Auth-surface change — please review the symmetry argument before promoting.

`Refs #222` (not `Closes`) — do not auto-close until macOS-verified.

### DIFF
--- a/src/gflow_cli/api/client.py
+++ b/src/gflow_cli/api/client.py
@@ -333,10 +333,16 @@ class FlowApiClient:
             "locale": "en-US",
             "extra_http_headers": {"Accept-Language": "en-US,en;q=0.9"},
             "channel": channel_for_profile(self.profile_dir),
+            # Do NOT suppress Playwright's default --password-store=basic here.
+            # auth login (auth/real_chrome.py) seals the profile's cookies with
+            # the *basic* store; if generation lets Chrome fall back to the OS
+            # keychain (macOS "Chrome Safe Storage") the basic-sealed cookies
+            # can't be decrypted -> logged-out context -> 401 on createProject.
+            # Keeping the flag (via Playwright's default) makes both paths
+            # symmetric. See issue #222.
             "ignore_default_args": [
                 "--enable-automation",
                 "--no-sandbox",
-                "--password-store=basic",
             ],
             "args": ["--disable-blink-features=AutomationControlled", "--disable-dev-shm-usage"],
         }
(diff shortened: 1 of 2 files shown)


ffroliva:
**`/code-review` (high-effort) — no correctness findings.** Verified the fix is on the path the issue hit: the #222 log shows `ui_automation.setup_shared_page`, which is the **shared-page** branch (`ui_automation.py:722-728`) where `FlowApiClient` owns the launch via `_persistent_context_kwargs` — exactly the code changed here. The standalone ui_automation path (`ui_automation.py:774`) already passed `--password-store=basic`; the FlowApiClient shared-page path was the lone inconsistent outlier, which is why the 401 surfaced only there.

Two low-severity follow-ups (non-blocking, not hotfix scope): the scattered `--password-store=basic` across ~8 launch sites could be centralized; and `scripts/dev/spike_patchright.py:113` (a dev replica) now drifts.

**Merge decision:** merging to `develop` now (regression risk near-zero — no-op on Windows, symmetric with login/standalone elsewhere). `develop` is not released; **#222 stays open** (`Refs`, not `Closes`) pending the reporter's macOS `gflow image t2i` confirmation before this ships in a release.

Issue #170, opened 2026-06-11
`--reference-entity` fails with RuntimeError on non-Portuguese locales — context-menu selector hardcodes pt-BR text "Incluir no comando"


papushin7987:
## Environment
- gflow-cli 0.15.1 (uv tool install), Python 3.13, Ubuntu 24.04
- Flow account UI language: **Russian** (note: the account language wins even though the CLI opens project URLs with `?hl=en`)
- `gflow auth login` OK, plain `t2i` / `i2v` / `character create` / `character list` all work

## Repro
```bash
gflow image t2i "test" \
  --project <project-id> \
  --reference-entity <entity-id> --reference-entity-name Lukas \
  --aspect 16:9
```
Result: `RuntimeError` right after `ui_automation.count_setter_completed`:

```
character 'Lukas' (<entity-id>) context-menu 'Incluir no comando' did not appear after right-click.
```

## Root cause
`api/transports/ui_automation_video.py` hardcodes the pt-BR menu caption:

```python
PICKER_INCLUDE_BUTTON = "button:has-text('Incluir no comando')"
PICKER_CONTEXT_INCLUDE = (
    "[role='menuitem']:has-text('Incluir no comando'), button:has-text('Incluir no comando')"
)
```

The right-click context menu on a Personagens tile is locale-dependent. On a Russian-locale account the item is **«Добавить в запрос»**, on English presumably "Add to prompt" — so the locator never matches and `_attach_character_entities` raises.

All neighbouring selectors are already locale-free (ligature icons / ids), e.g. `PICKER_PERSONAGENS_TAB` matches `i.google-symbols:text-is('accessibility_new')` — only these two constants still rely on visible text.

## Suggested fix (verified live on a Russian-locale account)
The context-menu item carries a `google-symbols` ligature icon `add`, so an icon-based locator works for every locale. Multi-locale texts kept as primary to stay conservative:

```python
PICKER_INCLUDE_BUTTON = (
    "button:has-text('Incluir no comando'), button:has-text('Добавить в запрос'),"
    " button:has-text('Add to prompt')"
)
PICKER_CONTEXT_INCLUDE = (
    "[role='menuitem']:has-text('Incluir no comando'), button:has-text('Incluir no comando'),"
    " [role='menuitem']:has-text('Добавить в запрос'), button:has-text('Добавить в запрос'),"
    " [role='menuitem']:has-text('Add to prompt'), button:has-text('Add to prompt'),"
[...]

PR #237, opened 2026-07-04
feature: Support remote image UUIDs in gflow_generate_video

C1ph3r404:
## Summary
Allows passing Flow UUID strings (e.g. from generated images) into video generation requests for both I2V (start_image, end_image) and R2V (reference_images).

Changes:
- api/video.py: Add `start_image_ref_name`, `end_image_ref_name`, and `ref_names` to `GenerateVideoRequest` and update validations.
- mcp/tools.py: Detect UUID inputs for video boundaries, bypassing local path validation. Resolve UUIDs to display names via DB lookup during task enqueue.
- worker/daemon.py: Extract resolved display names from payload into the video request.
- api/transports/ui_automation_video.py: Implement `_attach_remote_frame` to attach I2V remote frames via UI search modal, mirroring R2V remote references.
- data/repository.py: Adjust formatting of `get_asset_by_any_id` SQL queries to pass lint.

Why:
Users can now pipe the output of image generations (UUIDs) directly into video generation without needing to download and re-upload local files, streamlining end-to-end automation.

## Validation

- [x] Documentation updated or explicitly marked not applicable
  *(Not applicable: no new CLI commands or flags were added, this transparently extends the existing inputs to accept UUIDs, which is already how agents interact with image endpoints.)*
- [x] `uv run python scripts/ci/check_doc_links.py` (All 23 files passed)
- [x] `uv run ruff check src tests` (Fixed formatting in video.py and repository.py)
- [x] `uv run pyright src` (Passed; only pre-existing untyped third-party errors remained)
- [x] Relevant pytest command: `uv run pytest tests/api/transports/ tests/mcp/ -q` (Ran and passed to guarantee the original local-file upload behavior for frames and references continues to work flawlessly)
  *(Note: No new automated UI tests added due to the live headed-browser requirement for interacting with the Flow search bar. Trusted live observation and backwards-compatibility tests instead.)*

### DIFF
--- a/src/gflow_cli/api/transports/ui_automation_video.py
+++ b/src/gflow_cli/api/transports/ui_automation_video.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import random
 import time
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
@@ -295,6 +296,7 @@ UPLOAD_MEDIA_BUTTON = (
 UPLOAD_MEDIA_BUTTON_TEXT = "[role='dialog'][data-state='open'] button:has-text('Upload media')"
 # 'Add to Prompt' has no stable string anchor; selected structurally (the lone
 # iconless button) at the call site. This scope is the open media dialog.
+ADD_TO_PROMPT_BUTTON_TEXT = "button:has-text('Add to Prompt')"
 ADD_TO_PROMPT_DIALOG = "[role='dialog'][data-state='open']"
 # R2V references mode has NO Start/End slots — references are added via the
 # only button[aria-haspopup='dialog'] in the editor: a 'Create' button carrying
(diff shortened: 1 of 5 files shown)


ffroliva:
Thanks for this — piping generated-image UUIDs straight into video generation (instead of a download/re-upload round-trip) is a real gap and a good direction for the agentic/MCP workflow. Found a couple of issues while reviewing at `8cf02e8` that block merge:

**1. `AssetLookup` is missing fields this PR relies on — will raise `TypeError` at runtime**

`src/gflow_cli/data/repository.py`'s `_hydrate_asset_lookup` now constructs:
```python
AssetLookup(
    ...,
    flow_workflow_id=row["flow_workflow_id"],  # line 237
    ...,
    metadata_json=json.loads(row["metadata_json"]) if row["metadata_json"] else {},  # line 240
)
```
but `src/gflow_cli/data/models.py:217-223`'s `AssetLookup` dataclass only declares `id, profile_name, flow_project_id, flow_media_id, kind, local_files` — no `flow_workflow_id` or `metadata_json`. As written this throws `TypeError: __init__() got an unexpected keyword argument 'flow_workflow_id'` on every call to `get_asset_by_flow_media_id`, the new `get_asset_by_any_id`, and `find_assets_by_flow_media_id` — i.e. on the core new UUID-resolution path this PR adds. `AssetLookup` needs those two fields added (mirroring `SeedImage`'s `flow_workflow_id: str | None` right below it, plus a `metadata_json: JsonObject` field).

This is why it wasn't caught locally: the validation notes mention `tests/api/transports/ tests/mcp/`, which don't exercise `repository.py` directly, and CI itself never got that far — `test (3.11/3.12/3.13)` are all failing at the `ruff format --check` gate (`Would reformat: src/gflow_cli/api/video.py`) before pytest even runs. Worth a `uv run ruff format src tests` pass regardless.

**2. Unescaped display name in the Playwright selector**

`src/gflow_cli/api/transports/ui_automation_video.py:1242` and `:1421` both build:
```python
tile = page.locator(f"[role='option']:has-text('{name}')").first
```
`name` is either a stored `display_name` or (per the fallback in `mcp/tools.py`'s `_resolve_ref`) the original generation prompt text. Either commonly contains an apostrophe (contractions, quotes), which will break this single-quoted `:has-text()` selector. Worth escaping/quoting `name` (or using a different matching strategy) before interpolating it in.

Happy to take another pass once these are addressed.

ffroliva:
Follow-up after a fuller multi-dimensional pass (correctness, code quality, security, tests, UI/locale, data-layer) — a few more confirmed issues beyond the two I flagged earlier, so you have the complete picture before starting fixes:

**3. Same-shaped bug also breaks image generation, not just video**

`src/gflow_cli/worker/daemon.py:296` passes `ref_names=ref_names` into `GenerateImageRequest(...)`:
```python
refs = tuple(ImageRef(r) for r in payload.get("refs", []))
ref_paths = tuple(Path(p) for p in payload.get("ref_paths", []))
ref_names = tuple(payload.get("ref_names", []))
...
return GenerateImageRequest(..., ref_names=ref_names, ...)  # line 296
```
`GenerateImageRequest` doesn't have a `ref_names` field — only `GenerateVideoRequest` gained one in this PR. This throws `TypeError: GenerateImageRequest.__init__() got an unexpected keyword argument 'ref_names'. Did you mean 'ref_paths'?` on the image-generation path. Reproduced by actually running the suite (see #6 below).

**4. New "Add to Prompt" selector reintroduces a locale bug this file already fixed twice**

`src/gflow_cli/api/transports/ui_automation_video.py:299`:
```python
ADD_TO_PROMPT_BUTTON_TEXT = "button:has-text('Add to Prompt')"
```
This is inserted directly below this file's own comment (two lines above) explaining that "'Add to Prompt' has no stable string anchor; selected structurally... at the call site" — because a hardcoded English string here previously caused a ~34s hang / broke every non-English account (issues #170, #56). The existing tiered/icon-first pattern for exactly this action (`_resolve_include_action` + `PICKER_INCLUDE_BUTTON`/`PICKER_CONTEXT_INCLUDE`) is already in this file and wasn't reused. Relatedly, `_attach_remote_references` (unlike `_attach_remote_frame`) has no dialog-closed check after clicking, so on a locale mismatch it logs `remote_reference_attached` even though nothing was actually attached — a silent false-success rather than a visible failure.

**5. No test coverage on any new code path, plus two now-failing pre-existing tests**

A grep across the whole `tests/` tree for `_attach_remote_frame`, `_attach_remote_references`, `get_asset_by_any_id`, `_resolve_ref`, `ref_names`, `start_image_ref_name`, `end_image_ref_name` returns zero matches — none of the new functionality is exercised anywhere. Separately, `api/video.py`'s R2V validation error message changed ("reference_images or reference_entities" → "...reference_images, ref_names, or reference_entities") without updating the two existing tests that regex-match the old wording (`tests/api/test_video_request.py::test_r2v_requires_images_or_entities`, `tests/api/test_video.py::TestGenerateVideoRequest::test_r2v_requires_a_reference_image`).

**6. Running the actual suite confirms 5 failures**, not the "ran and passed" the PR body claims:
```
PYTHONPATH=<repo>/src .venv/Scripts/python.exe -m pytest tests/api/test_video_request.py tests/api/test_video.py tests/api/transports/test_ui_automation_video.py tests/mcp/ -q
5 failed, 188 passed
```
Two of the five are the stale-regex tests above; the other three trace back to the `AssetLookup`/`GenerateImageRequest` bugs (#1 from my earlier comment, and #3 above).

**7. Minor:** no `CHANGELOG.md` entry for this feature (`[Unreleased]` is currently empty), and an unresolvable UUID currently surfaces as a raw ~120s Playwright timeout rather than a clear "not found in your asset catalog" error.

None of this changes the direction — the underlying idea is still worth having — but there's a bit more to close out than the first comment suggested.

ffroliva:
Thanks again for this @C1ph3r404 — the direction is genuinely useful, so rather than let it stall I've taken it forward in #245 (your commit preserved, with a follow-up that closes the 7 review points: the AssetLookup/GenerateImageRequest TypeErrors, apostrophe-safe picker selectors, locale-safe Add-to-Prompt + dialog-closed guard, the unresolvable-UUID fast error, and test coverage). Full suite green at 89% coverage. #245 credits you as the original author. Closing this in favor of that one — and thanks for the contribution.

PR #253, opened 2026-07-06
fix(i2i): properly resolve and attach remote UUID refs in UI automation

C1ph3r404:
fix(i2i): properly resolve and attach remote UUID refs in UI automation

Fixes the regression where image tasks with remote UUID references were not
resolved to display names, causing the UI automation transport to ignore them
and submit standard text-to-image prompts.

- Updates _resolve_payload_ref_names to resolve image refs on a best-effort basis (satisfying PR #245).
- Adds ref_names to GenerateImageRequest.
- UI automation transport now properly searches and attaches UUID refs via the media dialog.

The display_name was previously ignored when parsing the batchGenerateImages response, causing the CLI to rely on the prompt text when searching for generated images in the remote UI picker.

Changes:
- api/dto.py: Extract the 'workflows' array from the response and map workflow IDs to their display names, injecting the 'display_name' into the GeneratedImage model.
- data/recorder.py: Save the extracted 'display_name' into the asset's metadata_json during database upsert.

### DIFF
--- a/src/gflow_cli/api/dto.py
+++ b/src/gflow_cli/api/dto.py
@@ -128,6 +128,7 @@ class GeneratedImage:
     fife_url: str  # CDN URL — usually expires after ~6 hours
     dimensions: tuple[int, int]  # (width, height)
     media_generation_id: str | None = None
+    display_name: str | None = None
 
     @property
     def is_signed_url(self) -> bool:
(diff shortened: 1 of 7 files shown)


ffroliva:
Thanks @C1ph3r404 — excellent root-cause work. You correctly found that the `displayName` was in the `workflows[]` array all along and the parser was ignoring it (nicely consistent with the existing `AssetInfo.from_upload_response` extraction). That's the real reason generated assets had no catalogued display name.

**Context on the conflict:** while this was in flight, **v0.25.0** shipped a rework of the **video** i2v/r2v path. We found generated media don't appear in Flow's frame picker at all (empty grid regardless of name), so the UUID is now resolved to its **local file** and attached via the proven upload path. That renamed `_resolve_payload_ref_names`→`_resolve_payload_refs` and removed `_resolve_ref_name` — which is what this PR conflicts with, and its video-path changes would regress that live-verified fix.

**Plan:** rather than have you rebase onto code that moved out from under you, we'll carry your foundation forward and land it **with credit to you**:
- Your `display_name` extraction (`dto.py` + `recorder.py`) goes in as-is.
- We'll build the **image-i2i** UUID-ref support on top, live-verifying whether the image reference picker actually surfaces generated media before choosing picker-vs-local-upload (if it behaves like the video picker, image i2i will reuse the local-upload path too).

Targeting **v0.26.0**. I'll link the follow-up PR here and close this one as superseded once it's up. Thanks for opening the door on this — the `workflows[]` display-name find is the key insight.

ffroliva:
Follow-up is up: #255 lands your `display_name`-from-`workflows[]` extraction (credited to you) onto the post-0.25.0 develop. The image-i2i UUID **attach** builds on it next, with a live e2e verification — and good news, I live-confirmed the image `add_2` reference picker DOES surface generated media (tile keyed by the media UUID, with the Flow display name), so the prefer-existing (select-in-place, no re-upload) path is viable. Will close this PR once #255 merges. Thanks again — the `workflows[]` find was the key.

ffroliva:
Landed: #255 is merged to develop — your `display_name`-from-`workflows[]` extraction is now in, credited to you (Co-Authored-By). Closing this as superseded. The image-i2i UUID **attach** builds on it next (prefer-existing: select the existing Flow asset in the picker rather than re-upload a duplicate — live-confirmed the image picker surfaces generated media). Thanks again @C1ph3r404 — the `workflows[]` find was the unlock.

PR #268, opened 2026-07-08
fix(agentic): reliable agentic binding when -i instructions are used (#267)

### LINKED ISSUE #267: Agentic binding is unreliable on fresh projects — -i instructions can silently no-op
## Summary

Surfaced during the v0.28.0 agentic-instructions work. On a fresh (REST-created) project, the UI-cohort probe (`factory.get_ui_driver` / `detect_ui_mode`) and `_force_agent_mode` do not reliably bind the **agentic** driver — observed roughly ~50/50 across live e2e runs with `GFLOW_CLI_FORCE_AGENT_UI=1`. When it binds **classic**, the classic driver has no instruction surface, so `gflow image t2i "…" -i "…"` **silently produces an image with the instructions ignored** (now warned, but still a no-op that costs a credit).

## Evidence

- `tests/e2e/test_live_agentic_instructions.py` (`-m e2e_image`, `GFLOW_CLI_FORCE_AGENT_UI=1`) bound `agentic` on some runs and `classic` on others for the *same* profile/project shape. The classic run failed the test's `ui_driver.bound mode==agentic` assertion.
- Root cause candidates:
  - `ui_automation._force_agent_mode` clicks the composer Agent toggle then confirms via the `tune` indicator with only a ~1.2s settle — the confirmation races the render, returns `False`, and the subsequent `detect_ui_mode` probe then defaults to `classic`.
  - Independent of the force flag: a fresh project may open the *classic* composer first, and `detect_ui_mode`'s 8s poll can resolve `classic` (crop_* present) before the agentic surface renders.

## Impact

- **Production, no force flag:** an agentic-cohort account running `-i` on a fresh scratch project may intermittently bind classic → instructions dropped (a warning fires, but the generation still runs and spends a credit).
- Makes the agentic-instructions feature best-effort rather than reliable.

## Possible fixes (to investigate)

- Make `_force_agent_mode` await confirmation properly (poll for the `tune` indicator up to a real timeout; retry the toggle click once) and check its return value in `_generate_images_locked`.
- When instructions are present, hard-require an agentic bind: retry force+probe, and fail fast (typed error) rather than silently generating on classic.
- Prefer `--project <agentic project>` guidance for instructioned generations until binding is deterministic.

## Acceptance

- A deterministic agentic bind (or a fail-fast) when `-i` instructions are supplied, verified by a green live e2e across several consecutive runs.

ffroliva:
Closes #267. **Held for 0.29.0** (draft until live-verified).

## Problem
Instruction cards (`-i`) are an agentic-only surface, but binding was ~50/50 on fresh projects: `_force_agent_mode` clicked the Agent toggle then checked the `tune` indicator after a single fixed 1.2s wait, racing the render → the probe often bound classic and `-i` silently no-op'd (found during the v0.28.0 spike).

## Fix
- **`_force_agent_mode`** now polls for the `tune` indicator up to a real timeout and re-clicks once if it doesn't take (`_wait_selector_present`), returning an accurate result.
- **`_generate_images_locked`** forces the agentic composer whenever `request.instructions` is present (not only under `GFLOW_CLI_FORCE_AGENT_UI`), and warns if the force fails so instructions never silently vanish (idempotent no-op when already agentic).

## Tests
Unit: poll-succeeds-after-render, retry-then-fail, already-active-no-click, toggle-missing; cohort tests updated to drive the new force path. `ruff` / `pyright` / `test_ui_automation.py` (164) all green.

## Pending before un-draft / 0.29.0
- **Live determinism verification** (`scratch/probe_binding_267.py`, credit-free — create → force → probe cohort, repeated): confirm it binds agentic every run. Blocked right now on an expired auth session; will run after re-auth.

### DIFF
--- a/src/gflow_cli/api/transports/ui_automation.py
+++ b/src/gflow_cli/api/transports/ui_automation.py
@@ -212,6 +212,14 @@ AGENT_FORCE_ENV_VAR = "GFLOW_CLI_FORCE_AGENT_UI"
 AGENT_TUNE_INDICATOR_SELECTOR = "i.google-symbols:text-is('tune')"
 AGENT_EXPAND_BUTTON_SELECTOR = "button:has(i.google-symbols:text-is('expand_content'))"
 
+# _force_agent_mode: after clicking the Agent toggle, POLL for the ``tune``
+# indicator instead of a single fixed sleep — the agentic composer renders a
+# beat after the click and a fixed-delay check races that render, so the probe
+# bound classic ~50% of the time and instructions silently no-op'd (#267).
+_AGENT_ACTIVATE_TIMEOUT_S = 6.0
+_AGENT_ACTIVATE_POLL_MS = 300
+_AGENT_FORCE_MAX_CLICKS = 2
+
 # Self-contained, locale-independent triptych instruction for body generation.
 # Live-verified 2026-06-02: produces a consistent front/side/back body image in ONE
 # generation, seeded by the auto-attached face reference. We replace Flow's own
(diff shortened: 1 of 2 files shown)


ffroliva:
### Live binding verification (credit-free) — deterministic ✅

Ran `scratch/probe_binding_267.py` on `profile_ffroliva`: 4 fresh projects → navigate → force agent → probe cohort. **4/4 bound `agentic`.**

```
[0] project=dc659ae8  forced=False  bound=agentic
[1] project=c6c89ba4  forced=False  bound=agentic
[2] project=e85973df  forced=False  bound=agentic
[3] project=662a7ad5  forced=False  bound=agentic
SUMMARY: 4/4 bound AGENTIC — PASS
```

**Honest caveat:** every run logged `forced=False` + `agent_toggle_not_found` — i.e. this account currently renders fresh projects **agentic natively** (no classic→agentic toggle present), so `detect_ui_mode` binds agentic on its own and the `_force_agent_mode` *recovery* path this PR hardens wasn't exercised (I couldn't reproduce a classic bind to recover from today). The change remains a safe, unit-tested defensive improvement (idempotent when already agentic; poll+retry only when a toggle exists). If a classic bind recurs, the hardened force + the `-i`-triggered force + the fail warning are what catch it.

Follow-up worth noting: `_force_agent_mode`'s "already active" early-return checks only the `tune` indicator, while `detect_ui_mode` checks `tune`/`apps_spark_2`/`article_spark`/`edit_square`. Aligning that check to the full indicator set would let the force short-circuit correctly here instead of hunting for a non-existent toggle (minor; tracked for a follow-up).

Issue #283, opened 2026-07-10
Follow-ups from #281/#282 fix branch: attribution consolidation, stable-break hardening, picker scroll off-by-one


ffroliva:
Follow-ups deferred from the #281/#282 fix branch's final review (fix/281-agentic-media-attribution). None are release blockers; grouped here so they aren't lost.

## Hardening (#281 residual gap)
- **Stable-break condition in `await_images`**: break the poll loop only when the new-UUID set is stable across two consecutive scrapes at `== expected_count` (shrinks the exact-count lazy-render race documented in KNOWN_ISSUES; ~0.5s cost).

## Consolidation / consistency
- **Move `_verify_media_attribution` onto `OperationRecorder`** — it now exists in three justified copies (cli_image.py, image_batch.py, worker/daemon.py); all three call sites already hold a recorder, so a recorder method has zero layering objection.
- **Add `route=` to the new `MediaAttributionError` raises** (cli_image, image_batch, daemon) for provenance consistency with sibling error constructions.
- **Dedupe remediation text** between `MediaAttributionError._default_remediation` and the agentic raise-site detail.
- **Bulk `is_media_recorded` lookup** if batch limits ever grow past ~50 (currently one SQLite query per image).

## Picker (shared with entity path)
- **Scroll-loop off-by-one**: tile count is checked before each scroll, so a tile rendered by the final scroll is never re-checked — same pattern in `_select_existing_asset` AND `_find_picker_entity_tile`; fix both together.
- **Composite test**: display-name search fails → scroll then succeeds (the realistic production sequence; halves are tested separately today).

## Docs / semantics
- **Document that `MediaAttributionError` overrides `continue_on_error=True`** in the manifest batch (deliberate, test-pinned; one docstring sentence).
- **`_run_upscale` downloads with no attribution guard** — low risk (user names the source asset), noted for completeness.

ffroliva:
Additions from the pre-merge high-effort code review of PR #284:

- **Wire attribution defenses into the shell multi-prompt path** — `gflow run --config` and `gflow image t2i` (multi-prompt/--prompts-file/--stdin) route through `image_batch.run_image_batch` → `run_one_image_prompt`, which records no history and carries no layer-2/3 guard (the `_recorder`/`_profile_name` params are accepted but never threaded). Pre-existing gap, now noted in KNOWN_ISSUES; threading the recorder through would give these commands the same protection as t2i/i2i/manifest/daemon.
[...]

PR #290, opened 2026-07-11
feat(video): i2v frames accept an in-project asset UUID; typed upload-rejection error (#287)

ffroliva:
## Summary

- **UUID frame refs:** `gflow video i2v --initial-frame <PATH|UUID>` / `--end-frame <PATH|UUID>` (and the positional IMAGE). A media-UUID-shaped value selects the existing in-project asset in place — no duplicate upload — by routing the frame slot's media dialog through the same `_select_existing_asset` UUID picker the image `--ref` flow uses (including the #282 scroll/search fixes). New `start_image_ref_id`/`end_image_ref_id` DTO fields with structural validation (UUID shape via the new public `is_media_uuid`, per-slot exclusivity vs local path / display-name ref, T2V/R2V must-not-carry symmetry).
- **Typed upload rejection:** an `uploadImage` 4xx now raises `MediaUploadRejectedError` (**exit 27**, RFC 9457 `media-upload-rejected`, remediation: re-encode with `ffmpeg -q:v 2 -map_metadata -1` or reference the asset by UUID) instead of a bare `RuntimeError` surfacing as `Unexpected error.` (exit 1) — the s1.jpg HTTP-400 diagnosis round-trip from the 2026-07-11 pilot.
- A UUID that can't be located in the picker fails fast with `TransportTimeoutError` (exit 9) naming the slot, the UUID, and the --project hint. There is deliberately no local-file fallback: the caller asked for THIS asset.
- Docs: USAGE i2v section (PATH|UUID forms, UUID example, exit 27/9 notes) + CHANGELOG.

## Design notes

- No new flags: UUID-vs-path detection on the existing options mirrors `gflow image i2i --ref`'s `_classify_ref` precedent (issue #287 explicitly offered this shape).
- The display-name remote-frame path (`*_ref_name`, daemon-only) is untouched; ref_id outranks ref_name in `_attach_i2v_frames` routing.
- Reuses `MediaAttributionError`'s error-class pattern (#281): class + `EXIT_CODE_MAP` entry + `__all__`.

## Verification status

- [x] TDD: all new tests written first (red confirmed), then implementation to green
- [x] Unit: DTO validation (6 cases), CLI classification via CliRunner (4 cases incl. usage-error regression), transport routing + picker-miss error, typed 400 raise through the real `_upload_via_open_dialog` flow
[...]

Issue #281, opened 2026-07-10
Agentic image generation silently downloads a pre-existing project asset (wrong-media attribution)


ffroliva:
## Summary
Under the agentic driver, `gflow image t2i` can **silently download a pre-existing project asset instead of the freshly generated image** and report success. Ground truth becomes a different image — the worst failure class for any consistency workflow (cf. the resolved "`--model` silent no-op", PR #48).

## Live evidence (2026-07-10, v0.30.0, agentic cohort, nano-pro)
- Command: `gflow image t2i "<documentary portrait prompt>" --model nano-pro --aspect 9:16 --project f6caf027-… --out <dir>` (profile denon82), run twice for two characters.
- Both runs printed `Generated media was saved, but local history was not updated.` (`data.persistence_failed_after_success`).
- Both saved jpgs were **old logo/end-card assets** already present in the project — not the generated portraits.
- Downstream impact: 11 i2i scenes were anchored (`--ref`) to a logo; the logo rendered into one scene's guitar case and character identity drifted exactly like unanchored generation.

## Root cause (code reading)
1. `drivers/agentic.py: await_images` (~444) identifies "new" media by **set-difference of DOM-scraped UUIDs** vs a one-shot baseline snapshot. Any old project tile that lazy-renders *after* baseline (virtualised media panel) counts as "new".
2. `_build_generated_images` (~522) then takes `list(uuids)[:expected_count]` from an **unordered set** — an arbitrary pick, no correlation with the submitted generation ("set is unordered" per its own comment).
3. The warning is the smoking gun: `record_generated_images` hit `UNIQUE(profile_name, flow_media_id)` (`DataIntegrityError` → `DataStoreError` → warn-and-continue in `cli_image._record_generated_images_safe` ~925) — i.e. the "new" UUID was **already recorded** for this profile. The evidence of mis-attribution was downgraded to a cosmetic warning.

## Proposed fix (fail-fast, never silent-wrong)
- `await_images`: settle the baseline (union of two scrapes) and **raise on ambiguity** (`len(new_uuids) > expected_count`) instead of arbitrary slicing.
- `cli_image`: **pre-download attribution guard** — if any returned `media_name` is already recorded for the profile (`repository.get_asset_by_flow_media_id`), abort before download with a dedicated `MediaAttributionError`.
- `_record_generated_images_safe`: escalate `DataIntegrityError` (duplicate media id) to a hard error naming the suspect file; keep the warning only for generic `DataStoreError`.

[...]

PR #292, opened 2026-07-11
fix: #283 follow-ups — picker off-by-one, await_images stable-break, phantom screenshots, UUID consolidation

ffroliva:
## Summary

Closes the actionable remainder of #283 (verified item-by-item against develop before starting — see the issue comment trail):

- **F — picker scroll off-by-one (real latent bug):** `_select_existing_asset` checked `tile.count()` only BEFORE each scroll, so a tile rendered into the virtualised grid by the final scroll was never re-checked. Post-loop re-check added; `_find_picker_entity_tile` shares the loop shape but returns the locator unconditionally (caller waits for visibility), so it was unaffected — now documented in its docstring.
- **A — `await_images` stable-break (#281 residual race):** break only when the new-UUID set is identical across two consecutive scrapes at `expected_count`; a transient exact-count that then grows now raises the #281 `MediaAttributionError` instead of returning wrong media. Cost: ~one extra 0.5s poll per generation, as the issue estimated.
- **Phantom screenshot paths (live-observed 2026-07-11):** `_capture_debug_screenshot` (both copies) now returns `None` when capture fails; new `screenshot_clause()` helper; all 9 `Screenshot: {shot}` interpolation sites append conditionally — error messages can no longer point at files that were never written.
- **C — `route=`** on the recorder's three `MediaAttributionError` raises; **D — remediation dedupe** (agentic raise site relies on the class default).
- **UUID consolidation:** the four private `_UUID_RE` copies (cli_image, cli_instructions, image_upscale, mcp/tools) now delegate to `api.video.is_media_uuid` (#290's public helper). Pure refactor, no behavior change.
- **UX crumbs:** i2i/i2v share the 'media UUID' vocabulary; a bad value via the deprecated `--end-image` alias names `--end-image` in its usage error.

## Deliberately not done (with reasons)

- **E — bulk `is_media_recorded`:** the issue itself gates it on batch limits growing past ~50; they haven't.
- **I — upscale attribution guard:** issue marks it 'noted for completeness, low risk' (the user names the source asset explicitly).
- **H:** superseded — v0.31.0 inverted the behavior it asked to document (continue-on-error now honors collisions; documented at image_batch.py:876).
- **Error-path browser teardown leak** (Chrome lingers holding the profile after error exits; observed 3× live on 2026-07-11): needs its own investigation of the client teardown path — scoped out rather than blind-fixed; tracked in the closing comment.

## Verification status

- [x] TDD: three failing tests written first (picker final-scroll, stable-break ambiguity, phantom screenshot), then fixes to green
- [x] 667 passed across transports/cli-video/cli-image/recorder/mcp; instructions + upscale suites green after consolidation; ruff + format + doc links + hygiene + pyright 0 errors
- [x] The stable-break and picker paths are browser-free logic — the unit tests ARE the affected surface; no live gap
- [ ] Optional live spot-check: any picker-based flow (e.g. `i2i --ref <uuid>`) exercising the new final-scroll path — behavior-preserving on the happy path, so not release-gating

Refs #283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Picker grid scroll no longer misses a tile rendered by the final scroll (#283 off-by-one).** `_select_existing_asset` checked the tile count only *before* each scroll, so an asset the last scroll brought into the virtualised grid was never re-checked and the picker gave up with the tile on screen. A post-loop re-check closes it (`_find_picker_entity_tile` shares the loop shape but returns the locator unconditionally, so it was unaffected — now documented).
+- **Agentic `await_images` no longer trusts a single exact-count scrape (#283 hardening of the #281 race).** The poll loop breaks only when the new-UUID set is identical across two consecutive scrapes at the expected count (~one extra 0.5s poll); a set that transiently hits the exact count and then grows surfaces as the #281 `MediaAttributionError` instead of being returned as "the" generated media.
+- **Debug screenshots that fail to capture are no longer reported as if they existed.** `_capture_debug_screenshot` returned the target path even when `page.screenshot` raised (observed live 2026-07-11: an error message pointed at a file that was never written). It now returns `None` on capture failure and every error message appends its `Screenshot:` clause conditionally (new `screenshot_clause` helper) (#283).
+
+### Changed
+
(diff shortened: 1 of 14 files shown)


ffroliva:
## Council review verdict (8 dimensions @ f579550): YELLOW → findings applied in 1764437

| Dim | Verdict | Headline |
|---|---|---|
| D0 CI-mechanical | 🟢 | All gates green locally |
| D1 Correctness | 🟢 | Stable-break edges traced (oscillation, first-poll ambiguity, deadline); caught 2 raw `{shot_path}` sites the sweep missed — fixed |
| D2+D8 Quality/CLI | 🟢 | Help surfaces verified live; consolidation clean; docstring section added |
| D3 Security | 🟢 | Regex equivalence proven exhaustively (all 1,114,112 codepoints, zero divergence) |
| D4 Tests | 🟡 | ui_automation.py copy of the capture-failure fix was untested — mirrored; oscillation + deadline fall-through now pinned |
| D5 Memory/docs | 🟡 | KNOWN_ISSUES #281 residual paragraph rewritten (stable-break shipped, narrowed-not-closed); orphaned #283 pointer re-pointed; teardown leak got its durable record as #293 |
| D6+D7 UI/Data | 🟢 | +0.5s vs 180s budget, no tighter caller; recorder route= additive-only, no branching confusion |

**Final state:** 714 passed / 1 skipped across all affected suites; ruff + format + pyright + doc links + hygiene green.

**Deliberately not shipped (recorded):** E (bulk lookup — gated on batch >50 by the issue itself), I (upscale guard — low-risk by design), teardown leak → #293.

Issue #404, opened 2026-07-29
image t2i fails when requested count != displayed count: count setter clicks report success but never change the value (-n defaults to 1, so this breaks the default invocation)


ffroliva:
## Summary

On 0.46.0, `gflow image t2i` fails whenever the requested image count differs from what Flow already displays. Flow's composer now defaults to **2** images; gflow's count setter clicks the "1" tab three times, each click reports `success: true`, the displayed count never changes, and the run dies with an unhandled `RuntimeError` **before any generation is submitted**.

**`-n` defaults to 1, so this breaks the default invocation.** Any script calling `gflow image t2i` without an explicit `-n` now fails.

## Evidence

Two runs, same profile, minutes apart. Only `-n` differs.

**Fails — `-n` omitted (defaults to 1):**
```json
{"desired_count": 1, "panel_currently_visible": true, "initial_displayed_count": 2,
 "event": "ui_automation.count_setter_entered"}
{"target": "count=1", "selector": "nth(0) of _count_tabs_locator", "panel_visible": true,
 "current_displayed_count": 2, "event": "ui_automation.count_click_attempted"}
{"target": "count=1", "success": true, "current_displayed_count_after": 2, "error": null,
 "event": "ui_automation.count_click_result"}
   ... identical attempt x3 ...
{"desired_count": 1, "final_displayed_count": 2, "success": false, "attempts": 3,
 "event": "ui_automation.count_setter_completed"}
{"exception_class": "RuntimeError", "event": "error_unhandled"}
```

**Succeeds — `-n 2` (matches what Flow displays):**
```json
{"desired_count": 2, "final_displayed_count": 2, "success": true, "attempts": 0,
 "event": "ui_automation.count_setter_completed"}
```
→ exit 0, both images generated and downloaded, 768x1376, correct content.

## What this rules out

Collected from the incident bundle rather than assumed:

- **Not auth** — 69x `200`, no `401` anywhere in the trace.
- **Not a blocking overlay** — `ui.json` reports `"overlays": []` and `"tag_counts": {"dialog": 0}`. (Notably *not* the new watermark changelog modal from #403.)
- **Not `--project-name`** — the run reached `/fx/tools/flow/project/{id}`, so the 0.44 feature worked.
- **Not server-side** — no 4xx/5xx at all. The 15x `net::ERR_ABORTED` are POSTs to telemetry hosts, the usual cancelled-beacon pattern.
[...]

Issue #288, opened 2026-07-11
video i2v: --duration is silently ignored when the duration tab probe fails (requested 4s, got 8s)


ffroliva:
## Observed (0.31.0, live run 2026-07-11, chalkboard i2v pilot)

`gflow video i2v --duration 4` on the Frames submode (portrait, veo-lite, denon82 profile) produced an **8-second clip**. The duration tab selector never matched and the failure was demoted to warnings:

```json
{"probe": "duration_tab", "event": "ui_automation_video.selector_probe_failed", "level": "warning"}
{"seconds": 4, "event": "ui_automation_video.duration_not_set", "level": "warning"}
```

Generation then proceeded with Flow's default (8s). The JSON result reports success with no indication the requested duration was dropped — the only evidence is ffprobe on the output file.

## Why this matters

Duration is a *contract* parameter: downstream timeline math (scene slots in a Remotion cut) sizes freeze-holds and cut points from the requested value. A silent 4→8 substitution corrupts the edit plan without any failure signal. Same design principle as the #281 wrong-media fixes: when the UI automation cannot honor an explicit user parameter, fail fast (or at minimum surface it in the result payload), don't warn-and-continue.

## Suggested fix

1. Investigate the `duration_tab` selector against Flow's current Frames-submode DOM (may be another agentic/classic cohort split or a UI update — compare #183's aspect-switch probe rot).
2. Policy: explicit `--duration` + probe failure → hard error (new typed error, non-zero exit) unless a `--best-effort` style flag is passed. If warn-and-continue is kept for any path, the final JSON result must carry `requested_duration` vs `effective_duration` so callers can detect the substitution programmatically.

## Repro context

- 3/3 i2v generations in the run showed the same probe failure; both completed clips measured 8.000s via ffprobe (one requested 4s).
- Full structured logs available (correlation ids 17c1d945, 477bc557, 8ad4b3a1).

ffroliva:
Fix in flight on draft PR #289: an explicit `--duration` whose `duration_tab` probe misses now raises `UiSelectorDriftError` (exit 23) with a `debug_no_duration_tab.png` viewport screenshot, instead of silently generating at Flow's default. Omitting `--duration` is unchanged.

[...]

Issue #125, opened 2026-05-30
video i2v silently routes to T2V endpoint — start/end frames dropped, every i2v call burns credits as T2V


ffroliva:
## Summary

`gflow video i2v` silently routes to the **T2V** endpoint (`batchAsyncGenerateVideoText`) with `startImage: null`, `endImage: null`, and `referenceCount: 0` — even when the `_attach_frame` ↔ `_upload_via_open_dialog` dance reports clean success for both Start and End slots. Every i2v call burns 1 credit for a pure text-to-video generation; the start/end frames never reach Veo. The result video has zero visual continuity with the ref images and reads as a fresh T2V output.

This was the actual root cause of the wf-004 promo "drift" originally diagnosed (wrongly) as Veo style-shift — `_STICKMAN_MOTION` ran via T2V the whole time, never as I2V.

## Reproduction (gflow v0.10.0, profile `promo-denon82`, locale pt-BR)

```bash
gflow video i2v 01-t2i.jpg "<long prompt>" \
  --end-image 02-i2i.jpg \
  --aspect 9:16 --model omni-flash --duration 8 \
  --profile promo-denon82 --out-dir <tmp>
```

(`01-t2i.jpg` and `02-i2i.jpg` are 9:16 stickman frames; any pair will reproduce.)

## Smoking-gun structured-log sequence

correlation_id `b85dde08-5a86-4384-8af5-d4f0ad800298`:

| ts (UTC) | event | notes |
|---|---|---|
| 12:39:17.014 | `ui_automation_video.image_uploaded` | target=Start, status=200 |
| 12:39:18.976 | `ui_automation_video.frame_attached` | slot=Start |
| 12:39:27.573 | `ui_automation_video.image_uploaded` | target=End, status=200 |
| 12:39:29.434 | `ui_automation_video.frame_attached` | slot=End |
| 12:39:29.441 | `ui_automation.prompt_input_found` | |
| 12:39:30.140 | `ui_automation.prompt_submitted` | via `arrow_forward` icon |
| 12:39:33.373 | `ui_automation_video.generate_captured` | **url=`.../v1/video:batchAsyncGenerateVideoText`** |

The `generate_captured` body summary (from `_summarize_request_image_inputs`):

```json
{
  "parsed": true,
  "startImage": null,
  "endImage": null,
  "referenceCount": 0,
  "referenceIds": []
}
```

[...]

PR #424, opened 2026-08-03
fix(video): fail closed on count-tab miss; re-enable omni-flash start-frame i2v (refs #125, #404)

Mauryanx:
## Summary

Two billing-safety fixes for the video path, plus the **resolution of the omni-flash i2v exclusion** (refs #125, #404): start-frame i2v with Omni Flash is re-enabled, wire-proven twice (route-aborted capture + one live x1 clip).

### 1. `_set_output_count` now fails closed (was: silent x2 double-bill on selector drift)

Before v0.46.1 the video count setter probed the pre-rename `1x` label, missed the renamed `x1` tab (#404), logged `count_not_set`, and proceeded — so a `--count 1` run generated at Flow's sticky default of **x2 and billed double**. v0.46.1 fixed the labels, but the miss path stayed fail-open: any future rename re-opens the double-bill.

This PR makes a count-tab miss **fatal pre-submit** (`UiSelectorDriftError`, exit 23, debug screenshot), mirroring the duration probe (#288) and the required i2v model select (#125). Count is the credit *multiplier* — every `GenerateVideoRequest` carries a definite 1–4 value, so "proceed on Flow's default" silently doubles spend for count=1 and under-delivers for count=3/4. The refusal runs before any frame attach or submit, so it spends nothing.

This resolves a pre-existing asymmetry: a missing *duration* tab was fatal (#288) while a missing *count* tab was ignorable — backwards from a billing standpoint, since duration changes cost per clip but count multiplies clips.

Label matching is now affix-agnostic per digit (`xN` current / `Nx` legacy) for **every** count — the same digit-keyed approach the image path's `_set_count` uses — so the next #404-class rename degrades to a fallback selector instead of a refusal.

### 2. omni-flash i2v (#125): exclusion resolved — start-frame re-enabled, end-frame still gated

The #125 exclusion rested on a 2026-05-30 route-abort capture: Flow's frontend silently rerouted omni+frames submits to `batchAsyncGenerateVideoText` with `image_inputs: null`. Re-verified 2026-08-03 in three independent ways:

1. **Credit-free UI recon** (`scripts/dev/spike_omni_flash_i2v_ui_recon.py`, new in this PR): today's editor offers Omni Flash with Frames active, keeps a bound Start frame across the model switch, and exposes the 10s duration. Google's official support matrix (support.google.com/flow/answer/16352836) lists **"Frames to Video: First" as supported for Gemini Omni Flash** (4s–10s; first+last "coming soon").
2. **Route-aborted submit capture** (zero credits, same methodology that established #125): `capture_i2v_intercept_submit.py --model omni-flash --start-only` → the frontend now fires `batchAsyncGenerateVideoStartImage` with a **non-null `startImage`**.
3. **One live x1 generation** through the real CLI (`gflow video i2v … --model omni-flash --duration 10 --count 1`): captured request `startImage: 7fff88eb` on the StartImage route, `MEDIA_GENERATION_STATUS_SUCCESSFUL`, and the downloaded 10.01s 720×1280 mp4's **first frame is pixel-identical to the uploaded start frame** (solid-red probe: frame-0 center mean RGB 200,55,36 vs source 200,60,40), fading into the prompted scene by frame 30 — genuine interpolation, not T2V.

Changes:

- `VideoModel.supports_i2v_interpolation()` now returns True for all current models (evidence-dated docstring); new `supports_i2v_end_frame()` gates the **end frame**: `--end-frame` with omni-flash is rejected pre-spend (exit 17) — Flow's matrix lists first+last as "coming soon" for Omni Flash and there is no wire proof of that route.
- `gflow video i2v` accepts `--model omni-flash` and `--duration 10`; default stays `veo-lite`.
- **Chains keep rejecting omni-flash** (CLI + `chain.run_chain`): single-clip start-frame i2v is proven, N seeded links back-to-back is not — deliberate conservatism, can be lifted in a follow-up after a chain-scale verification.
- Transport guard (`_resolve_i2v_model`), MCP docstring, USAGE.md, and the stale "drops frames to T2V" rationales across model-select messages all updated to the current evidence.

### Drive-by observation (not changed here)

Today's model picker lists only 4 models — `Veo 3.1 - Lite [Lower Priority]` is gone from the menu (evidence in the spike run output). `VIDEO_MODEL_OPTION_SELECTORS` still registers it; selecting `veo-lite-lp` today would log `model_option_not_found` and fall through to Flow's default on t2v/r2v, and refuse on i2v. Left for a follow-up issue.

## Test plan

- [x] Count-miss now asserts `UiSelectorDriftError` + billing message + screenshot capture; both label cohorts pinned for count=1; legacy-affix fallback pinned for count=3
- [x] omni-flash i2v: start-only acceptance pinned end-to-end at CLI (`_run_i2v` → request carries OMNI_FLASH) and transport level; omni + `--end-frame` rejection pinned at CLI (exit 17) and transport level; chain rejection pinned at `run_chain` and CLI
- [x] `ruff check` / `ruff format --check` green; `pyright src` strict: 0 errors
- [x] Repo hygiene / doc-links / website-PII gates green; website mirror regenerated
- [x] Full `pytest -m "not live and not e2e and not smoke"` sweep green
- [x] Live verification 2026-08-03 on profile `mauryan-uppalapati`: 2 credit-free recon sessions + 1 route-aborted capture (zero credits) + exactly **one** billed x1 omni-flash 10s clip (media `ba7f1ed0-49a0-439d-958e-5933ec839918`), first-frame interpolation verified programmatically

CHANGELOG.md deliberately untouched — entry left to the release flow.


ffroliva (changes requested):
## Council review — YELLOW (changes requested, all addressable)

Thanks @Mauryanx — this is a high-quality PR. A 10-dimension review (correctness, quality, security, tests, docs, CLI UX, scripts, YAGNI, memory, live-verify) ran against `93629db3`, plus the full mechanical gate locally (ruff check/format, doc-links, repo hygiene: green; offline pytest sweep: **2900 passed, 0 failed**; duplication proxy: no new duplicate code vs `develop`).

Confirmed independently: the fail-closed refusal genuinely fires pre-spend (traced `configure_video_settings` → count step before any frame attach/submit), the digit-keyed selectors are locale-safe, every test-plan claim maps to a real behavioral test, and the security audit of the new spike script came back clean. Nice work on the evidence-dated docstrings and the route-abort methodology.

### Must-fix before merge

1. **Stale "omni drops frames → T2V" prose survives in 5 places and now contradicts this PR's own change** — including the user-facing exit-17 output, which contradicts itself on one screen (the detail says "drop `--end-frame`", the hint still says a *start* frame needs veo-lite):
   - `src/gflow_cli/errors.py:404` — `ModelModeIncompatibilityError` docstring + remediation hint ("start or start+end frames")
   - `docs/USAGE.md:700-703` — chain section ("silently drops the seed frame and routes to text-to-video"); chain staying gated is fine, but the stated *reason* is now falsified — reword to "gated pending chain-scale verification"
   - `docs/USAGE.md:1328` — exit-17 table row (example + "use veo-lite for i2v" remedy are start-inclusive stale)
   - `docs/CONFIGURATION.md:414` — chain `--model` row, same rationale
   - `KNOWN_ISSUES.md:708` — "the only model known to silently drop frames"
   - (plus the `website/docs/` mirrors of the above)

2. **Delete the dead `_set_output_count_one` shim** (`ui_automation_video.py:1469`) — zero production callers on both base and head (`grep -rn "_set_output_count_one" src/` → only the def), yet 3 rewritten tests target it. Point those asserts at `_set_output_count(page, 1)` directly (the screenshot test already does).

3. **`supports_i2v_interpolation()` is now constant-True** (`video.py:71`) — `cli_video.py:392`'s `elif not …` branch is unreachable, and the `X is OMNI_FLASH or not X.supports_i2v_interpolation()` disjuncts in `cli_video.py:498` / `chain.py:323` / `ui_automation_video.py:3262` are vacuous. Either delete the method + shrink the call sites (the #125 history is already preserved in `supports_i2v_end_frame()`'s docstring), or keep it as an explicit seam and simplify the vacuous disjuncts — half-dead conditions reading as live logic is the worst of both.

4. **Spike script: use `scripts/dev/_spike_common.py`** (`spike_omni_flash_i2v_ui_recon.py:50,169-171`) — the hand-rolled `sys.path.append` + `profile_subdir(default_home(), …)` bootstrap is what `_spike_common` (used by all 23 recent spikes, e.g. `spike_issue404_count_tabs_recon.py`) exists to kill. ~4-line swap to `resolve_profile_dir` + `build_client`; also fixes the run-from-repo-root-only path assumption.

5. **CHANGELOG entry** — repo convention is that feature/fix PRs carry their own `[Unreleased]` entry (recent precedents: #400, #404, #407, #408); "left to the release flow" doesn't match observed practice. Please add one.

Maintainer-side (not on you): we'll re-run the 0-credit route-abort capture (`capture_i2v_intercept_submit.py --model omni-flash --start-only`) to independently confirm the wire claim, and approve the workflow run so SonarCloud reports on the real diff.

### Nice-to-have (follow-up sized, no need to block on these)

- Port the image path's read-back + `effect_observed` retry to the video `_set_output_count` — the fail-closed change covers a *miss*, but an overlay-intercepted no-op click can still log success (`ui_automation_video.py:1462-1466` vs `ui_automation.py:1967-2080`).
- Parametrize the count tests over 1–4 (only n=1 and n=3 pinned); a one-line MCP parity pin for omni i2v.
- i2v `--count` help lacks t2v's ">1 multiplies credit cost" warning.
- Stale "non-interpolation model" docstrings in `tests/test_chain.py:31,278` and `tests/e2e/test_transports_e2e.py:184-208` comments.
- Error detail prints `omni_flash` (enum value) where the CLI surface says `omni-flash`.

Once the must-fixes land, this is merge-ready pending the live-verification gate above.

Mauryanx:
Thanks for the thorough review — all five must-fixes are addressed in `959e4d0`, plus a new fail-fast fix in `d893a55` (details at the end). All commits are now DCO signed-off.

### Must-fixes

1. **Stale "omni drops frames → T2V" prose — purged at all five sites** (+ regenerated `website/docs/` mirrors). You were right that the exit-17 output contradicted itself on one screen; `ModelModeIncompatibilityError`'s docstring and remediation hint now describe the *current* gates (omni = start-frame only → drop `--end-frame`; Veo 3.1 for first+last and for chains). `docs/USAGE.md` chain block and exit-17 row, `docs/CONFIGURATION.md` chain `--model` row, and the `KNOWN_ISSUES.md` #125 section now state the real reason chains stay gated — pending chain-scale verification, not a falsified frame-drop claim. The KNOWN_ISSUES section keeps the historical observation, dated, since the per-link wire-route abort still exists for it.

2. **`_set_output_count_one` deleted.** Its three tests now call `_set_output_count(page, 1)` directly, matching the screenshot test.

3. **`supports_i2v_interpolation()` deleted** rather than kept as a seam — a constant-True predicate guarding four call sites was the "half-dead condition reading as live logic" you flagged, and the #125 history now lives in `supports_i2v_end_frame()`'s docstring where it stays load-bearing. The vacuous disjuncts in `cli_video.py` / `chain.py` / `ui_automation_video.py` collapsed to the explicit `is OMNI_FLASH` chain gate and the end-frame capability check. One knock-on worth flagging: the i2v-default test now pins `supports_i2v_end_frame()` — the *full*-surface capability — so it still catches a future default pointed at a narrower model.

4. **Spike script now uses `_spike_common`** (`resolve_profile_dir` + `build_client`), matching `spike_issue404_count_tabs_recon.py`; the hand-rolled `sys.path` + `profile_subdir` bootstrap and the run-from-repo-root assumption are gone.

5. **CHANGELOG `[Unreleased]` entries added** — my mistake on the convention, the recent precedents you listed are unambiguous.

Also picked up three of the nice-to-haves since they were one-liners: the i2v `--count` help now carries t2v's credit-cost warning, and the stale "non-interpolation model" rationale is refreshed in `tests/test_chain.py` and the e2e transport comments. The remaining nice-to-haves (read-back retry port for the video count setter, 1–4 parametrization, MCP parity pin, `omni_flash` vs `omni-flash` in error text) I've left for follow-ups — happy to fold any in if you'd rather they land here.

### New: silent-hang fix (`d893a55`)

While this was in review, `gflow video i2v` was reproduced hanging **silently** twice: last log line `ui_automation_video.frame_attached`, then nothing — browser alive, no error, ~3 minutes until killed.

Static analysis of that window rules out both obvious suspects. The overlay dismisser logs `overlay_detected` whenever it acts and that line is **absent** from both hung logs, so it returned clear in ≤9s. The prompt-box lookup is 4 selectors × 10s and then *raises* with a screenshot, so it can't produce minutes of silence either. Every probe in the window is individually bounded — which means the stall is a Playwright call that stopped honouring its own deadline (the page never answers the CDP command, so the per-probe timer never fires). A media dialog / file chooser left open by the preceding upload step is the leading candidate, but **that root cause is not proven and stays open** — there's a correlation worth noting (all 9:16 runs succeeded, both 16:9 runs hung, same model and duration) that I couldn't chase further without more repros.

What's fixed here is the silence itself: the submission stage now runs under a named wall-clock watchdog. On expiry the run aborts **pre-submit** with `TransportTimeoutError` naming the stage, emits `stage_stalled`, and captures a screenshot under its own short deadline (an unbounded capture would re-hang the path meant to end the hang). Nothing is submitted, so no credit is spent. Tests cover pass-through, the named-stage timeout, the wedged-screenshot case, and the event payload.

### Checklist

Targets `develop` · real identity (`Mauryanx`, GitHub noreply) · DCO sign-off on all 5 commits · no secrets/cookies/tokens/signed-URLs/private captures in the diff (scanned; no binaries, HARs, logs or screenshots included) · focused tests and docs for every behaviour change.

Gates: `ruff check` / `format` clean, `pyright src` strict 0 errors, repo-hygiene / doc-links / website-PII green, full offline sweep green.

On your maintainer-side re-run of the route-abort capture — worth knowing the exact invocation I used, since the script defaults to inheriting Flow's last-used model: `capture_i2v_intercept_submit.py --profile <p> --model omni-flash --start-only --probe-image `. It aborts the XHR in-browser, so it costs nothing.

Mauryanx:
**Root cause found for the silent hang — and it was a dependency bug, not a Flow/UI one.** `71d71cb` adds the real fix; my earlier note that a lingering media dialog was the leading candidate was wrong, and I'm correcting it rather than leaving it to stand.

### What it actually was

`uv tool install <path>` **ignores `uv.lock`** and resolves from the `pyproject.toml` ranges. `playwright>=1.45.0` had no upper bound, so that install resolved **1.62.0** against a project locked to **1.59.0**. Every `video i2v` run on that build hung silently right after the frame upload; reinstalling with `--with playwright==1.59.0` fixed it instantly and the next clip submitted first try. That also explains the 9:16-vs-16:9 correlation we were chasing — it was which *build* each run used, not the aspect and not omni-flash.

### (a) The constraint: `playwright>=1.59.0,<1.60.0`

Reasoning, since the bound is a judgement call:

- **Playwright is not an ordinary dependency here.** It ships the browser driver, and 100% of gflow's product surface is automation through it — so an untested Playwright minor is an untested product. This repo already applies exactly that reasoning to the other engine (`patchright==1.61.2`, exact-pinned in the extra).
- **Upper bound `<1.60.0`** excludes untested minors. I chose it over an exact `==1.59.0` deliberately: `playwright` is the default engine every user installs, so leaving patch headroom means a CVE fix doesn't require a gflow release. The extra can be stricter because it's opt-in.
- **The floor moves 1.45.0 → 1.59.0** because 1.45.0 was aspirational — nothing has ever been tested there; the lockfile and CI have always exercised 1.59.0. An honest range beats a wide one.
- **Verified, not assumed:** a lockfile-ignoring `uv pip install --upgrade .` (what `uv tool install <path>` does) now resolves `playwright==1.59.0`, where it previously reached 1.62.0.

One thing worth flagging for the maintainers: **`resolve-drift` did not catch this**, and couldn't have. It installs from the ranges ignoring the lockfile — exactly the right mechanism — but then only smoke-imports, and a driver *behaviour* regression is invisible to an import. The upper bound is what closes that specific gap; the watchdog below is the general net for the next one. If you want defence in depth there, the job could additionally assert the resolved `playwright` version falls inside a known-tested set, which is cheap and catches the class rather than the instance. Happy to add it here if you'd like.

`tests/test_playwright_pin.py` locks four things: that an upper bound exists at all, that the tested version satisfies the range, that the known-bad 1.62.0 is excluded, and that the range gflow *prints* to users matches what the package actually allows.

### (b) The stall now names the cause

The watchdog from `d893a55` stands — a silent multi-minute hang is indistinguishable from slowness in an overnight batch, which is precisely what made this expensive to find. Now that the cause is known, the error is specific: it names the stage, prints **your installed Playwright version against the supported range**, and gives the pinned reinstall command. I kept the version check in the error rather than as a startup assert on purpose — a future driver regression may present the same way on an *in-range* version, so the watchdog has to stay a general net rather than become a version guard.

### (c) Install note

README carries the note (local/tool installs must carry the locked Playwright, with the check command), plus a `KNOWN_ISSUES.md` entry — status *Mitigated*, both layers described — and the exit-9 row in `docs/USAGE.md` now points at it. PyPI installs were never exposed once the published range is bounded.

Gates: ruff clean, `pyright src` 0 errors, hygiene/doc-links/website-PII green, full offline sweep **2904 passed**. All commits DCO signed-off.

ffroliva (approved):
Approved — all review must-fixes, lockfile pin, wire intercept, and live E2E generation verified.

ffroliva:
Merged! Thank you @Mauryanx for the exceptional contribution — especially the deep investigation into the `uv tool install` Playwright version drift bug and the clean upper-bounding fix, alongside the 0-credit UI recon spike and comprehensive must-fix cleanup. Outstanding work!

ffroliva:
Update: We are shipping your contribution right now in release v0.49.0! 🚀

ffroliva:
Coming back to this a month later, because it deserves more than a merge.

The `_set_output_count` fail-closed change fixed a **silent x2 double-bill on selector drift** — the tool quietly charging people twice when Flow moved a label. That's the worst class of bug this project can have, and you found and closed it.

It's also, as it turns out, the thing gflow-cli now leads with. "A broken run costs a message, not credits" went on the front page of the site today, and the whole current pitch is that failures fail loudly instead of spending money. Your PR is a good part of why that claim is true.

So: thank you, properly.

If you're up for it, I'd like to know what you're building with it — you went into the billing path far enough to spot a double-bill, which means you were watching credits closely for a real reason. Almost the entire roadmap so far has come out of my own head, and the people actually running it are a much better source.

Mauryanx:
Thanks for the kind words. I really appreciate all the work you've been doing. Need more people like you .

Would love to connect and talk about what i'm been building and hear about what you are doing as well.

Let me know when works

PR #432, opened 2026-08-04
release: v0.50.0

ffroliva:
Release v0.50.0 containing MCP Tasks extension (SEP-2663) and output path hardening.


ffroliva:
Superseded by #433. The `Only release/hotfix branches may target main` guard requires a `chore/release-*` head, so this branch name could never pass. #433 carries the same commits plus the plugin.json bump and the cryptography CVE patch.

PR #448, opened 2026-08-04
test(e2e): two e2e tests could never pass — daemon transport and obsolete strategies

ffroliva:
## Summary

`test_daemon_e2e_lifecycle` **could never pass**. Found while live-verifying the dependency upgrades in #445 — and confirmed pre-existing by A/B against the pre-upgrade lockfile, so this is not upgrade fallout.

## The bug

The test hand-rolled the **legacy SSE transport**:

```
GET  /mcp/sse                      -> parse `event: endpoint`
POST /mcp/message?session_id=...   -> read response off the SSE stream
```

The daemon serves **Streamable HTTP** at `HTTP_PATH` (`/mcp`) — `server.py` even tells users to *"migrate to --transport http (Streamable HTTP at /mcp)"*.

So the readiness probe waited for a `200` that never arrives, and every run died in the `"Daemon failed to start"` branch **while the daemon was up and healthy** — the captured stderr in the failure shows Uvicorn bound and the session manager started. The misleading error message is why this went unnoticed.

**Net effect: the MCP daemon lifecycle has had no live coverage since the transport migration.**

## The fix

Speak the real protocol via the SDK's own client (`streamable_http_client` + `ClientSession`) rather than hand-rolled framing — hand-rolled transport plumbing is precisely what rotted here. The test now asserts:

- `server_info.name == "gflow-cli"` (initialize handshake completes)
- `gflow_list_projects` is in the registry **served over the wire**, not an in-process import
- a real `call_tool` returns a non-error result

Two hardening details:

- **Readiness is a TCP connect, not an HTTP 200.** Streamable HTTP answers a bare GET with 4xx/406, so "port accepts a connection" is the honest signal.
- **`HTTP_PATH` is imported, not hardcoded**, so a future mount-path change fails the import instead of silently making the endpoint unreachable again.

## Verification

```
GFLOW_CLI_E2E_PROFILE=denon82 pytest -m e2e tests/e2e/test_daemon_e2e.py
  1 passed
```

Against a real spawned daemon. Also `ruff check` clean, `ruff format --check` clean, `pyright` **0 errors**.

Zero credits — no generation is triggered.

## Related, not fixed here

`test_transports_e2e.py::test_e2e_health_check_returns_true_when_active[bearer|sapisidhash]` also fail, identically on both lockfiles. They ask `FlowApiClient` to run a transport it deliberately refuses (`"acquires its own profile lease... would self-lock"`). That is a test contradicting a design guard and needs a separate decision, so it is not bundled here.

### DIFF
--- a/tests/e2e/test_daemon_e2e.py
+++ b/tests/e2e/test_daemon_e2e.py
@@ -5,10 +5,12 @@ import socket
 import subprocess
 import sys
 from pathlib import Path
-from urllib.parse import parse_qs, urlparse
 
-import httpx
 import pytest
+from mcp import ClientSession
+from mcp.client.streamable_http import streamable_http_client
+
+from gflow_cli.mcp.server import HTTP_PATH
 
 
 def _free_port(host: str) -> str:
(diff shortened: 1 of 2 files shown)


ffroliva:
## Second fix pushed — `fcf027b`

While verifying the daemon fix I cleared the other two failures from the same live-verify pass. Same class of defect: **tests asserting behaviour the system no longer provides.**

`test_e2e_health_check_returns_true_when_active` parametrized over the full `STRATEGIES` list and asserted `health_check() is True` — a **success path** that `bearer` and `sapisidhash` cannot satisfy. KNOWN_ISSUES.md is explicit: *"These are obsolete — only `evaluate_fetch` is viable."*

The sibling `test_e2e_health_check_false_after_close` **already documents that exact reason** for not parametrizing. This one kept the full list, so it failed 2 of 3 strategies on every run.

Added `LIVE_STRATEGIES` for success-path assertions. Error-path tests deliberately keep the full `STRATEGIES` list — asserting a broken transport degrades cleanly is the point of those, and `test_e2e_unrecoverable_auth_expiry` passes on all three.

### Knock-on fix worth noting

This also cleared `test_rest_upload_image_authenticates_after_sapisidhash`, the third failure — the one that passed in isolation but failed in a full-suite run. **The obsolete transports were the ones acquiring the contending profile lease.** Removing them removed the contention, so that was a symptom, not an independent bug.

### Result

```
GFLOW_CLI_E2E_PROFILE=denon82 pytest -m e2e_auth
  before:  15 passed, 3 failed
  after:   16 passed, 0 failed
```

18 → 16 tests because two dead parametrizations were removed. The full zero-credit gate is now green. `ruff`, `ruff format --check`, `pyright` (0 errors) all clean.

PR #450, opened 2026-08-05
chore(deps): raise playwright 1.59.x -> 1.61.x (needs a live generation before merge)

ffroliva:
> [!IMPORTANT]
> **Draft — do not merge until a live generation verifies it.** Merging on offline evidence alone would repeat exactly the mistake this pin was created to prevent.

## Summary

`playwright` **`>=1.59.0,<1.60.0` → `>=1.61.0,<1.62.0`**, lockfile to 1.61.0.

## Why 1.61.0 and not latest

**1.62.0 stays excluded deliberately.** It is the version observed on 2026-08-03 making every `video i2v` run hang *silently* right after the frame upload — browser alive, no error, no timeout. That was never root-caused; it was worked around by pinning back to 1.59.0.

Stepping to 1.61.0 takes two minors of driver improvements while leaving the known-bad version out. Raising to 1.62.0 should happen only after someone reproduces that hang against it deliberately and finds the cause.

## The pin guard did its job

`tests/test_playwright_pin.py` **failed on the first run**, which is the system working:

```
AssertionError: the tested version 1.59.0 does not satisfy '<1.62.0,>=1.61.0'
AssertionError: the range shown in the stall error drifted from pyproject.toml
```

Raising the bound also requires `PINNED_PLAYWRIGHT` and `SUPPORTED_PLAYWRIGHT_RANGE` in `ui_automation_video.py` to move — those strings are printed in the **stall error that tells a user how to recover**, so a drifted pin would hand out a wrong remediation command. Both updated; the guard now passes.

## Verified on 1.61.0

| gate | result |
|---|---|
| `pytest` (offline) | **2923 passed**, 3 skipped |
| `-m e2e_auth` (0 credits) | **16 passed** — browser launch, cookie state, session verification, transport health checks |
| daemon e2e | **1 passed** — MCP over Streamable HTTP against a live daemon |
| MCP Tasks live-verify | **5/5 layers** |
| `ruff` / `ruff format` | clean |
| `pyright src` | **0 errors** |
| `pip-audit --all-extras` | no known vulnerabilities |

## NOT verified — and it is the whole point of this bound

**The generation path.** The 2026-08-03 failure was a *silent hang* after frame upload. No offline test and no zero-credit e2e can observe that — the pyproject comment states it outright:

> offline tests cannot see a driver-behaviour regression (`resolve-drift` in CI only smoke-imports)

Confirming 1.61.0 requires a real `video i2v` run, which **spends Veo credits**.

## To finish this

One live `i2v` generation on this branch. If it completes normally, this is safe to merge and I'll add the evidence ledger to `docs/`. If it hangs after frame upload, the bound goes back to `<1.60.0` and we know 1.60/1.61 carry the same regression as 1.62 — which is itself valuable information nobody has today.

A cheaper partial signal: an `i2i` run with a local reference image exercises the same upload mechanism at **Imagen** cost rather than Veo. It would not prove the video path, but a hang there would rule 1.61.0 out immediately.


ffroliva:
## Live verification complete — the regression does not reproduce on 1.61.0

Ran both generations. Evidence ledger added: `docs/LIVE_VERIFICATION_playwright_1.61.md`.

### Live i2v — the exact failing path

Driven via the CLI (see the caveat below for why). Structlog trace, in order:

| event | value |
|---|---|
| `video_submode_entered` | `sub=frames` |
| `aspect_set` / `output_count_set` | `portrait` / `count=1` |
| **`image_uploaded`** | **`target=Start status=200`** |
| **`frame_attached`** | **`slot=Start`** |
| `prompt_submitted` | via `arrow_forward` |
| **`generate_captured`** | **`status=200`**, `startImage=893a001d` parsed |
| `poll_terminal` | terminal state reached |

**1.62.0 hung silently at that upload step** — browser alive, no error, no timeout. On 1.61.0 the upload returns 200 and every downstream step executes.

| | 1.62.0 (2026-08-03) | 1.61.0 (this run) |
|---|---|---|
| after frame upload | **hung silently** | `status=200`, continued |
| error surfaced | none | structured, terminal |
| timeout | never fired | polled normally |

### Live i2i — same upload mechanism

```
test_e2e_i2i_local_ref_attach   1 passed in 96.71s
```

Uploaded, attached, generated, downloaded.

### Not claimed

**A finished mp4.** Flow's own generation returned `PUBLIC_ERROR_VIDEO_GENERATION_TIMED_OUT` — server-side capacity, which this change cannot influence. The thing under test is the driver chain, and it passed end to end.

---

## Unrelated defect found — A/B proven pre-existing

`test_e2e_i2v_start_end_frame_attach` fails before generation:

```
UiSelectorDriftError: probe=duration_tab: the 4s duration tab was not found
```

**Identical on playwright 1.59.0 (24.33s) and 1.61.0 (28.16s)**, and reproduced with `GFLOW_CLI_E2E_VIDEO_DURATION=8` too — so Flow dropped the whole duration-tab UI, not one value.

**`--duration` is currently broken for i2v against live Flow, on every playwright version.** The guard is behaving correctly (refusing rather than silently accepting Flow's default, #288), but the selector needs re-deriving. This is why the i2v proof above was driven via the CLI without `--duration`, and it needs its own issue.

## Ready to merge

1.62.0 stays excluded until its hang is root-caused.

PR #578, opened 2026-08-25
refactor(api): consolidate request headers into one _request_headers() (401 premise REFUTED by spike)

ffroliva:
## What this is

A refactor, plus the spike that stopped it being sold as a bug fix.

### The refactor

`_post_json`, `_patch_json` and `_get_json` each assembled headers inline behind an `if is_aisandbox` gate — three copies of the same decision, free to diverge. A header a lane needs could go missing on one verb and nobody would notice. Construction now lives in one `_request_headers()` used by all three.

The labs.google lane also now sends `origin`/`referer`, which every other lane in this tree already sends (aisandbox, agentic driver, sapisidhash transport, httpx auth probe) and which two independent reverse-engineered Flow clients send on this exact route.

Keyed on the **URL**, not on `not is_aisandbox` — `routes.py` defines two hosts today, so an `else` would be correct only by coincidence; a third host added later would silently inherit a `labs.google` Origin. A sibling project documents that exact failure: *"an Origin naming the other host fails Google's origin-bound auth checks."*

## What this is NOT — read before merging

This began as a cherry-pick of `03a1e52`, on the claim that an origin-less tRPC mutation gets 401 as CSRF-shaped — the suspected cause of the recurring canary RED.

**That claim is false.** Live A/B against `project.createProject` (profile `denon82`, no credits — a tRPC mutation is not a generation):

```
session probe            -> HTTP 200 alive=True
A_control_no_origin      -> HTTP 200, 200     (content-type only, what develop sends)
B_with_origin_referer    -> HTTP 200, 200
```

Origin-less mutations succeed. The July 401 was account-specific, transient, or has since changed server-side.

So: **these headers are harmless and consistent, not necessary.** No issue is closed here. The canary 401 remains unexplained — and note it fails on the **aisandbox** lane, not this one; the two were never the same failure.

## The harness

`scripts/dev/spike_trpc_origin_referer_401.py` is committed because it is worth more than its result. Two controls were each the difference between a right answer and a confident wrong one:

1. **Session liveness probe before any arm.** A dead session 401s BOTH arms, which reads as "no difference" and would have refuted the hypothesis for the wrong reason.
2. **Verify the instrumentation, not just the result.** The first run captured wire headers via `page.on('request')` and reported `origin=None` for BOTH arms — including the one that explicitly passed one. Broken instrumentation presenting as evidence: `page.request` is an `APIRequestContext` and emits no page-level request events. Confirmed against a header-echo endpoint that Playwright adds no Origin of its own, so the control arm was genuinely bare.

Arms interleave A/B/A/B in one session so ordering, rate limiting and session decay cannot masquerade as the effect.

## Also not included

`LaneAuthRejectedError` from `03a1e52` depends on `5d57044`, a 1204-line error-contract feature not on `develop`. ⚠️ `git cherry-pick 03a1e52` applies **cleanly** onto develop and then fails at import, referencing a class develop does not have.

## Verification

- [x] `tests/api`: 1167 passed, 3 skipped
- [x] Full non-e2e suite: 3374 passed, 5 skipped, 0 failed
- [x] `ruff` clean, `pyright` 0 errors
- [x] Negative control: reverting the header line fails 3 tests; restoring passes them
- [x] New test pins that a third host receives no labs Origin
- [x] **Live-verified** — unusually, the behaviour change is the thing we proved *unnecessary* rather than necessary

## Merge or not — your call

The consolidation is worth having on its own. The headers ride along at proven-zero risk. If you would rather not change live auth-path behaviour without a demonstrated need, drop the `elif` branch and keep the pure refactor — one line.

Owed cleanup: 8 projects named `spike401-*` on denon82's account.

### DIFF
--- a/scripts/dev/spike_trpc_origin_referer_401.py
+++ b/scripts/dev/spike_trpc_origin_referer_401.py
@@ -1,0 +1,151 @@
+"""Spike: does the labs.google tRPC lane 401 without origin/referer? (#578)
+
+We do not guess, we verify. PR #578 claims a tRPC MUTATION rejected for lacking
+`origin`/`referer` answers 401. That claim is inference from a commit message
+plus two third-party clients — nothing in our tree had ever observed it.
+
+Experiment (A/B, interleaved, one session):
+
+    A = control  — headers exactly as `develop` sends today: content-type only
+    B = treatment — content-type + origin + referer (what PR #578 sends)
+
+Both run against the SAME live `project.createProject`, on the same page, the
+same cookie jar, within seconds of each other, alternating A/B/A/B so ordering,
+rate limiting or session decay cannot masquerade as the effect.
+
+Costs no Veo/Imagen credits — createProject is a tRPC mutation, not a
+generation. Successful calls DO create real projects, named `spike401-*` so
+they are obvious to delete afterwards.
+
+Run:
+    uv run python scripts/dev/spike_trpc_origin_referer_401.py --profile ffroliva
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import sys
+from pathlib import Path
+from typing import Any
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from _spike_common import build_client, default_out_path, resolve_profile_dir  # noqa: E402
+
+TRPC_CREATE = "https://labs.google/fx/api/trpc/project.createProject"
+LABS_ORIGIN = "https://labs.google"
+LABS_REFERER = "https://labs.google/fx/tools/flow"
+
+ARMS = {
+    # exactly what develop sends today
+    "A_control_no_origin": {"content-type": "application/json"},
+    # exactly what PR #578 sends
+    "B_with_origin_referer": {
+        "content-type": "application/json",
+        "origin": LABS_ORIGIN,
+        "referer": LABS_REFERER,
+    },
+}
+
+
+async def one_call(page: Any, arm: str, headers: dict[str, str], n: int) -> dict[str, Any]:
+    """Fire one createProject and record what ACTUALLY went on the wire.
+
+    Critical control: `page.request` is bound to the page's context, and
+    Playwright/Chromium may add `origin`/`referer` of its own accord. If it
+    does, the "control" arm is not origin-less at all and the whole A/B is
+    meaningless. So we capture the real outgoing headers rather than trusting
+    the dict we passed in.
+    """
+    seen: list[dict[str, str]] = []
+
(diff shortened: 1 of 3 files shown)


ffroliva:
## ⚠️ Premise refuted by live spike — do not merge this as a 401 fix

I built an A/B spike (`scripts/dev/spike_trpc_origin_referer_401.py`) and ran it against live `project.createProject` on profile `denon82`. **The header hypothesis does not hold.**

### Method

Interleaved A/B, one session, one cookie jar, seconds apart, alternating A/B/A/B so ordering, rate limiting and session decay cannot masquerade as the effect:

- **A (control)** — headers exactly as `develop` sends today: `content-type` only
- **B (treatment)** — `content-type` + `origin` + `referer` (what this PR sends)

Two controls on the experiment itself:

1. **Session liveness probe** before any arm — a dead session would 401 *both* arms and read as a false null. Probe returned HTTP 200, `flow_session_cookie_present=True`.
2. **Instrumentation check.** My first attempt captured wire headers via `page.on('request')` and recorded `origin=None` for *both* arms — including the arm that explicitly passed one. That was broken instrumentation, not evidence: `page.request` is an `APIRequestContext` and does not emit page-level request events. Re-checked against a header-echo endpoint using the identical call shape: control → `origin present? False`, treatment → `origin present? True`. **Playwright adds no Origin of its own, so the control arm was genuinely origin-less.**

### Result

```
session probe            -> HTTP 200 alive=True
A_control_no_origin      round 1 -> HTTP 200
B_with_origin_referer    round 1 -> HTTP 200
A_control_no_origin      round 2 -> HTTP 200
B_with_origin_referer    round 2 -> HTTP 200
```

**An origin-less tRPC mutation succeeds.** 4/4. The 401 described in `03a1e52` does not reproduce on this account today — whatever caused it in July was account-specific, transient, or has since changed server-side.

### What this means for this PR

- ❌ It is **not** a 401 fix. Retitled accordingly. The original `Refs #222` framing is withdrawn.
- ❌ It does **not** explain the nightly canary RED. That investigation is still open, and merging this as the canary fix would have been a false close.
- ✅ The code change is still defensible **on its own merits**: header construction was triplicated across three verbs and could diverge per-verb; every other lane in the tree already sends these headers; two independent reverse-engineered clients send them on this exact route. Consolidating is better code and the headers are demonstrably harmless (arm B: 200).

So this is a **refactor with a behaviour change that is not known to be necessary**. That is a much weaker case than the PR body originally made, and it is your call whether it is worth merging at all.

I would rather leave it draft than merge it under a claim the evidence does not support.

Evidence: `scripts/dev/_spike_out/spike_trpc_origin_referer_401_20260825_203605.json`

Issue #539, opened 2026-08-14
investigate: veo_3_1_lite_lower_priority — is it a cheaper/unmetered video tier, and is its picker selector still valid?


ffroliva:
## Why now

`VideoModel.VEO_3_1_LITE_LOWER_PRIORITY` is a registered model (aliases `veo-lite-lp`, `lite-lp`, `lower-priority`) that we ship, expose in `gflow models`, and have **never verified live**. During the #451/#288 capability spike it **missed its picker selector on both accounts**, and we deliberately left it alone rather than change a selector on a guess.

Two open questions, one cheap and one valuable:

1. **Does the model still exist in Flow's picker at all?** Our selector is `[role='menuitem']:has-text('[Lower Priority]')`. A miss means either the selector is stale *or* Flow removed the entry. We do not currently know which.
2. **What does it cost?** If "Lower Priority" is Flow's queued/relaxed tier, it may be materially cheaper — or unmetered — in exchange for wait time. That would be a significant capability for `gflow video`.

## What we actually know (2026-08-14, verified)

| Fact | Evidence |
|---|---|
| Picker selector missed on **both** accounts (`ffroliva`, `denon82`) | `scripts/dev/capture_video_model_capability_matrix.py` |
| The "menu items offered" capture came back **empty** | taken *after* the menu closed → **inconclusive**, not proof of absence |
| `reference_cap_for` = 3 (same as `veo_3_1_lite`) | `api/video.py` |
| `supports_duration()` = False | verified matrix |
| Excluded from `_resolve_chain_model` | `cli_video.py` |
| Credit cost | **unknown — never observed** |

Known costs of its siblings, from the live popover: `omni-flash` 7–15 (duration-scaled) · `veo-lite` 10 · `veo-fast` 20 · `veo-quality` 100.

## External context (contested — do not treat as settled)

Public sources describe a general "relaxed mode" pattern: a low-priority queue trading wait time for free or discounted generation. But at least one source states Veo 3.1 specifically **does not** support relaxed mode, being a pay-per-use API model with no queue-priority control. Others describe Flow's free tier as ~50 daily credits (180 in some countries) and Ultra as 25,000/month at ~10 credits per video.

So the "unlimited" framing is **unverified internet lore**. This ticket exists to settle it with our own evidence, not to adopt it.

## The decisive test is credit-free

Flow's settings popover renders a **live cost line** (`Generating will use N credits`) that updates with the selected model — we already read it in the capability spike. So:

[...]

PR #604, opened 2026-08-30
fix(transport): match Omni tier after Flow's "Omni 1.1 Flash" rename

kittinan:
## Summary

Screenshot at 2026-08-30 16-51-11

Flow renamed the video model picker entry **"Omni Flash" → "Omni 1.1 Flash"**. Playwright's `has-text` is a **contiguous** substring match, so a version number injected mid-label dropped `has-text('Omni Flash')` to zero matches. Every explicit `--model omni-flash` run then failed with `VideoModelSelectionError` (exit 18).

The fail-loud gate did its job — no credits were spent and nothing was billed to the wrong tier — but the model was unselectable. `v0.62.0` is on PyPI with this broken.

**Fix:** widen the selector to two ANDed `has-text` clauses that span the version segment, and keep the `[Lower Priority]` exclusion its `VEO_3_1_LITE` sibling already carries.

```python
# before
VideoModel.OMNI_FLASH: "[role='menuitem']:has-text('Omni Flash')",
# after
VideoModel.OMNI_FLASH: (
    "[role='menuitem']:has-text('Omni'):has-text('Flash')"
    ":not(:has-text('[Lower Priority]'))"
),
```

Both positive clauses must hold on the *same* menuitem, so it matches the old label, the current one, and the next version bump, while staying unique against the four `Veo 3.1 - *` entries. The `:not` matters because Flow already ships a `[Lower Priority]` variant of a tier: without it, an `Omni 1.1 Flash [Lower Priority]` entry would match two menuitems and put the model straight back at exit 18. If Flow ever offers two genuinely concurrent Omni *versions*, this resolves AMBIGUOUS — which `_select_video_model` already refuses before submit rather than guessing with `.first`.

**Not changed, deliberately:** the CLI alias (`--model omni-flash`) and the enum value (`omni_flash`). Those are gflow's own identifiers and a stable contract for chain files, resume state, and JSON output — renaming them would break resume on in-flight work.

## Changes

| File | What |
|---|---|
| `src/gflow_cli/api/transports/ui_automation_video.py` | The selector fix, plus the `[Lower Priority]` exclusion and the note on when *not* to widen a Veo anchor. |
| `tests/flow_selectors/test_model_governance.py` | Three new guards (below). |
| `tests/fixtures/flow_model_inventory.json` | Rewritten from a live probe read (see Evidence). |
| `KNOWN_ISSUES.md` + `website/docs/` | Correct the "brand names are stable anchors" claim. |
| `tests/api/transports/test_video_model_selection_loud.py` | Its 2026-08-26 menu snapshot marked historical, not current. |
| `scripts/dev/spike_omni_flash_i2v_ui_recon.py` | Same hardcoded literal in its A3/A4 probe, matched case-insensitively like `has-text`. |
| `CHANGELOG.md` | `[Unreleased] / Fixed`. |

### Test guards

- `test_omni_selector_survives_the_version_number_rename` grades the selector against **both** labels Flow has shipped. It uses recorded labels rather than the live fixture on purpose: the fixture holds only the **current** menu, so refreshing it makes a selector that only handles the new name look clean — and the next bump breaks it again.
- `test_omni_selector_matches_the_omni_entry_and_nothing_else` grades positively *and* negatively over a menu carrying both `[Lower Priority]` variants. Positively on purpose: `_matches` returns `[]` for a selector it cannot parse, so an excludes-everything assertion alone would bless a registry entry matching nothing at all.
- `test_the_live_omni_label_is_recorded_in_the_seen_list` pins the fixture's live Omni label into the recorded history, so the next probe refresh cannot quietly leave the guard grading labels nobody ships.

### Documentation correction

`KNOWN_ISSUES.md` claimed branded product names ("Veo 3.1 - \*", "Omni Flash") are confirmed-safe selector anchors. That holds for **translation** only — locale-stable is not version-stable.

The rule is narrower than "always match tokens": the discriminator is whether gflow's own identifier pins the version. `omni_flash` carries none, so `1.1` is noise the anchor must span. `VEO_3_1_FAST` makes `3.1` part of the model's identity, so a `Veo 3.2` swap **should** be a loud MISS — widening those anchors would silently bind `--model veo-fast` to a different tier at a different credit price.

## Evidence

The inventory fixture is a real probe read, not an observation: `scripts/dev/spike_model_inventory.py` on profile `ffroliva` (en), 2026-08-30, byte-confirms `volume_up Omni 1.1 Flash` and that the Veo labels did **not** move.

What is *not* byte-verified, and is labelled as such in the fixture: the rename **direction**. The `Omni Flash` baseline comes from a different account, locale and date (`denon82`, pt, 2026-08-26), so a single read cannot separate a global in-place rename from a cohort- or locale-gated label. The corroboration is that `has-text('Omni Flash')` stopped matching for the reporter on an account where it had worked. The selector is robust either way — it matches both labels.

## Test plan

- [x] `ruff check src tests` + `ruff format --check src tests` + `pyright src` — clean
- [x] `tests/flow_selectors` + `tests/api/transports/test_video_model_selection_loud.py` — 44 passed, 2 skipped
- [x] Full offline suite — green on CI across 3.11 / 3.12 / 3.13
- [x] **Live verification, zero credits.** `_select_video_model` driven against the live picker resolves and clicks the widened selector (`model_selected model=omni_flash`) and the trigger then reads `Omni 1.1 Flash`. That function never submits, so this proves the model binding without spending a credit.

Reviewed at `xhigh` after the two follow-up commits. Deliberately not changed: the branch prefix (`fix/` is not in AGENTS.md's list, but renaming closes this PR — next time), and the authorship of 6a8ea10, which is @kittinan's contribution and correctly signed off.

A separate Windows-only test-isolation bug found while running the local gates is tracked in #605; it does not touch this change.


ffroliva:

⚠️ Found **2** MUST-FIX issue(s). Please review and resolve them before merge.

<details>
<summary>View Sandboxed Review Output</summary>

Building Docker sandbox image...
Creating network triage-net-604...
2fdfce109f8563c9f703cf3520e90e17f1312b4de469918257d8520b4e6da4bd
Hardening network isolation for subnet 172.20.0.0/16 via iptables...
Launching sandboxed review for PR 604...
# PR #604 — Council Review Verdict (Autonomous Mode)

**REVIEWED_SHA:** `6a8ea10acd77744928c2a53cec4a033ee57cfa20` (head unchanged since dispatch — no divergence) · **Branch:** `fix/omni-flash-picker-rename` → `develop` · External contributor (Kittinan), labeled `needs-maintainer-review`.

## Consensus: 🟡 YELLOW

|---|---|---|
| D0 — CI mechanical gate | 🟢 GREEN | All 3 test-matrix jobs (3.11/3.12/3.13) pass, incl. Lint/Format/DocLinks/Hygiene/PII/Type-check steps (verified via `gh pr checks`; local `uv run` was sandbox-denied) |
| D1 — Correctness | 🟡 YELLOW | Fix is sound and root-cause-correct; missing `CHANGELOG.md` entry breaks this repo's own precedent (v0.61.0 #539 got one for the identical bug class) |
| D2 — Code quality | 🟢 GREEN | Style, naming, comments, typing all match established file conventions |
| D3 — Security | 🟢 GREEN | No injection vector, no secrets, no CI/dependency-trust surface touched |
| D4 — Tests | 🟡 YELLOW | Real, well-targeted regression tests against the production selector; automatic YELLOW per protocol for a UI-automation change with no live-verify evidence |
| D5 — Memory hygiene | 🟡 YELLOW | No must-fix; flags that the general "token-match, not whole-string" lesson isn't captured anywhere durable outside this one PR |
| D6 — UI/live-verify | 🟡 YELLOW | Selector logic and locale-invariance are consistent with existing sibling entries; live-verify against real Flow was not run (self-flagged by the PR author) |
| D9 — Docs drift | 🟡 YELLOW | Both `KNOWN_ISSUES.md` copies are byte-identical and technically correct, but were hand-duplicated instead of run through `scripts/ci/generate_website_docs.py` — the exact anti-pattern that generator was built to prevent after a past PII leak (#362). Harmless this time (no PII/links in the added text) |
| D13 — Dev scripts | 🟢 GREEN | Pure string-logic change in a non-shipped spike script; no Windows/lint/type/packaging impact |
| D14 — Over-engineering | 🟢 GREEN | No dead code, no speculative abstraction; comment density matches pre-existing file conventions |

## Must-fix (2)
1. **Missing CHANGELOG entry** — `CHANGELOG.md` `[Unreleased]` section (verified via `git show 6a8ea10a:CHANGELOG.md`) has no entry for this fix, breaking the precedent set by the identical bug class in v0.61.0 (#539). (D1)
2. **No live-verification evidence** — the PR body itself marks `gflow video t2v --model omni-flash` against real Flow as "Not run" and states it "should gate the merge." No structlog/screenshot/magic-bytes evidence exists. Per protocol this is an automatic gate for any UI-automation selector change. (D4 + D6, deduplicated — same finding, self-disclosed by the author)

## Nice-to-have (3)
1. Re-derive `website/docs/KNOWN_ISSUES.md` via `python scripts/ci/generate_website_docs.py` instead of hand-copying the hunk — no bug today, but sets a precedent the generator exists specifically to eliminate. (D9)
2. Capture the general lesson ("anchor on stable tokens via ANDed `has-text`, not whole branded strings, because names get versioned in place") as a durable memory entry — it exists only in `KNOWN_ISSUES.md` today, not in the reusable memory store. (D5)
3. Minor: stale `"volume_up Omni Flash"` label still appears in `tests/api/transports/test_video_model_selection_loud.py`'s mock data/docstring — harmless (that file doesn't derive matches from it), but now inconsistent with the rest of the cleanup. (D1)

## False positives
None — all 9 agents' claims were internally consistent; spot-checks (CHANGELOG gap, dual-doc byte-parity, generator/test existence, SHA non-divergence) all confirmed as reported.

## Confirmed-good (high-confidence positives)
- Root cause correctly fixed: `has-text('Omni'):has-text('Flash')` spans the version-token-injection drift shape, not just today's exact string; repo-wide grep confirmed no other hardcoded `"Omni Flash"` literal was missed.
- Ambiguity-refusal safety net (`_select_video_model` raising on >1 match) is real and verified in code — a future two-tier collision fails loud, not silently via `.first`.
- Both new tests (`test_omni_selector_survives_the_version_number_rename`, `test_omni_selector_does_not_collide_with_the_veo_tiers`) exercise the real production constant (not a copied literal) and would have caught this exact regression.
- Fixture amendment is honestly and explicitly labeled as an unverified maintainer observation, not passed off as a probe capture — provenance metadata intact.
- CLI alias (`--model omni-flash`) / enum value (`omni_flash`) confirmed untouched — no resume/chain-file breaking change.
- No security-relevant surface touched; no secrets, no CI/dependency-trust changes.

## Memory actions (D5)
- **ADD**: a durable entry capturing "Playwright `has-text` is contiguous-substring; anchor selectors on stable tokens via ANDed clauses, not whole branded strings, because product names get versioned in place" — currently unrecorded outside this PR's `KNOWN_ISSUES.md`.
- **ADD** (lower priority): a pointer tracking the still-outstanding `spike_model_inventory.py` re-run needed to byte-verify the `Omni 1.1 Flash` fixture edit and confirm the Veo labels didn't also move.
- No UPDATE/DELETE needed — nothing existing is contradicted.

## Next step — live validation
Static review is green modulo the two must-fix items above. Per this sandbox's autonomous-mode ceiling, **no live e2e or credit-spending verification was run or is possible here** (no network egress to Flow, no credentials). The final gate — `gflow video t2v --model omni-flash` against real Flow, run deliberately last and outside this sandbox with real credentials — is expected to be the true test of whether the widened selector resolves against the renamed picker; a bounce there is the process working, not a review miss. Do not treat this council's YELLOW-modulo-fixes as merge-ready without that run.

`SUMMARY_VERDICT: YELLOW | MUST_FIX_COUNT: 2 | PR_URL: https://github.com/ffroliva/gflow-cli/pull/604`
Cleaning up network rules and Docker network...

</details>

ffroliva:
Thank you for this, @kittinan — this is an unusually good first contribution.

Two things in particular stood out. You found the *shape* of the bug rather than
just the instance: "`has-text` is a contiguous match, so a version number
injected mid-label drops it to zero" is the real defect, and widening to two
ANDed clauses is the right fix rather than swapping in today's exact string.
And you left the CLI alias and the enum value alone, for the right reason —
those are our contract for chain files, resume state and JSON output, and
renaming them would have broken resume on in-flight work. That was the correct
call and it is not an obvious one.

I also want to call out the "Evidence caveat" section of your PR description.
You edited the inventory fixture from an observation rather than a probe run,
and instead of quietly letting it pass you labelled exactly what was and wasn't
verified, and said so where a reviewer would see it before merging. That is the
behaviour this repo's fixture rule is trying to produce. It is much easier to
review an honest gap than to discover a hidden one.

**I have taken this over from here** so you are not blocked on the two things
that needed live account access — both are now done, and I pushed the follow-up
to your branch as `626f330`.

**Your fixture edit was byte-exact.** I re-ran `spike_model_inventory.py` on a
live profile (navigation and menu clicks only, zero credits). The picker renders
`volume_up Omni 1.1 Flash`, exactly as you recorded it, and the four Veo labels
did **not** move. I have rewritten the fixture from that read and removed the
caveat note, so your amendment is now backed by a probe capture rather than a
report. Nothing about your line was wrong — only its provenance changed.

**The live gate is closed, and it cost nothing.** You flagged
`gflow video t2v --model omni-flash` as the merge gate. It turns out we did not
need to spend a credit to get there: `_select_video_model` opens the picker,
counts matches and clicks — it never submits, so only Generate spends. Driving
it against the live picker gives:

```
model_selected model=omni_flash via="[role='menuitem']:has-text('Omni'):has-text('Flash')"
picker trigger now reads: 'Omni 1.1 Flash'
```

Your widened selector resolves, clicks, and binds the correct tier on the real
DOM. Worth remembering for future selector work here — the model-binding path
can be verified for free, and only the generate path actually costs.

Two review items I fixed on top of your work:

1. **Missing `CHANGELOG.md` entry.** Easy to miss as an outside contributor;
   the identical bug class got one in a previous release, so I have added it
   under `[Unreleased]`.

2. **The `KNOWN_ISSUES.md` lesson was too broad.** As written it read as a
   blanket "match stable tokens, never whole strings" — but the four
   `Veo 3.1 - *` selectors two lines below deliberately keep the contiguous
   `3.1`, so the doc contradicted the code it sat next to. The discriminator is
   whether *our own* identifier pins the version: `omni_flash` carries none, so
   the version in Flow's label is noise the anchor must span; `VEO_3_1_FAST`
   makes `3.1` part of the model's identity, so if Flow swaps that tier for a
   `Veo 3.2` a loud MISS is the **correct** outcome — widening those anchors
   would silently bind `--model veo-fast` to a different tier at a different
   credit price. I have narrowed the wording and put the same note at the
   selector registry, where the next person will actually read it.

That second one is a subtlety in our own docs that your PR surfaced, not a
mistake on your part.

CI is green across 3.11/3.12/3.13. Merging to `develop` now — it will ship in
the next release. Thanks again.

ffroliva:
Shipped — your fix is live in **[v0.62.1](https://github.com/ffroliva/gflow-cli/releases/tag/v0.62.1)**, on PyPI now:

```bash
pip install --upgrade gflow-cli   # 0.62.1
```

Thanks again. Two follow-up commits landed on top of yours before merge, and it's only fair to say what changed and why, since one of them was a real gap:

**The selector picked up a `:not` clause.**

```python
"[role='menuitem']:has-text('Omni'):has-text('Flash'):not(:has-text('[Lower Priority]'))"
```

Your two ANDed clauses were the right call — widening to span the version segment rather than swapping in today's exact string is what makes this survive the *next* bump. What they lost was the `[Lower Priority]` exclusion that `VEO_3_1_LITE` carries three lines below. Flow already ships that suffix for Veo Lite, and #539 is open precisely because the LP tier looks cohort- or region-gated, so it can turn up for Omni on any account at any time. The moment it did, `'Omni'` + `'Flash'` would match two menuitems, `_select_video_model` would raise AMBIGUOUS, and `--model omni-flash` would be unselectable again — the same outage, from the same drift.

Not a criticism of the fix. It's the kind of thing you only see with the Veo Lite entry side by side in your head.

**Live-verified before shipping, at zero credits.** Model selection happens in the settings panel entirely before submit, so the whole matrix is free to run. On profile `ffroliva`, 2026-08-30:

```
model_selected model=omni_flash
  via="[role='menuitem']:has-text('Omni'):has-text('Flash'):not(:has-text('[Lower Priority]'))"
```

6/6 as expected — all four offered tiers select, a tier Flow doesn't offer refuses, and a deliberately ambiguous selector refuses rather than resolving `.first`. The transport's own refusal diagnostic read the menu back as `Omni 1.1 Flash / Veo 3.1 - Lite / Fast / Quality`, which independently corroborated the fixture. Full ledger in [`docs/LIVE_VERIFICATION_v0.62.1.md`](https://github.com/ffroliva/gflow-cli/blob/main/docs/LIVE_VERIFICATION_v0.62.1.md).

**Also corrected, and this one was mine, not yours:** the KNOWN_ISSUES paragraph I wrote illustrated the *do-not-widen* rule with `has-text('Veo'):has-text('Flash')`. No Veo label contains "Flash" — that example selector matches nothing, so anyone following the doc would have broken `--model veo-fast` with the exact bug the paragraph was documenting. It's `'Veo'` + `'Fast'` now, matching what the selector-registry comment already said.

One note for next time, no action needed here: the branch prefix. `AGENTS.md` lists `feature/`, `bugfix/`, `hotfix/`, `chore/`, `docs/`, `test/`, `release/` — `fix/` isn't one of them, and `bugfix/` was the intended prefix. Renaming would have closed this PR, so it stayed as-is.

Genuinely good first contribution — you found the shape of the bug rather than the instance, and that's the harder half.

ffroliva:
Thank you for this — and the screenshot is what made it land in one pass rather than three.

`has-text` being a contiguous match, so a version number injected mid-label silently drops it to zero matches, is a genuinely nasty failure shape: nothing errors, the tier just never selects.

Worth being honest about something: v0.60.0 added a selector registry and a nightly CI probe against the live UI precisely so this class of drift gets caught by a scheduled run rather than by a person. Flow renamed the label and **you** caught it before the probe did. That's a gap on my side, not a compliment I can take — I'm going to go look at why the probe didn't flag the picker entry.

If you're willing: what are you building with gflow-cli? You're one of a small number of people who have actually shipped code into it, and I've never asked any of you what you're using it for. That's been a mistake — the roadmap so far is essentially my own guesses.

kittinan:
I have created an automated affiliate product review system using my avatar with gflow-cli.

PR #618, opened 2026-09-01
fix(video): the referenceEntity guard was never running (#615)

### LINKED ISSUE #615: referenceEntity guard never runs: route patterns cannot match flowMedia:batchGenerateImages (and page.route misses the Web Worker)
### Summary

`_intercept_reference_entities` never runs. Its route patterns cannot match the
real endpoint URL, so the referenceEntity guard is a silent no-op on every image
and video generation.

This is not the UI rollout of #174 — it is independent of which Flow UI cohort you
are in, and it disables a safety net rather than a feature.

### Two independent reasons the handler never fires

**1. The glob never matches the URL.**

The endpoint is

```
https://aisandbox-pa.googleapis.com/v1/projects/<project>/flowMedia:batchGenerateImages
```

The last path segment is `flowMedia:batchGenerateImages`, not `batchGenerateImages`.
The registered pattern

```python
await page.route("**/batchGenerateImages", intercept_generation_request)
await page.route("**/batchAsyncGenerateVideo*", intercept_generation_request)
```

requires the final segment to be exactly `batchGenerateImages`, so it never matches.

The response listener is unaffected because it does a substring test
(`if route in response.url`), which is why responses are captured normally and the
problem is invisible in normal runs.

**2. The request is Web-Worker-delegated, so `page.route` is the wrong level.**

Even with a matching pattern, `page.route` does not see it. The comment right above
the call site already records this for the capture path:

> page-level network capture is dead in this cohort — requests are Web-Worker-delegated

`BrowserContext.route` does see these requests.

### Why this matters beyond a missing feature

`_intercept_reference_entities` is described in the code as the guard that

> programmatically filters outgoing batchGenerateImages requests so that only the
> explicitly requested reference_entities are permitted to go to the server,
> preventing stale or poisoned entities from smuggling themselves into unrelated runs

That protection has never fired on this account. `ui_automation.batch_request_modified`
was never logged in any run, before or after the change below.

### Evidence

Three `gflow image t2i` calls with a prompt that names **no** appearance detail at
all — only "the shop assistant behind the counter of a 1940s British shoe shop".

| run | route registration | `batch_request_modified` logged | result |
|---|---|---|---|
| 1 | `page.route` + glob (stock) | no | — |
| 2 | `page.route` + `context.route`, glob | no | — |
| 3 | both levels, predicate matcher | **yes** | entity honoured |

In run 3 the generated portrait matches the character entity's face image exactly
(face shape, side parting, collar and waistcoat). A control run of the identical
prompt without `--reference-entity` produced a completely different man — older,
grey hair, brown jacket, a cobbler's workshop instead of the shop. So Flow does
evaluate `referenceEntities` when the field actually reaches the wire.

### Fix that worked here

Replace the glob patterns with predicates and register on both levels:

```python
def _matches_images(url: str) -> bool:
    return "batchGenerateImages" in url

def _matches_video(url: str) -> bool:
    return "batchAsyncGenerateVideo" in url

await page.route(_matches_images, intercept_generation_request)
await page.route(_matches_video, intercept_generation_request)
context = getattr(page, "context", None)
if context is not None:
    await context.route(_matches_images, intercept_generation_request)
    await context.route(_matches_video, intercept_generation_request)
```

with the matching `unroute` calls in the `finally` block. Same matcher objects, so
`unroute` still works.

### Side note on #174

With the handler actually running, the same hook can also *add* the requested
entities when the UI could not stage them — which is a workable path for accounts on
the new media-library UI, where the context menu offers only "Delete". Whether you
want that in the tool is your call; the point of this issue is only that the guard
itself is currently dead.

### Environment

- gflow-cli 0.62.0
- Windows 11, Chromium via the bundled profile
- German-language Flow UI, new full-page media library (the #174 cohort)

Project and entity ids are redacted above; no tokens or signed URLs are included.

ffroliva:
Closes #615.

Reported by @DioServis with both root causes correctly separated. I verified each against the source before touching anything.

## The defect

`_intercept_reference_entities` registered `page.route("**/batchGenerateImages", ...)`, and it could never fire — for two independent reasons:

**1. The glob never matched.** `routes.batch_generate_images_url()` emits:

```
.../projects/{id}/flowMedia:batchGenerateImages
```

A `**/batchGenerateImages` glob requires the final path segment to equal `batchGenerateImages` exactly. The real segment is namespaced, so it never matched.

**2. Page-level routing is the wrong level.** These requests are Web-Worker-delegated in the current Flow cohort, so `page.route` cannot see them at all. The capture path already recorded this (`ui_automation.py:2881`, `:3087`) and the guard was left at page level anyway.

**Why nobody noticed:** the response listener does a substring test, so it kept working normally. The guard failed *open*, silently, on every image and video generation — a safety net reporting for duty and doing nothing. That is worse than a missing feature.

## The fix

- register on `page.context.route` with `_GENERATION_ROUTE_RE`, a substring matcher that matches the way the response listener already did
- unroute the same matcher

## The test was the real culprit

The existing test asserted that a pattern *string* had been registered, then invoked the handler by hand. It stayed green for months while the guard never fired once.

It now asserts context-level registration, **refuses** page-level registration, and — the part that would actually have caught this — checks the matcher against URLs the code really builds. A second test pins the defect itself so it cannot quietly return.

## Scope — one thing this deliberately does *not* fix

Direct-wire routes issued through `client._post_json` use `page.request.post(...)` (Playwright's **APIRequestContext**), which route handlers never observe — those requests leave from the Python side using the context's cookies and never enter the browser's network stack. So this guard covers browser-initiated generation only, by construction.

That is not a problem today (the direct-wire routes that exist send no `referenceEntities`), but it is a structural blind spot for any future direct-wire route that does. Filed separately rather than smuggled into this PR.

## Test plan

- [x] `ruff format` / `ruff check` clean
- [x] 191 tests pass in `tests/api/transports/test_ui_automation.py`
- [x] new regression test fails against the old glob (asserts the real segment is `flowMedia:batchGenerateImages`)
- [ ] live verification on a real generation run — the guard should now log `ui_automation.batch_request_modified` when a poisoned entity is stripped, which it has never once done


ffroliva:
## Council review — 🟡 YELLOW, must-fixes applied in `9646c5b`

Seven dimensions plus the mechanical CI gate. Verdict was YELLOW; five must-fixes are now addressed, two findings were verified as false positives and dropped, one is filed separately.

| Dimension | Verdict |
|---|---|
| CI-mechanical | 🟢 ruff / format / doc-links / hygiene clean |
| Correctness | 🟡 CHANGELOG missing; Service-Worker caveat |
| Code quality | 🟡 dead capture group; explanation shipped 3× |
| Security | 🟡 `json.dumps` alters wire shape |
| Tests | 🟡 **matcher never bound to the call site** |
| Memory hygiene | 🟢 |
| UI / live-verify | 🟡 root cause #2 unprovable without a live run |
| Over-engineering | 🟡 ~30 lines of prose for a 2-line fix |

### The finding worth reading

The registration assertion was `assert_any_call(ANY, ANY)` — it accepts *any* first argument. The regex was verified in isolation and `context.route` was verified to be called, but **nothing asserted production passes that pattern**. The old glob string could have been handed to `context.route` with `_GENERATION_ROUTE_RE` left intact, and both new tests would still pass with the bug fully restored.

That is the same defect shape this PR exists to fix, reintroduced one level up. Fixed.

### False positives (verified, dropped)

Two reviewers independently argued that context-wide routing cross-talks between pooled pages — Playwright dispatches to the newest matching handler and this one never calls `fallback()`, so run A's request would be filtered against run B's `expected_entities`.

It does not, today: one transport per client, one `_generate_lock`, every generation entry point acquires it, `_intercept_reference_entities` sits inside those serialized bodies, and there is no `gather` / `create_task` / `TaskGroup` fan-out anywhere in either transport.

But it is a correct hazard with the wrong severity, and it made the lock **load-bearing and undocumented** — an invariant sitting directly opposite a `concurrency` setting that accepts up to 16. A `ponytail:` comment now names the ceiling and the upgrade path, including that `route.request` has no owning page for Worker-delegated requests, so per-page scoping would need a different key.

### Also applied

- wire shape of the rewritten body now matches the SPA's own `JSON.stringify` (`separators`, `ensure_ascii=False`) — the default adds spaces and escapes non-ASCII, a gratuitous bot-signature difference on a rewrite path
- dead capturing group dropped
- triplicated explanation split: constant carries the pattern-shape reason, call site the context-level reason
- `VIDEO_GENERATE_ROUTES` iterated instead of a one-element tuple
- CHANGELOG entry, credited to @DioServis

### Still open, honestly

**Root cause #2 cannot be proven by any test in this repo.** A mock has no Web Worker; the unit tests assert registration level, never visibility. And `context.route` does **not** intercept Service-Worker-owned requests, with nothing setting `service_workers="block"` — if Flow delegates via a Service Worker rather than a dedicated Web Worker, this fix is still a no-op. Only a live run distinguishes them.

The live-verify box stays unchecked. The only event that proves the handler fired is `ui_automation.batch_request_modified` on a run carrying a poisoned entity — a clean run emits nothing, so "ran a generation and nothing broke" is not evidence.

Filed separately: **#620** — `tests/e2e/test_entity_smuggling_e2e.py` asserts inside `if bodies:` and has been passing on an empty loop. Same vacuous-assertion defect as #615, one directory over, present throughout and never caught it. Live-gated, so not changed blind from here.

ffroliva:
## Unblocked — the Worker question is answered by measurement, at zero cost

The hold on this PR was: `context.route` sees a **dedicated** Web Worker but not a **Service Worker**, so if Flow delegates via the latter this fix is a no-op that passes every test and changes nothing in production. No offline test could tell those apart.

It can now be settled by a run, because #637 makes the guard announce itself on **every** intercepted request rather than only when it strips something. "Never ran" and "ran, nothing to strip" used to be byte-identical silence; absence is now evidence.

### The A/B

Identical test, account, and prompt against `denon82`. One variable: whether this PR's route fix is present. Both branches carry #637's signal, or the control could not be read.

| Branch | Registration | Result |
|---|---|---|
| control — **this PR absent** | `page.route("**/batchGenerateImages")` | **FAILED** — `The referenceEntity guard never ran` |
| **this PR present** | `page.context.route(_GENERATION_ROUTE_RE)` | **PASSED** — guard fired |

```
control : 1 failed in 51.25s   (image generated; no batch_request_intercepted)
with fix: 1 passed in 67.72s   (image generated; guard observed the submit)
```

**It is a dedicated Web Worker.** `context.route` is sufficient; no `service_workers="block"` is needed. Merge is unblocked.

Two things the run establishes that no mock could:

- **#615 reproduces live** — the control generated an image successfully while the guard never saw the request. Confirmed on the affected surface, not merely reasoned about.
- **`route.continue_(post_data=...)` does not corrupt the body.** Both runs produced a real image. This was the quieter risk: if the rewrite mangled the request, the fix would have broken generation outright, and nothing offline exercises that path.

Cost: zero. Image generation is daily-capped, not billed — the repo's own markers said otherwise, which is part of why this went unverified for so long; corrected in #637.

Full ledger, including what is **not** covered: [`docs/LIVE_VERIFICATION_reference_entity_guard.md`](https://github.com/ffroliva/gflow-cli/blob/bugfix/entity-smuggling-e2e-asserts-620/docs/LIVE_VERIFICATION_reference_entity_guard.md).

### Two things worth folding in before merge

**1. The video glob was equally dead, and the description doesn't say so.** Proven offline with Playwright's own `glob_to_regex_pattern`:

```
'**/batchAsyncGenerateVideo*' -> ^(.*/)batchAsyncGenerateVideo([^/]*)$
    https://…/v1/video:batchAsyncGenerateVideoText   -> NO MATCH
```

This PR's substring regex fixes both — verified — but the changelog entry names only `batchGenerateImages`. Worth stating, since it doubles the scope of what was broken.

**2. "Requests are Web-Worker-delegated in the current Flow cohort" is narrower than written.** The `page.on('request')` listener at `ui_automation.py:2554` does observe `batchGenerateImages` on the classic path and is documented as the entity spike's make-or-break signal. The "page-level capture is dead" comment sits inside the *agentic* branch. On classic, the **glob was the whole bug**. Doesn't change the fix — `context.route` is strictly broader and now proven — but the reasoning shouldn't outrun the evidence.

### Merge order

Land this first, then #637. #637's e2e correctly goes red until this lands (it is detecting #615, exactly as designed). This branch also needs a rebase — its merge-base is ~19 commits behind `develop`; the only conflict is a union-resolvable constant block from #626, and I verified the two PRs do not conflict with each other.

Issue #639, opened 2026-09-02
Flow's migrated flow.google.com frontend: text-to-video is driven (v0.67.0); image, i2v/r2v, characters, scenes, extend, instructions, tools and project creation still pending


maipmacrothorax-75:
## Summary

Google is moving Flow off Labs onto its own domain. On a **migrated** page load,
`https://labs.google/fx/tools/flow/project/<id>` redirects to
`https://flow.google.com/project/<id>` and serves a rewritten frontend that contains **zero
`<i>` elements**. Since cohort detection and every mode control are expressed as Material
Symbols ligature selectors (`i.google-symbols:text-is(...)`), nothing matches:
`detect_ui_mode` falls through to its classic default, the classic driver then misses
`mode_switch_trigger`, and the run dies with `UiSelectorDriftError`, exit 23,
`retryable: false`. Image and video alike.

**The migration flaps per page load.** The same account, same profile and same project land on
the old host on one load and the new host on the next, minutes apart, with no client change.
On an old-host load everything works end to end. That is almost certainly why this has not
been reported yet: whether you can reproduce it depends on which side your load happens to get.

`flow.google.com` appears nowhere in the source. Reproduced on 0.59.0, 0.62.1, 0.63.0 and
0.65.0 — nothing in between touches it. The incident bundle's own `ui.json` already flags the
host as unrecognised: `url.host_category: "other"`, `url.route: "other"`.

This is not #493 recurring and not the agentic cohort — the agentic indicators are absent too.
It is a different origin serving different markup.

## The two sides, measured

Same profile, same project, ~35 minutes apart, gflow **0.63.0** on both, nothing changed
locally in between:

| | old host | migrated host |
|---|---|---|
| final URL | `labs.google/fx/<locale>/tools/flow/...` | `flow.google.com/project/...` |
| `document.querySelectorAll('i').length` | **55** | **0** |
| `i.google-symbols` | **49** | **0** |
| `crop_*` trigger | **present** | **absent** |
| agentic indicators | absent | absent |
| `gflow image t2i` | `mode_switch_trigger` matched, `image_mode_tab` matched, NARWHAL, 9:16, submitted, HTTP 200, image downloaded, **exit 0** | `UiSelectorDriftError`, **exit 23** |

So the selectors are not stale in themselves — they are correct for the host they were written
against, and that host is being replaced under them.

Other notes on the migrated page: `console` is clean (0 errors — it is a healthy page, not a
[...]

Issue #634, opened 2026-09-02
video chain and movie run die mid-spend on duration × model — and no chain-usable model has a duration control


ffroliva:
## Summary

`gflow video chain` and `gflow movie run` carry the same `duration` × `model` defect that #632 fixed for `video i2v` (#630) — but on these two surfaces it dies **mid-spend**: earlier links/scenes have already rendered and real credits are gone before the crash, and it surfaces as exit 1 `"Unexpected error … file a bug"`.

For `chain` it is worse than a mismatch: **no model a chain can use has a duration control at all**, so *any* manifest `duration` is a guaranteed crash.

## Root cause

`_reject_duration_without_control` lives in `cli_video.py:166` and is called from exactly three places — `cli_video.py:1151`, `:1329`, `:1501` (t2v / i2v / r2v). Neither pipeline calls it:

```
$ grep -n "supports_duration" src/gflow_cli/movie_manifest.py src/gflow_cli/chain.py src/gflow_cli/cli_movie.py
(no matches)
```

So the mismatch is never caught at parse or validation time. It falls through to the DTO's bare `ValueError` (`api/video.py:331-339`) at link/scene render time, which the CLI reports as `error_unhandled` → exit 1 — the same lost-explanation failure #630 described, only now after money has been spent.

**Why chain is categorically broken, not just occasionally:**

- `VideoModel.supports_duration()` returns True for `OMNI_FLASH` only (`api/video.py:79-94`).
- `chain.py:326` rejects `OMNI_FLASH` outright ("wire-verified for SINGLE-clip i2v only").

The intersection of "models a chain accepts" and "models with a duration control" is empty. Every `duration` in a chain manifest crashes.

For `movie`, `_parse_scene_numeric_fields` (`movie_manifest.py:507-520`) validates `duration` against `_VALID_DURATIONS` — a **value** check only. `duration = 4` with a Veo model passes manifest validation cleanly and dies later at render.

## Repro

```bash
cat > story.jsonl <<'EOF'
{"prompt": "a lone wolf on a snowy ridge at dawn, cinematic", "model": "veo-lite", "duration": 4, "aspect": "16:9"}
{"prompt": "it lifts its head and turns to face the camera"}
EOF
gflow video chain story.jsonl
```

Link 1 renders and bills. Then exit 1, `"Unexpected error"`.

This manifest is copied verbatim from `docs/USAGE.md:847` — see the companion issue on the documentation entrenchment.

## Expected

[...]

Issue #630, opened 2026-09-02
cli: --duration without --model on i2v dies as "Unexpected error" (exit 1) instead of the usage error that already exists


ffroliva:
## Symptom

`gflow video i2v --initial-frame A --end-frame B "prompt" --duration 4` (no `--model`) exits **1** with:

```
Unexpected error. Re-run with GFLOW_CLI_DEBUG_TRACEBACK=1 to see the real
error. If this persists, file a bug at https://github.com/ffroliva/gflow-cli/issues.
```

With `GFLOW_CLI_DEBUG_TRACEBACK=1` the real cause is visible and perfectly clear:

```
ValueError: model 'veo_3_1_lite' has no duration control in Flow's UI, so ...
```

The user is told to file a bug for a plain usage error whose explanation already exists.

## Root cause

`_reject_duration_without_control` (`src/gflow_cli/cli_video.py:166`) exists precisely to stop this. Its own docstring says so:

> The DTO guards this too (defence in depth for API callers), but a bare `ValueError` surfaces through the CLI as "Unexpected error." (exit 1) and the explanation is lost. Raising a `UsageError` here gives exit 2 and prints the reason.

But it returns early when the user did not pass `--model`:

```python
if duration is None or model is None:
    return
```

For `video i2v` an omitted `--model` is not "no model" — it defaults to `I2V_DEFAULT_MODEL` (`veo_3_1_lite`), which renders **no duration control**. So the guard is bypassed in exactly the default case, and the DTO's `ValueError` escapes as `error_unhandled`.

Called at `cli_video.py:1151`, `:1329`, `:1501`.

## Impact

Every user who passes `--duration` to `gflow video i2v` without also passing `--model` — the most natural way to try the flag — gets "Unexpected error, file a bug" instead of the actionable message. Exit 1 instead of the intended exit 2.

`t2v` differs: with no `--model` the run inherits Flow's *UI* default, which gflow does not control, so the effective model genuinely is unknown there. Any fix must not assume otherwise.

## Suggested fix (two options)

1. **Narrow:** have the i2v call site pass its effective default (`I2V_DEFAULT_MODEL.value` when `model is None`) into the guard. Smallest diff; leaves `t2v` semantics untouched.
[...]

PR #650, opened 2026-09-03
feat(video): support duration selection (4s, 6s, 8s) for Veo 3.1 models

stgmt:
## Summary
Adds one canonical duration contract across API, CLI, chain, movie, and MCP: Veo 3.1 accepts 4s/6s/8s, while 10s remains omni-flash-only. Flow UI exposure is account/cohort-dependent, so this is not a claim that every account renders the controls.

## Live evidence
- Positive: `presentation-reels-google`, `labs.google`, 2026-09-04 — `veo_3_1_lite`, `veo_3_1_fast`, and `veo_3_1_quality` exposed 4s/6s/8s; no Generate action; $0 spent.
- Receipt SHA256: `ac394db744ddcba301764057f9761325b1dfd345b5474f2a5cd83bbbeb299529`; full JSON is attached in the PR discussion.
- Negative control: the maintainer's `denon82` capture exposed no Veo duration row. Both captures are valid for their cohorts.
- `veo_3_1_lite_lower_priority` was absent from the positive picker, so this PR makes no positive live-UI claim for that model.

## Changes
- Centralizes model/duration validation in `validate_duration_for_model` and reuses it across all request edges.
- Accepts Veo 4s/6s/8s for single clips, chains, movies, and MCP; rejects invalid values and Veo 10s before queued/rendered work.
- Aligns CLI help, project status, known issues, canonical docs, and generated website mirrors with the cohort-dependent evidence.
- Isolates CLI tests from browser/incident side effects and verifies every documented duration.

## Verification
- `uv run pytest tests/cli/test_cli_video_chain.py tests/cli/test_cli_video.py tests/cli/test_movie_manifest.py tests/mcp/test_server.py`
- `uv run ruff check src tests`
- `uv run ruff format --check src tests`
- `uv run pyright src`
- `uv run python -m pytest -q --cov=gflow_cli --cov-report=xml --junitxml=pytest-report.xml`
- Documentation link, PII, mirror-sync, and test-count gates pass.


ffroliva:
## Council review — `75f663c` · consensus 🔴 RED

Thanks for this, and for touching all five surfaces plus both doc mirrors — that breadth is the right instinct and most contributions miss it. Ten review dimensions ran; below is everything that needs to change before this can merge, ordered by what blocks first.

### The gating question (needs a maintainer or an eligible account)

The PR's premise — *"Flow now renders 4s/6s/8s tabs for all Veo 3.1 models"* — reverses a finding this repo captured **live**, on two Google accounts and two locales, with screenshots: `docs/superpowers/spikes/2026-08-14-video-model-capability-matrix.md` (PRs #534/#535), which records duration tabs as **absent** for `veo_3_1_lite/fast/quality`. That spike file is still in the tree and untouched by this PR, so merging as-is would ship a repo asserting both things.

Per [AGENTS.md](../blob/develop/AGENTS.md) ("Working discipline"), a blackbox claim without a capture is **LIKELY, not CONFIRMED** — this project drives a private API it does not own, so offline green never settles a Flow-behaviour question.

**This is cheap to settle and costs $0 in credits:** re-run `scripts/dev/capture_video_model_capability_matrix.py` and paste the resulting matrix here. If it shows the tabs, the premise is proven and the rest is mechanical. If you can't run it (it needs a Flow-authenticated Chrome profile), say so and a maintainer will.

Worth noting the downside is bounded: if the claim turns out wrong, `configure_video_settings` raises `UiSelectorDriftError` (exit 23) *before* submit, so no credits burn — you'd get a confusing error, not a wrong-length billed clip.

### Blockers you can fix now

**1. CI never reaches pytest.** `uv run ruff format --check src tests` fails — 4 files: `tests/api/test_video_request.py:114`, `tests/cli/test_cli_video.py:1648`, `tests/cli/test_movie_manifest.py:555`, `tests/mcp/test_server.py:848` (missing blank lines). All three `test` jobs die at that step, so the "204 tests pass" result isn't reproduced in CI. `uv run ruff format src tests` fixes it.

**2. One test outside the diff now fails.** `tests/cli/test_cli_models.py:36` pins `by_vid["veo_3_1_fast"]["max_duration"] == 0`, fed from `cli_models.py:69` → `max_duration_for()`, which this PR changes to `8`. Verified: `assert 8 == 0`. (A full run of `tests/cli tests/mcp tests/api` on this head gives 8 failures, but the other 7 are pre-existing order-dependent pollution in `test_client_incidents.py` — they fail on `develop` too and are not yours.)

**3. `i2v` with no `--model` now accepts `duration=10`.** Verified against this head:

```
GenerateVideoRequest(mode=I2V, start_image="a.png", duration=10)  -> ACCEPTED
GenerateVideoRequest(model=VEO_3_1_FAST, duration=10)             -> rejected ("caps at 8s")
```

`_validate_duration` (`api/video.py:398`) gates on `self.model`, which is `None` here. `_validate_model_capabilities:314-317` resolves `effective = I2V_DEFAULT_MODEL` specifically to close this hole (#630) — but that resolution now feeds only the `supports_duration()` branch, which can no longer fire. The CLI and MCP entry points each re-resolve the default so they still catch it; the shared DTO does not, so direct `FlowApiClient` callers slip through.

**4. Chain pre-flight regression — this one can cost credits.** `chain.py:233` narrowed from rejecting *any* duration to `spec.duration == 10`. A manifest link with `duration: 7` now passes pre-flight and dies at the DTO **mid-loop, after earlier links have already rendered and billed**, as a bare `ValueError` → exit 1 "Unexpected error". Confirmed: `reject_unusable_links` accepts `duration=5`. The docstring this PR deleted described exactly that failure. Fix: reject `duration not in (4, 6, 8, 10)` in `reject_unusable_links`.

**5. Five dead branches, two of them printing false messages.** `supports_duration()` is now `return True`, so every `if not ...supports_duration()` is unreachable: `api/video.py:318`, `cli_video.py:206`, `mcp/tools.py:1064`, `movie_manifest.py:556` — plus `api/video.py:191`'s `else 0`. Two still say "the Veo 3.1 models render no duration row at all". The surviving rule is only *"10 ⇒ omni_flash"*, which `_validate_duration` already owned before this PR; the diff hand-rolled it four more times, in four different wordings. Deleting the predicate and its callers is about **-41 lines** and leaves one source of truth.

**6. Six more places still state the old rule.** `docs/USAGE.md:499` contradicts line 503 *in the block this PR rewrote*; also `docs/USAGE.md:843-851`, `docs/MCP.md:86`, `docs/MOVIE.md:60,72-79`, `docs/PROJECT_STATUS.md:114,242`, `KNOWN_ISSUES.md:270-293` (still Status: Open, #288), `cli_video.py:1589` (`video chain --help`), `cli_movie.py:100-102,122` (comments shipped into every generated `movie.toml`), `chain_manifest.py:13-17`. `docs/MCP.md:86` matters most — MCP tool docs are what an LLM agent reads to decide whether a call is legal, and it currently promises a 400 that no longer happens. Mirror every edit into `website/docs/`, and add a `CHANGELOG.md [Unreleased]` entry.

### Smaller items

- `veo-lite-lp` (`VEO_3_1_LITE_LOWER_PRIORITY`) has no duration test on any surface, though the PR claims support for it.
- Four tests still pass if the source change is reverted — `test_chain.py:318`, `test_cli_video_chain.py:564`, `test_server.py:835,847` assert only `"duration" in ...`. Assert the new substrings.
- No accept-path test for CLI, chain, or MCP — the relaxation's actual point.
- `tests/cli/test_cli_video.py:1649` dropped `assert "Unexpected error" not in result.output`, which was the #630 guard.

### What's already solid

- Supply chain clean: exactly the 13 declared files, no dependency or workflow changes.
- The MCP queued-payload round trip is intact (`tools.py:921` writes `duration`, `codec.py:244` reads it) — a string mismatch there type-checks clean and silently no-ops, and yours lines up.
- `--duration` help text is byte-identical across `t2v`/`i2v`/`r2v` — none missed.
- `click.Choice` and the default are unchanged, so users who omit `--duration` see no behaviour change.
- Keeping `10s` omni-only was the right call, and it holds on all five surfaces.
- Both doc mirrors were edited together.

Happy to re-run the council once the capture matrix is attached and the format gate is green.

ffroliva:
## Capture run done — the matrix is unchanged. This PR's premise doesn't hold for the frontend gflow drives.

I ran the credit-free recon myself rather than leave you waiting on it. **$0 spent — navigation and popover DOM reads only, nothing submitted.**

```
profile: denon82 (pt-BR, page_lang=pt) · project 2ddc3a33 · 2026-09-03 22:54:59Z
scripts/dev/capture_video_model_capability_matrix.py

  MODEL                        DURATION TABS        COUNT  CREDITS  composer chip
  omni_flash                   4s / 6s / 8s / 10s   x1..x4  12      'Video · 720p · 8s...'
  veo_3_1_lite                 NONE                 x1..x4  10      'Video · 720p...'
  veo_3_1_fast                 NONE                 x1..x4  20      'Video · 720p...'
  veo_3_1_quality              NONE                 x1..x4  100     'Video · 720p...'
  veo_3_1_lite_lower_priority  PICKER MISS  -- not offered in the model picker at all
```

Full JSON: `scripts/dev/_spike_out/video_model_capability_matrix_20260903_225459.json`.

This reproduces the 2026-08-14 matrix exactly. Three things make it a solid read rather than a missed selector:

1. **The count tabs came back** (`x1`–`x4`) on every model, so the popover was open and parsed correctly. `duration=NONE` is a real absence, not a failed query.
2. **The composer chip agrees independently.** `omni_flash` renders `Video · 720p · 8s...`; every Veo model renders `Video · 720p...` with no duration segment. Two unrelated DOM signals, same answer.
3. **`veo-lite-lp` isn't even selectable** — the model picker doesn't offer it. The popover tab list read `[Imagem, Video, Frames, Elementos, 9:16, 16:9, x1, x2, x3, x4]`. So the PR's claim for that model has nothing behind it either way.

### But I think you probably did see duration controls — on a different frontend

This is the part worth your attention. Google is migrating Flow onto **`flow.google.com`**, and that frontend is a *different app* from the `labs.google` one gflow-cli drives (see #639). While running this, my other account (`ffroliva`) was served the migrated frontend on **7 out of 7** consecutive loads and could not be measured at all — `FlowHostMigratedError` every time.

So the most likely reading of your PR: **your observation is real, and it's about the migrated frontend.** That UI may well expose 4s/6s/8s on every Veo model. gflow-cli cannot drive it today — none of the controls it targets exist there — so relaxing `supports_duration()` wouldn't unlock the capability; on the old frontend it would just replace a clear pre-flight error with a selector hunt that finds nothing, and on the new one the run never gets that far.

That makes this **not a code fix**, which is why the council came back RED beyond the mechanical items.

### If you want to take this further

The genuinely valuable contribution here would be evidence about the migrated frontend, not a change to the capability table:

- Which host does your Flow session serve — `labs.google/fx/tools/flow/...` or `flow.google.com/...`? (The URL bar answers it.)
- If it's `flow.google.com`: a screenshot of the settings popover with a Veo 3.1 model selected, showing the duration row. That would be the first capture of that frontend's capability matrix in this repo, and it's genuinely useful — it feeds the migration work, not this validator.
- If it's `labs.google` and you still see duration tabs on Veo, then there's a cohort difference worth chasing, and re-running `scripts/dev/capture_video_model_capability_matrix.py --profile <you> --project <id>` on your account would prove it. Paste the matrix and I'll re-open this on the spot.

Either way, thank you for the report — it's a real signal about the migration even though the patch itself doesn't land. Separately, the earlier review comment lists the mechanical items (format gate, `test_cli_models.py:36`, the i2v `duration=10` escape, the chain pre-flight narrowing); those are genuine bugs in the current diff regardless of how the capability question resolves.

Also for the record, on the test claim: a full run of `tests/cli tests/mcp tests/api` gives 8 failures on this head vs 7 on `develop` — the 7 are pre-existing order-dependent pollution, so exactly **one** failure is from this PR, and it's `test_cli_models.py:36`. Your "204 relevant tests pass" was accurate for the files you edited.

stgmt:
> But I think you probably did see duration controls — on a different frontend

image

stgmt:
## Positive capability matrix from the screenshot profile

Ran the credit-free capability collector on the authenticated local Chrome profile used for the screenshot.

- profile: `presentation-reels-google`
- project: `d2e1c023-de75-4196-a9c4-4be3fba5bc54`
- editor URL: `https://labs.google/fx/ru/tools/flow/project/d2e1c023-de75-4196-a9c4-4be3fba5bc54`
- resolved Flow URL: `https://labs.google/fx/ru/tools/flow?hl=en`
- page locale: `ru`
- run: `2026-09-04 04:14:14Z`
- spend: `$0` — navigation, model selection, and popover reads only; Generate was never clicked

```json
{
  "omni_flash": {
    "selected": true,
    "duration_tabs": ["4s", "6s", "8s", "10s"],
    "count_tabs": ["x1", "x2", "x3", "x4"]
  },
  "veo_3_1_lite": {
    "selected": true,
    "duration_tabs": ["4s", "6s", "8s"],
    "count_tabs": ["x1", "x2", "x3", "x4"]
  },
  "veo_3_1_fast": {
    "selected": true,
    "duration_tabs": ["4s", "6s", "8s"],
    "count_tabs": ["x1", "x2", "x3", "x4"]
  },
  "veo_3_1_quality": {
    "selected": true,
    "duration_tabs": ["4s", "6s", "8s"],
    "count_tabs": ["x1", "x2", "x3", "x4"]
  },
  "veo_3_1_lite_lower_priority": {
    "selected": false,
    "picker": "miss"
  }
}
```

Full local receipt:
`scripts/dev/_spike_out/video_model_capability_matrix_20260904_041428.json`

SHA256:
`ac394db744ddcba301764057f9761325b1dfd345b5474f2a5cd83bbbeb299529`

This confirms the screenshot profile has the Veo duration controls on `labs.google`. The earlier `denon82` capture showed no Veo duration controls, so the capability is account/cohort-dependent; the implementation must not claim this matrix is universal without a session-level capability check.

stgmt:
## Full machine-generated matrix JSON

Credit-free navigation and popover capture; no Generate action.

```json
{
  "project": "d2e1c023-de75-4196-a9c4-4be3fba5bc54",
  "profile": "presentation-reels-google",
  "note": "credit-free: navigation + popover reads only, never submitted",
  "baseline": {
    "menu_present": true,
    "tabs": [
      {
        "label": "imageImage",
        "id": "radix-:r1b:-trigger-IMAGE",
        "selected": false
      },
      {
        "label": "videocamVideo",
        "id": "radix-:r1b:-trigger-VIDEO",
        "selected": true
      },
      {
        "label": "crop_freeFrames",
        "id": "radix-:r1e:-trigger-VIDEO_FRAMES",
        "selected": true
      },
      {
        "label": "chrome_extensionIngredients",
        "id": "radix-:r1e:-trigger-VIDEO_REFERENCES",
        "selected": false
      },
      {
        "label": "crop_9_169:16",
        "id": "radix-:r1h:-trigger-PORTRAIT",
        "selected": true
      },
      {
        "label": "crop_16_916:9",
        "id": "radix-:r1h:-trigger-LANDSCAPE",
        "selected": false
      },
      {
        "label": "Veo 3.1 - Litearrow_drop_down",
        "id": "radix-:r1k:",
        "selected": false
      },
      {
        "label": "4s",
        "id": "radix-:r1m:-trigger-4",
        "selected": true
      },
      {
        "label": "6s",
        "id": "radix-:r1m:-trigger-6",
        "selected": false
      },
      {
        "label": "8s",
        "id": "radix-:r1m:-trigger-8",
        "selected": false
      },
      {
        "label": "x1",
        "id": "radix-:r1q:-trigger-1",
        "selected": true
      },
      {
        "label": "x2",
        "id": "radix-:r1q:-trigger-2",
        "selected": false
      },
      {
        "label": "x3",
        "id": "radix-:r1q:-trigger-3",
        "selected": false
      },
      {
        "label": "x4",
        "id": "radix-:r1q:-trigger-4",
        "selected": false
      }
    ],
    "credits_text": "5",
    "composer_chip": "Video · 720p · 4scrop_9_16x1",
    "menu_text": "imageImagevideocamVideocrop_freeFrameschrome_extensionIngredientscrop_9_169:16crop_16_916:9Veo 3.1 - Litearrow_drop_down4s6s8sx1x2x3x4Generating will use 5 credits",
    "page_lang": "en",
    "ingredient_reject": false
  },
  "model_picker_trigger": "button[aria-haspopup='menu']:has(i.google-symbols:text-is('arrow_drop_down'))",
  "models": [
    {
      "model": "omni_flash",
      "selected": true,
      "duration_tabs": [
        "4s",
        "6s",
        "8s",
        "10s"
      ],
      "has_duration_control": true,
      "count_tabs": [
        "x1",
        "x2",
        "x3",
        "x4"
      ],
      "aspect_tabs": [
        "crop_9_169:16",
        "crop_16_916:9"
      ],
      "credits_text": null,
      "composer_chip": null,
      "ingredient_rejected": false,
      "page_lang": "ru",
      "menu_text": "imageИзображениеvideocamВидеоcrop_freeКадрыchrome_extensionОбразцыcrop_9_169:16crop_16_916:9Omni 1.1 Flasharrow_drop_down360p720p4s6s8s10sx1x2x3x4Стоимость генерации в бонусах: 7"
    },
    {
      "model": "veo_3_1_lite",
      "selected": true,
      "duration_tabs": [
        "4s",
        "6s",
        "8s"
      ],
      "has_duration_control": true,
      "count_tabs": [
        "x1",
        "x2",
        "x3",
        "x4"
      ],
      "aspect_tabs": [
        "crop_9_169:16",
        "crop_16_916:9"
      ],
      "credits_text": null,
      "composer_chip": null,
      "ingredient_rejected": false,
      "page_lang": "ru",
      "menu_text": "imageИзображениеvideocamВидеоcrop_freeКадрыchrome_extensionОбразцыcrop_9_169:16crop_16_916:9Veo 3.1 - Litearrow_drop_down4s6s8sx1x2x3x4Стоимость генерации в бонусах: 5"
    },
    {
      "model": "veo_3_1_fast",
      "selected": true,
      "duration_tabs": [
        "4s",
        "6s",
        "8s"
      ],
      "has_duration_control": true,
      "count_tabs": [
        "x1",
        "x2",
        "x3",
        "x4"
      ],
      "aspect_tabs": [
        "crop_9_169:16",
        "crop_16_916:9"
      ],
      "credits_text": null,
      "composer_chip": null,
      "ingredient_rejected": false,
      "page_lang": "ru",
      "menu_text": "imageИзображениеvideocamВидеоcrop_freeКадрыchrome_extensionОбразцыcrop_9_169:16crop_16_916:9Veo 3.1 - Fastarrow_drop_down4s6s8sx1x2x3x4Стоимость генерации в бонусах: 10"
    },
    {
      "model": "veo_3_1_quality",
      "selected": true,
      "duration_tabs": [
        "4s",
        "6s",
        "8s"
      ],
      "has_duration_control": true,
      "count_tabs": [
        "x1",
        "x2",
        "x3",
        "x4"
      ],
      "aspect_tabs": [
        "crop_9_169:16",
        "crop_16_916:9"
      ],
      "credits_text": null,
      "composer_chip": null,
      "ingredient_rejected": false,
      "page_lang": "ru",
      "menu_text": "imageИзображениеvideocamВидеоcrop_freeКадрыchrome_extensionОбразцыcrop_9_169:16crop_16_916:9Veo 3.1 - Qualityarrow_drop_down4s6s8sx1x2x3x4Стоимость генерации в бонусах: 100"
    },
    {
      "model": "veo_3_1_lite_lower_priority",
      "selected": false,
      "menu_text": "imageИзображениеvideocamВидеоcrop_freeКадрыchrome_extensionОбразцыcrop_9_169:16crop_16_916:9Veo 3.1 - Qualityarrow_drop_down4s6s8sx1x2x3x4Стоимость генерации в бонусах: 100",
      "page_lang": "ru",
      "tabs": [
        "imageИзображение",
        "videocamВидео",
        "crop_freeКадры",
        "chrome_extensionОбразцы",
        "crop_9_169:16",
        "crop_16_916:9",
        "Veo 3.1 - Qualityarrow_drop_down",
        "4s",
        "6s",
        "8s",
        "x1",
        "x2",
        "x3",
        "x4"
      ],
      "model_trigger_count": 1,
      "menuitems_seen": []
    }
  ]
}
```

ffroliva:

⚠️ Found **4** MUST-FIX issue(s). Please review and resolve them before merge.

<details>
<summary>View Sandboxed Review Output</summary>

Building Docker sandbox image...
Creating network triage-net-650...
a8458414d885f96791b57d645ebd9051693fff736e787be9a01c85ee12a06cd6
Hardening network isolation for subnet 172.20.0.0/16 via iptables...
Launching sandboxed review for PR 650...
All 11 dimension reports are in, and I independently re-verified the three most-corroborated findings via `git show` against `REVIEWED_SHA` — all confirmed. Synthesizing the final verdict now.

---

# PR #650 — Council Review Verdict

**Repo:** `gflow-cli` · **Title:** feat(video): support duration selection (4s, 6s, 8s) for Veo 3.1 models
**REVIEWED_SHA:** `cc650faddaea00a9648073fe9ec5c0904bcf3c22` (branch `fix/veo-3-1-duration-support`, external-contribution fork PR). No SHA divergence detected — head unchanged since dispatch.

## Consensus: 🔴 RED

|---|---|---|
| D0 — Mechanical CI gate | ⚠️ UNVERIFIED | `uv` blocked in this sandbox; `gh pr checks` shows only the fork-routing bot ran — Lint/Format/pytest/pyright never triggered |
| D1 — Correctness | 🔴 RED | Reverses a previously "confirmed live" finding with zero live evidence attached |
| D2 — Code quality | 🟡 YELLOW | Dead code left over from the refactor (see must-fix #2) |
| D3 — Security | 🟢 GREEN | Pure validation-logic change; no injection/env/path/secret concerns |
| D4 — Tests | 🟡 YELLOW | Test-coverage gap on `test_movie_duration_with_veo_lite_is_accepted`; no live-verify evidence |
| D5 — Memory hygiene | 🟢 GREEN | No contradicting Claude-memory entry; PLAN.md edit self-consistent |
| D6 — UI/live-verify | 🟡 YELLOW | Behavior-reversal claim unsubstantiated by any capture/screenshot/structlog trace |
| D8 — CLI UX | 🟢 GREEN | Help text and error messages consistent, accurate, actionable |
| D9 — Docs drift | 🟡 YELLOW | Stale `VideoModel` docstring still says Veo models "render none at all" |
| D13 — Dev scripts | 🟢 GREEN | Capture-script selector change is safe; classifier logic sound |
| D14 — Over-engineering | 🟡 YELLOW | Same dead-code finding as D2 |
| D15 — Surface parity | 🟢 GREEN | CLI↔MCP↔chain↔movie-manifest all converge on one shared validator |

## Must-fix (4)

1. **Core claim reverses a prior confirmed-live finding with no new live evidence.** The PR asserts Flow's UI now renders 4s/6s/8s duration controls for all Veo 3.1 models, reversing the "confirmed live on two accounts and two locales (2026-08-14)" finding behind issues #451/#288/#630 — but attaches **no** live-verification artifact (no structlog trace, screenshot, or captured-matrix output). The one file that could supply such evidence, `scripts/dev/capture_video_model_capability_matrix.py`, only widens a DOM selector; its new test (`tests/scripts/test_capture_video_model_capability_matrix.py`) exercises the classifier against hand-authored synthetic labels, not real Flow output. `docs/PROJECT_STATUS.md` and `KNOWN_ISSUES.md` state the reversal as settled fact, and #288 is downgraded from "Open" to "Cohort-dependent" on this basis alone. Per AGENTS.md's Working Discipline: *"If a claim can't be verified in the current environment, it's LIKELY — not CONFIRMED... ship diagnostics rather than a blind fix."* — Flagged by D1, D4, D6, D9.

2. **Dead/unreachable code in `src/gflow_cli/mcp/tools.py:1071–1076`.** `validate_duration_for_model(effective, duration)` (line 1065) already raises+returns for any duration>cap (10 on a non-omni model), making the trailing `if duration in (10, "10") and effective is not VideoModel.OMNI_FLASH:` block unreachable for every input — traced and independently re-verified against `REVIEWED_SHA` by three separate sub-agents plus my own spot-check. The adjacent comment (`# Only omni-flash renders a duration control (#451/#288).`) is also now factually stale. Delete lines 1071–1076 (~6 net lines) and refresh the comment. — Flagged by D1, D2, D14, D15.

3. **Stale `VideoModel` class docstring**, `src/gflow_cli/api/video.py:47–56` — still reads *"They render **none at all** — only `OMNI_FLASH` shows the 4s/6s/8s/10s row"*, directly contradicted two lines below by this same PR's `max_duration_for()`/`validate_duration_for_model()`. Verified: the PR touched this exact docstring block (removing a dangling `supports_duration` cross-ref) but left the substantive false claim untouched. — Flagged by D9.

4. **Weak test coverage in `tests/cli/test_movie_manifest.py::test_movie_duration_with_veo_lite_is_accepted`.** The `assert` (8-space indent) sits outside the inner `for dur in (4, 6, 8)` loop (12-space body) but inside the outer `for model` loop, so it only ever checks the last-written duration (8); durations 4 and 6 are written to TOML but never parsed or asserted — despite the docstring claiming "Veo 3.1 models support 4s, 6s, and 8s durations." **Verified low real-world severity**: `validate_duration_for_model` has no per-value branching for 4 vs 6 vs 8, so no actual bug is masked — but the test still misrepresents its own coverage. Fix: indent the `assert` to 12 spaces. — Flagged by D1, D4.

## Nice-to-have (7)

1. `mcp/tools.py`'s `duration: int | None` signature doesn't match the dead code's implicit `str` handling (`"10"`) — resolves itself once must-fix #2 is deleted. (D2)
2. `api/video.py::_validate_duration` duplicates the choices-membership check also done in `validate_duration_for_model` — intentional (it's the only guard on the model-less t2v/r2v path) but undocumented; a one-line comment would save the next reader a trace. (D2)
3. `tests/mcp/test_server.py::test_duration_8_on_veo_lite_is_not_rejected` only asserts `"duration" not in title.lower()`, not that `duration=8` actually reached the request object — weaker than the CLI-layer equivalent. (D4)
4. `tests/e2e/test_video_r2v_uuid_name_e2e.py` comment lost its `#537` issue-tracking breadcrumb when reworded. (D6)
5. CHANGELOG entry is filed under `### Fixed`; arguably a `### Changed`/new-capability entry given the practical effect is unlocking more values, not just fixing a bug. (D9)
6. `KNOWN_ISSUES.md` #288 status downgrade ("Open" → "Cohort-dependent") cites no live-verification artifact specific to #288 itself. (D9)
7. `scripts/dev/capture_video_model_capability_matrix.py`'s broadened selector (`button, [role='tab'], [role='button'], [role='option'], [role='menuitem']`) could sweep in incidental popover buttons (e.g. Cancel/Close) into the `other` bucket — low risk, `_classify` still won't misfile them as duration/count/aspect. (D13)

## False positives (filtered out)

None — all flagged claims were independently verified against `REVIEWED_SHA` (three separate re-verifications on the dead-code finding, one dedicated re-verification each on the test-indentation and stale-docstring findings, plus my own synthesis-stage spot-check of all three).

## Confirmed-good (high-confidence positives)

- **Security**: no injection vectors, no new env-var reads, no path/secret/signature-bypass concerns; `int(duration)` conversion path safely handles malformed/huge/negative input without crashing. (D3)
- **Real centralization, not partial**: `validate_duration_for_model()` is now the single source of truth across the DTO, CLI, chain, movie-manifest, and MCP surfaces — the old duplicated `supports_duration()` method and its per-call-site error strings are fully removed with zero stale references left in `src/`. (D2, D14, D15)
- **CLI UX**: help text for `--duration` is identical and accurate across `t2v`/`i2v`/`r2v`; error messages name the model, its cap, and the valid alternative; `gflow models --json` `max_duration` correctly updated (8, was 0). (D8)
- **Docs mirror consistency**: `docs/*.md` ↔ `website/docs/*.md` changes are byte-identical (KNOWN_ISSUES differs only by the expected PII-redaction substitution). (D9)
- **Surface parity**: CLI, MCP, chain, and movie-manifest all converge on the one validator; CLI/MCP error-message content matches; `tests/mcp/test_cli_parity.py` correctly untouched since no new leaf command was added. (D15)
- **Memory hygiene**: no Claude-memory entry contradicts this PR (the prior "no duration control" finding was only ever captured in an in-repo spike doc, never in memory); the touched `docs/superpowers/plans/2026-09-01-video-extend/PLAN.md` line reads coherently in context. (D5)
- **Dev script**: `_classify()`'s duration/count/aspect bucketing is mutually exclusive and well-tested; no Windows-specific risk introduced. (D13)

## Memory actions (D5) — reported only, not applied (autonomous mode)

- **ADD**: a durable-pattern entry noting Flow's model-conditional UI capability matrices are time-stamped observations, not permanent facts — the 2026-08-14 "Veo 3.1 has no duration control" spike was reversed by this PR (2026-09-04, itself unverified); future agents should re-verify before citing either capture as current.
- **UPDATE**: the memory entry still characterizing #451 as "Flow dropped the duration-tab UI" is now doubly stale and should be corrected or retired.
- **Housekeeping note** (unrelated to this PR): `project_hardening_roadmap_2026_07_09.md` exists on disk with zero references in `MEMORY.md` — pre-existing index drift, flag for separate cleanup.

## D0 — Mechanical CI gate

**UNVERIFIED.** `uv` execution is blocked by sandbox policy in this session (no approval available), so `ruff`/`ruff format`/`pyright`/`pytest` could not be run locally against `REVIEWED_SHA`. Fallback per protocol: `gh pr checks 650` shows only `Label and route external PRs` (a routing bot check) — the actual Lint/Format/type/test CI job has **not triggered at all**, consistent with GitHub's default policy of gating workflow runs on fork PRs behind maintainer approval. This must be confirmed green (after a maintainer approves the workflow run) before merge, independent of the LLM council's findings.

**SonarCloud**: no analysis run visible for this fork PR — per autonomous-mode protocol this is treated as an informational note, not a block, but should be confirmed by a maintainer before merge per the repo's stated "PR is not done until SonarCloud is green" rule.

---

`SUMMARY_VERDICT: RED | MUST_FIX_COUNT: 4 | PR_URL: https://github.com/ffroliva/gflow-cli/pull/650`
Cleaning up network rules and Docker network...

</details>

ffroliva:
## Correction: your capture stands. I withdraw the rejection.

You were right, and I was wrong about *why* you were right. Two things to clear up before the code.

### 1. The premise holds — it's cohort variance, on `labs.google`

My "the premise doesn't hold" comment guessed you were on the migrated `flow.google.com` frontend. Your matrix says otherwise: `labs.google/fx/ru/tools/flow`, the same frontend gflow drives. So this is **not** a frontend difference, it's an **account/cohort** difference — which is the conclusion you reached yourself, ahead of me.

Your capture passes the read-validity check this repo wrote for exactly this purpose: count tabs came back on every model, so `4s/6s/8s` is a real read, not a lucky selector. The composer chip agrees independently (`Video · 720p · 4s…` with a Veo model selected). And the credit prices differ from mine — `5/10/100` against my `10/20/100`, with your `omni` exposing `360p/720p` and billing 7 — which is a second, unrelated signal that these are genuinely different buckets rather than one of us misreading a popover.

Both matrices are now recorded as valid for their own cohorts. Yours reopened a finding this repo had marked settled after two accounts agreed; the lesson taken is that two agreeing accounts are evidence of a shared cohort, not of universality.

### 2. I checked whether your own patch had biased the measurement

Being explicit, because it would have been unfair not to say it out loud: this PR also widens the collector's scrape from `[role='tab']` to `button, [role='tab'], [role='button'], [role='option'], [role='menuitem']` (`capture_video_model_capability_matrix.py:109`). My negative run used the old selector, yours the new one — different instruments, so I had to rule that out before accepting the result.

**Ruled out.** Your JSON pins the elements as `radix-:r1m:-trigger-4/6/8` — Radix tab triggers, which carry `role="tab"` and which the base selector demonstrably reads (it found omni's `4s/6s/8s/10s` row on my account). Two fields the diff never touches, `menu_text` and `composer_chip`, corroborate independently. The widening is a **fidelity fix**, not a thumb on the scale — the transport's own cascade already probed all five roles, so the recon was narrower than the code it models. Good catch on its own merits.

### 3. About the bot comment above

The `PR-Triage Autopilot` verdict posted at 02:09 leads with *"reverses a confirmed-live finding with no live evidence attached"* — 50 minutes after you posted a machine-generated matrix with a SHA-256 and a screenshot in this thread. It never read the thread: the council's gather step pulled the diff, the files and the checks, but not the comments. Its D5 dimension also reported "no memory entry contradicts this PR" from an empty directory it could not read.

Both are fixed in #654, along with the raw Docker/iptables logging that block should never have contained. Please disregard its must-fix #1; #2, #3 and #4 are real and are folded into the list below.

Its `D0 UNVERIFIED` was correct, incidentally — CI was queued awaiting fork approval and only ran at 09:00.

---

## Where the code actually stands

`cc650fa` is a good rework and I want to be clear about that. It deleted `supports_duration()` outright rather than papering over it, collapsed five hand-rolled checks into one `validate_duration_for_model()` consumed at six edges, and comes in at **net −31 lines** in `src/`. Prior blockers **2, 3 and 4 are verified fixed**:

| Prior blocker | Status | Evidence |
|---|---|---|
| Format gate | FIXED | `ruff format --check src tests` clean at `cc650fa` |
| `test_cli_models.py:36` | FIXED | `pytest tests/cli/test_cli_models.py` → 21 passed |
| i2v `duration=10` escape | FIXED | `GenerateVideoRequest(mode=I2V, start_image='a.png', duration=10)` → rejected |
| Chain pre-flight | FIXED | `reject_unusable_links(duration=5)` → rejected before link 0 |

An 11-dimension council ran against `cc650fa`. Twelve items remain. Ten are yours if you want them — all offline, no Flow account needed. Two are mine.

### The CI red is a test bug, not a source bug

`tests/cli/test_cli_video_chain.py:565` asserts `"only available"`. Rich splits that phrase across a line break at console width 80, which is CI's non-TTY default; local Windows Rich reports 79, which is why it passes for you and fails there. The message itself is correct and complete.

Fix: assert a wrap-safe token (`"omni_flash"`), or normalise with `" ".join(result.output.split())`. Note `"caps at 8s"` in `test_cli_video.py:1634,1646` and `test_movie_manifest.py:547` is fragile the same way — it survives only because Click prints `UsageError` unwrapped.

### Yours (10)

**Tests**
1. `tests/cli/test_cli_video_chain.py:565` — the wrap-fragile assertion above. Sole CI blocker; everything else is `3702 passed`.
2. `tests/cli/test_cli_video.py:1674` — `--dry-run` was dropped (t2v never had that flag), so the test now runs unpatched and boots the real transport: exit 36, real profile/browser work, an incident-DB write inside a unit test. Patch `_run_t2v`/`_resolve_profile` the way the new accept test at :1652 does.
3. `tests/cli/test_movie_manifest.py:552` — the `assert` sits outside the inner `for dur` loop, and `_write_toml` reuses one path, so only `duration = 8` is ever parsed. 4s and 6s are written and discarded, despite the docstring claiming all three.

**Docs that now contradict the code**

4. `docs/USAGE.md:844-851` + mirror — "duration is not supported in a chain". Chain accepts 4/6/8 on this head.
5. `docs/PROJECT_STATUS.md:105,112-113` — "no model a chain can use has a duration control" contradicts line `:114`, which this PR added to the same paragraph.
6. `KNOWN_ISSUES.md:292-293` + mirror — "the fail-fast is the correct behavior", after this PR removed that fail-fast for 4/6/8.
7. `docs/USAGE.md:503-507` and `docs/PROJECT_STATUS.md:242-244` — "verified live in Flow UI" names no account, locale or host. Given what we now know, these have to say cohort-dependent. This is the mirror image of my mistake, so I'd rather it be explicit than quietly flipped.

**Strings and dead code**

8. `cli_video.py:1586-1589` — `video chain --help` still says a per-link duration is rejected.
9. `api/video.py:47-53` and `mcp/tools.py:1052` — comments still assert Veo models "render none at all".
10. `mcp/tools.py:1071-1077` — unreachable: `validate_duration_for_model` raises first for every input that can reach it. −7 lines.

Also worth a look, not blocking: `capture_video_model_capability_matrix.py:108` keeps `scope = menu || document.body` while the selector widened, so on a popover-open failure it now sweeps every button on the page and a stray `4s` would set `has_duration_control: true`. Given this script is what settles capability arguments, it shouldn't be able to produce a false positive. Its docstring rewrite also drops the dated `#451/#288` root-cause record — worth restoring, and the "duration controls are interactive buttons in some cohorts" line is contradicted by your own Radix ids.

### Mine (2)

11. `ui_automation_video.py:1786-1798` — the duration cascade's candidates 5–10 aren't scoped to `[role='menu']` (the aspect selectors at `:236-239` are), and nothing reads back after the click. A clean miss is safe — `configure_video_settings` runs before `_submit_and_poll`, so it's exit 23 at $0 — but a false-positive match on a no-control cohort would click something else, log `duration_set`, and submit. This is our highest-risk surface and needs verifying on both cohorts, which only I can do.
12. `ui_automation_video.py:1806` + `errors.py:627-635` — the exit-23 remediation tells the user Google changed the frontend and to file a bug. On a no-control cohort that's a wrong diagnosis that would generate junk issues.

I'll take those two plus the `KNOWN_ISSUES` #288 record in a follow-up, and I'll live-verify the cascade change on my cohort.

---

Nothing here is urgent — no release is waiting on it. Take the ten if you want them; say so if you'd rather hand any back and I'll pick them up.

Thank you for pushing back with a measurement instead of an argument. The cohort finding is more valuable than the patch, and it wouldn't exist if you'd let the first review stand.

stgmt:
Thanks for the detailed review and confirming the cohort-variance finding.

All 10 requested items are addressed and verified on head `96d097d`:

### Tests
1. `tests/cli/test_cli_video_chain.py:565`: replaced wrap-fragile `"only available"` assertion with the continuous token `"omni_flash"`.
2. `tests/cli/test_cli_video.py:1674`: patched `_run_t2v`, `_resolve_profile`, and `_make_provider_dir` on `test_duration_on_t2v_without_model_is_left_alone` matching `test_veo_duration_is_accepted_by_cli_preflight` to prevent real transport execution and incident writes.
3. `tests/cli/test_movie_manifest.py:552`: moved manifest parsing and `duration == dur` assertion inside the inner `for dur in (4, 6, 8)` loop so all 6 model × duration combinations are verified.

### Docs & website mirrors
4. `docs/USAGE.md` & `website/docs/USAGE.md`: updated chain manifest section to document accepted 4s/6s/8s per-link durations for Veo, early pre-flight rejection of 10s / invalid values, and chain exclusion of omni-flash.
5. `docs/PROJECT_STATUS.md`: updated chain/movie section to describe the unified `validate_duration_for_model` pre-flight check across single-clip, chain, and movie surfaces.
6. `KNOWN_ISSUES.md` & `website/docs/KNOWN_ISSUES.md`: updated #288 description to reflect that explicit duration fails fast on cohorts lacking the control, while supporting cohorts expose 4s/6s/8s.
7. `docs/USAGE.md` & `docs/PROJECT_STATUS.md`: scoped duration support explicitly as cohort-dependent with references to both the historical negative capture (`denon82`) and the positive capture (`presentation-reels-google`, `labs.google`, 2026-09-04, PR #650). Regenerated website mirrors via `scripts/ci/generate_website_docs.py`.

### Strings & dead code
8. `src/gflow_cli/cli_video.py`: updated `gflow video chain --help` text and expanded `test_chain_help_states_cost_and_scene_followup` to verify 4/6/8 duration guidance.
9. `src/gflow_cli/api/video.py` & `src/gflow_cli/mcp/tools.py`: updated `VideoModel` docstring and MCP comments to describe the cohort reality and centralized validation.
10. `src/gflow_cli/mcp/tools.py`: removed the unreachable `-7` line secondary check in `gflow_generate_video`.

### Maintainer items
- Items 11–12 (`ui_automation_video.py` scoping/read-back and `errors.py` remediation) are left for your follow-up as discussed.

### Local verification
- `pytest tests/cli/test_cli_video_chain.py tests/cli/test_cli_video.py tests/cli/test_movie_manifest.py tests/mcp/test_server.py` → 187 passed
- `scripts/ci/check_doc_links.py` → All links resolved across 29 files
- `scripts/ci/check_website_docs_pii.py` → No private identifiers across 25 files
- `scripts/ci/generate_website_docs.py --check` → website/docs mirror in sync (20 files)
- `ruff check src tests` & `ruff format --check src tests` → clean
- `pyright src` → 0 errors, 0 warnings
- `python -m pytest -q --cov=gflow_cli --cov-report=xml --junitxml=pytest-report.xml` → 3714 passed, 6 skipped, 94 deselected (floor 2000 satisfied)

All commits are rebased onto current `develop` and signed off (`git log FETCH_HEAD..HEAD` has DCO sign-offs on all 3 commits). Ready for a re-run of council review.

ffroliva:
Small scope correction on the list I posted — my fault for double-assigning.

**Please drop `KNOWN_ISSUES.md` (and its `website/docs/` mirror) from this PR.** I said I'd take the #288 record, then still listed `KNOWN_ISSUES.md:292-293` as item 6 for you. That was item 6 on your list; it is now #655, which rewrites the whole `#288` row. If you leave your edit in, the two conflict on the same lines.

Everything else on the list stands. #655 also settles a detail worth knowing while you work: #288, #451 and #634 are all closed on GitHub, so any doc text you touch that describes them as open behaviour is stale in the same way.

Nothing else changes.

stgmt:
Dropped `KNOWN_ISSUES.md` (and its `website/docs/` mirror) from this PR branch and rebased onto latest `develop` containing #655. Head is now `cc220d5`. All documentation gates (`check_doc_links`, `check_website_docs_pii`, `generate_website_docs --check`), linters (`ruff`, `pyright`), and tests pass cleanly.

ffroliva:
## Council review — `cc220d5` · consensus 🔴 RED

**All ten items from the last round are fixed.** I verified each one mechanically before the council ran, and the fixes are the right ones rather than the minimum ones — the wrap-fragile assertion now uses a single wrap-safe token, the t2v test patches `_resolve_profile` / `_make_provider_dir` / `_run_t2v` instead of booting a browser, and the movie-manifest assert moved inside its loop. CI is green on all 14 checks. The revert control is strong: your nine changed test files produce **28 failures** against the merge base, so the suite genuinely pins the new behaviour.

RED is for what the rebase exposed, and one hole that is mine.

---

### Mine, not yours — and it blocks the merge

**`KNOWN_ISSUES.md` on `develop` is falsified by this PR.** I asked you to drop that file, then #655 landed text describing the *pre-#650* gate as present-tense fact. Because the two changes touch disjoint files, git merged them cleanly while the prose collided. Three reviewers found it independently:

- `:283` — "`--model omni-flash`, the only model gflow currently allows a duration on"
- `:285-287` — "gflow refuses `--duration` on every named Veo 3.1 model at the CLI edge — `click.UsageError`, exit 2"
- `:308-309` — "the flag is unavailable on every cohort today"

All three are false at `cc220d5` (`_reject_duration_without_control('veo-fast','6')` → accept). `:276-277` also says relaxing the gate is "tracked in #650", which becomes "shipped in". **I will land the correction alongside this PR** — you do not need to touch it, and it should not go in before this merges.

---

### Yours

**1. The CHANGELOG entry is under a shipped release.** `[Unreleased]` (line 8) is empty; the bullet landed in `## [0.66.2] — 2026-09-03`'s `### Fixed`. A rebase artifact — move it up.

**2. MCP accepts `duration=99`.** The range check sits inside `if effective is not None:`, so omitting the model skips it:

```
gflow_generate_video(prompt="x", mode="t2v", duration=99)  -> ACCEPT (queued)
gflow video t2v --duration 99                              -> Invalid value: not one of '4','6','8','10'
GenerateVideoRequest(..., duration=99)                     -> REJECT
```

`tools.py:1053` claims it "rejects an invalid duration before queuing the job so the agent gets a clean 400 instead of a ValueError inside the worker" — true only when a model is named. Hoisting the `VIDEO_DURATION_CHOICES` membership test out of that branch is a one-line fix. `tests/mcp/test_cli_parity.py` cannot see this by construction: it is command-level, and no leaf changed.

**3. The MCP `duration` schema lost its constraint entirely.** `{"anyOf":[{"type":"integer"},{"type":"null"}]}` with no `enum`, against `click.Choice(["4","6","8","10"])` on the CLI. The docstring (`tools.py:988`) now says only "Optional clip length in seconds", and the tool description (`:940`) just "duration (seconds)". The pre-PR docstring *did* state the restriction and nothing replaced it — so an agent has no signal that 10 is omni-only while `--help` spells it out.

**4. Two new unconditional claims — the mirror image of my own mistake.** While fixing the old absolutes you introduced fresh ones. `docs/MCP.md:87` states "4/6/8 for Veo 3.1" with no cohort wording and drops the `#451/#288/#630` refs; `docs/MOVIE.md:53-54` says "Veo 3.1 models support 4, 6, and 8" (`grep -c -i cohort docs/MOVIE.md` → 0). `USAGE.md:504-510` got this right — these two need the same caveat.

**5. `docs/PROJECT_STATUS.md:322-323`** still says `gflow models` "stops advertising a duration users cannot set". Now false: `cli_models.py:69` emits `max_duration_for(m)` = 8 for every Veo model. You rewrote `:316-322` of that paragraph and left the next two lines.

**6. `src/gflow_cli/api/video.py:57-58`** declines to make a live claim for `veo_3_1_lite_lower_priority` ("picker miss… no positive live-UI claim is made for it") while `max_duration_for` returns 8 for it, so the CLI accepts `--duration 8 --model veo-lite-lp`. The code asserts what its own docstring refuses to. Pick one.

**7. `tests/test_chain_manifest.py:46-50`** still states `run_chain` "rejects **any** link carrying a `duration`" — false now, it rejects only 10. Missed in the sweep.

### Test gaps

**8. Two chain-help assertions are vacuous.** `test_cli_video_chain.py:534`'s `assert "duration is rejected" not in out` passes against the *pre-fix* help too — the old text is `"duration" is rejected`, with quote characters. And `:531,533` assert bare digits: on the base tree `'4'` and `'6'` already matched, so the revert is caught by two digits by accident. Assert a phrase like `"4, 6, 8"`.

**9. The MCP accept test never reaches the queue.** `test_server.py:853` short-circuits under the autouse `_isolate_settings` fixture — reproduced with an empty `GFLOW_CLI_HOME`: `{"title":"No Profile Found","status":400}`. It asserts only that the error *title* lacks "duration", so nothing proves `duration=8` reaches the queued payload the way `test_cli_video.py:1662` does for the CLI.

**10. Accept-path coverage is uneven.** i2v and r2v have no accept test at all, though the reject test covers all three; chain and MCP exercise **8 only**. DTO, CLI t2v and movie now cover 4/6/8 properly.

### Worth knowing, not blocking

`tests/scripts/test_capture_video_model_capability_matrix.py` passes verbatim on the pre-PR tree — it exercises `_classify`, which this PR does not change, while the actual edit (the widened `querySelectorAll`) has zero coverage. Related: that script still uses `scope = menu || document.body`, so with the wider selector a failed popover open can now sweep every button on the page and read as a positive. Given this is the instrument that settles capability arguments, it should not be able to manufacture one — dropping the `document.body` fallback, or recording `menu_present`, would close it.

Also minor: `chain.py:235` recomputes `effective` under a second name, and `_reject_duration_without_control` is now a misnomer — it validates a cap.

---

Items 1 and 4-7 are mechanical. Item 2 is the one I would prioritise: it is a real hole an agent can drive through today, and it is invisible to every automated gate we have.

ffroliva:
Merged. Thank you — and I want to be specific about what you did, because it was better than the patch.

You were right and I was wrong, and you settled it the hard way. When I posted "the premise doesn't hold", the easy responses were to argue or to walk away. You ran the collector on your own account and posted a machine-generated matrix with a SHA-256 instead. That is what changed my mind, and nothing else would have.

You also reached the correct conclusion before I did. Your line — *"the capability is account/cohort-dependent; the implementation must not claim this matrix is universal without a session-level capability check"* — is sharper than the PR you opened, and sharper than my rejection of it. It is now recorded in the repo's own docs and in this project's durable memory, and it changed how we review capability claims generally: agreement across two accounts is evidence of a shared cohort, never of universality. That lesson outlives the diff.

For the record on the code: three review rounds, and you fixed every item raised in each one — including using exactly the wrap-safe token the reviewer suggested, and patching the three seams that stopped a unit test booting a real browser. `supports_duration()` is gone rather than inverted, five hand-rolled checks collapsed into one shared validator, net −31 lines in `src/`.

The remaining items from the last round were mine, not yours, and are already landed or filed so nothing is left hanging on you:

- #658 — the `KNOWN_ISSUES` row I asked you to drop, plus the CHANGELOG placement and the two docs that picked up new unconditional claims. Merged.
- #659 — the MCP `duration=99` escape. Confirmed **pre-existing**, not something you introduced, which is why it did not block this.
- #660 — a $0 probe on the collector, partly because of something your PR surfaced: the widened selector meant a popover that failed to open could scrape the whole page and read a stray `8s` as a duration tab. Worth closing on an instrument this repo uses to settle arguments.

If you want a next one: #657 is the same unscoped-cascade shape on the **count** tabs, where a wrong match doubles spend rather than mis-sizing a clip. It is gated on #660's probe result, so it may turn out to need no code at all — which is the honest reason I have not started it.

Thanks again for pushing back with evidence.

Issue #643, opened 2026-09-03
bug: locale resolution is blind on flow.google.com and DEMOTES a learned locale (measured en + pt-BR)


ffroliva:
## Summary

On Flow's migrated `flow.google.com` origin, locale resolution is **structurally blind** — and worse, it **demotes a correctly-learned locale** on every migrated page load.

Measured 2026-09-03 on two profiles, read-only, zero credits.

## The measurement

| | `ffroliva` (en) | `denon82` (pt-BR) |
|---|---|---|
| **old host** URL | `labs.google/fx/tools/flow` | `labs.google/fx/**pt**/tools/flow` |
| `LOCALE_SEGMENT_RE` matches | false (bare, not redirected) | **true** |
| `locale_segment_from_url` | `null` | **`"pt"`** |
| `html lang` | `en` | `pt` |
| **migrated** URL | `flow.google.com/` | `flow.google.com/?pli=1` |
| `LOCALE_SEGMENT_RE` matches | **false** | **false** |
| `locale_segment_from_url` | `null` | **`null`** |
| `html lang` | `en-GB` | **`pt`** — still correct! |
| `await_url_settled` | `None` after **4018 ms** | `None` after **4017 ms** |

## Root cause

`routes.py:180`:

```python
LOCALE_SEGMENT_RE = re.compile(r"^https://labs\.google/fx/([a-z]{2,3})(?:-[A-Za-z]{2})?/tools/flow")
```

The migrated URL is `/project/<id>` — it drops `/fx/<locale>/tools/flow` entirely, so this can never match. **The locale did not disappear; it moved from the URL into the document** (`html lang=pt` is still correct on the migrated host).

## Two consequences, both measured

**1. It demotes a known-good locale.** `next_locale_state("pt", None)` returned **`"?"`** (PROVISIONAL) — measured, not inferred. Every migrated load discards denon82's learned `pt`. Not a correctness break (bare URLs are documented as "never worse"), but on a fully-migrated account the locale can never be learned again, silently undoing #587.

**2. Four seconds wasted per navigation.** `await_url_settled` burns the full `URL_SETTLE_TIMEOUT_MS` waiting for a shape that can no longer exist. This lands on top of the ~8 s `detect_ui_mode` window and ~24 s crop cascade already spent before a migrated run fails.

## Suggested fix

[...]

PR #652, opened 2026-09-03
fix(651): read `<html lang>` after hydration, not before

### LINKED ISSUE #651: bug(locale): the <html lang> fallback races hydration — reads the 'en' shell before the app sets the real locale
Split out of [#639](https://github.com/ffroliva/gflow-cli/issues/639), where [@maipmacrothorax-75](https://github.com/maipmacrothorax-75) measured it on a **pt-BR** account: gflow resolves `locale=en` while both `flow.google.com/` and `flow.google.com/project/<id>` serve `<html lang="pt">`, `navigator.language` is `pt-BR`, and the whole UI renders in Portuguese.

They offered the mechanism as an explicit guess. **It is now confirmed**, reproduced locally on the *old* host at zero credits — so this is not a migration bug.

## Measurement

`labs.google/fx/tools/flow?hl=en` on a pt account, sampling `document.documentElement.lang` after `goto(wait_until="domcontentloaded")` returns:

| t (ms) | `<html lang>` |
|---|---|
| 805 | `en` |
| 1072 | `en` |
| 1342 | `en` |
| 1860 | `en` |
| **2874** | **`pt`** |
| 4885 | `pt` |
| 7900 | `pt` |

Identical across two consecutive loads. The attribute starts at the server-default `en` shell and flips to the account's real locale at **~2.4–2.9 s**. `client.account_locale_resolved` fires at ~1.6 s — inside the `en` window.

## Why it hid

`_resolve_account_locale` tries the URL first and only falls back to `<html lang>`. An account whose URL carries `/fx/pt/` never reaches the fallback, so it is immune — which is why maintainer testing missed it. The exposed population is exactly the one #643 was written for: accounts where the URL cannot answer (the migrated origin, or a bare non-redirecting URL).

Also worth correcting: v0.66.1's docstring licenses the fallback with *"`<html lang>` agreed with the URL segment wherever both existed"*. That was measured **after** settle. Read early, they disagree — which is the whole defect.

## Impact

- Affected: `gflow-cli` **0.66.1** and **0.66.2**, any account where the URL segment is absent.
- The wrong segment is cached, so `project_editor_url` then builds `/fx/en/tools/flow/...` for a non-`en` account. `locale_segment_from_url`'s own docstring notes Flow does **not** correct a wrong-but-valid segment, so whether this causes a redirect bounce (the #395 class) or merely an English UI is **not yet measured** — worth settling before choosing the fix's strictness.

## Fix shape

Do not read `lang` at a fixed instant. Poll until it stabilises — or until it stops equalling the shell default — bounded, and fall back to `None` (bare URLs) rather than to a guess. The probe above is the regression test; it needs no migrated account and costs nothing.

Refs #639
Refs #643

ffroliva:
Fixes [#651](https://github.com/ffroliva/gflow-cli/issues/651). Reported by [@maipmacrothorax-75](https://github.com/maipmacrothorax-75) on a pt-BR account, with the mechanism offered explicitly as a guess. **The guess was right**, and it reproduces on the *old* host — so this is not a migration bug.

## The measurement that decides the design

Flow serves an `en` shell and the app rewrites `lang` during hydration. Sampling a pt account after `goto` returns, two consecutive loads:

| t (ms) | `lang` | `readyState` |
|---|---|---|
| 887 | `en` | interactive |
| **1510** | `en` | **complete** |
| 2092 | `en` | complete |
| **2488** | **`pt`** | complete |

**`readyState` is not a usable signal** — it reaches `complete` a full second *before* the flip, so the obvious "wait for complete, then read" is wrong. DOM-node count oscillates (136 → 300 → 249 → 251) and doesn't discriminate either. Nothing cheap predicts the flip, so `_settled_lang` observes it: read, bounded-wait for a change, re-read.

**A timeout is an answer, not a failure.** Either the account's locale genuinely equals the shell default, or the page never hydrated in the window. Both leave the first read as the best available answer — which is exactly the pre-#651 behaviour, so this branch can never be worse than what it replaces. It logs `client.account_locale_lang_unchanged` so the field can distinguish them.

## Cost, measured

An account whose locale **is** the shell default pays the 4 s bound once per process (`ffroliva` bootstrap ~2.7 s → ~6 s). That is the price of not guessing, and it is bounded; the alternative shipped a wrong locale for every account whose URL couldn't answer.

## Live proof — reported as two separate measurements, deliberately

Conflating these is the error [v0.66.1](https://github.com/ffroliva/gflow-cli/blob/develop/docs/LIVE_VERIFICATION_v0.66.1.md) made and v0.66.2 corrected:

- **A · component** (helper on a live page): naive `'en'` → `_settled_lang` **`'pt'`** in **638 ms**.
- **B · end-to-end** (real bootstrap, shell-default account): resolves `'en'` correctly, pays the 4 s bound.

Two zero-credit instruments ship with it, neither needing a migrated account: `scripts/dev/measure_html_lang_settle.py` and `scripts/dev/verify_lang_settle_fix.py`.

## Open questions — carried, not smoothed over

1. **A cold-load case I cannot explain.** On one session's *first* navigation, `lang` did not flip within 4 s — yet a fresh navigation 2 s later flipped in 638 ms. Either Flow genuinely served English on that load (its URL settle also timed out, so no redirect either), or 4 s is too short cold. The `lang_unchanged` event makes this visible in the field rather than silent.
2. **A demotion introduced in v0.66.2, not by this PR.** Folding only the URL-derived segment means a learned `pt` becomes `PROVISIONAL` whenever the URL settle times out. It self-heals on the next redirecting run and the locale is re-derived per run, so nothing is permanently lost — but it is the same *shape* of demotion #643 originally complained about, and worth a decision rather than a rediscovery.

## Test plan

- [x] Red-first tests; A/B control fails exactly 3 (the two new cases plus the probe-failure test, since the helper owns the error handling)
- [x] ruff / format / pyright (78 = `develop` baseline) / repo hygiene
- [x] Scoped suite on the changed area
- [ ] **Full local sweep inconclusive** — this machine was saturated and the run stalled; CI's three-version matrix is the gate
- [ ] Non-`en` account end-to-end on a *migrated* origin — needs the reporter's account

Refs #651
Refs #643

### DIFF
--- a/scripts/dev/measure_html_lang_settle.py
+++ b/scripts/dev/measure_html_lang_settle.py
@@ -1,0 +1,81 @@
+"""Does `<html lang>` change between goto-return and hydration, and what signals it? (#651)
+
+The reporter measured `locale=en` on a pt-BR account whose pages serve
+`<html lang="pt">`, and offered the mechanism as an explicit guess: the probe reads
+the initial HTML shell before the app sets `lang`.
+
+Run 1 confirmed the race on the OLD host (no migrated account needed):
+`en` until ~1.9 s, `pt` from ~2.9 s.
+
+This run adds `readyState` and a DOM-node count alongside `lang`, to find a
+*hydration signal* the fix can key on instead of a guessed timing constant.
+
+Zero credits: navigation only.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+from time import perf_counter
+from typing import Any
+
+sys.path.insert(0, r"C:\development\github\gflow-cli\src")
+sys.path.insert(0, r"C:\development\github\gflow-cli\scripts\dev")
+
+from gflow_cli.api import routes  # noqa: E402
+from gflow_cli.api.routes import locale_segment_from_lang_attr  # noqa: E402
+
+from _spike_common import build_client, resolve_profile_dir  # noqa: E402, isort: skip
+
+SAMPLES_MS = [0, 200, 400, 600, 900, 1200, 1600, 2000, 2400, 2800, 3200, 4000, 6000]
+
+_SNAP = """() => ({
+  lang: document.documentElement.lang || '',
+  ready: document.readyState,
+  nodes: document.querySelectorAll('*').length
+})"""
+
+
+async def _sample(page: Any, label: str, url: str) -> None:
+    t0 = perf_counter()
+    await page.goto(url, wait_until="domcontentloaded", timeout=45_000)
+    rows: list[tuple[int, str, str | None, str, int]] = []
+    prev = 0
+    for target in SAMPLES_MS:
+        if target - prev > 0:
+            await page.wait_for_timeout(target - prev)
+        prev = target
+        try:
+            snap = await page.evaluate(_SNAP)
+        except Exception as exc:  # noqa: BLE001
+            snap = {"lang": f"<{type(exc).__name__}>", "ready": "?", "nodes": -1}
+        rows.append(
+            (
+                round((perf_counter() - t0) * 1000),
+                snap["lang"],
+                locale_segment_from_lang_attr(snap["lang"]),
+                snap["ready"],
+                snap["nodes"],
+            )
+        )
+
+    print(f"\n--- {label}  {url}")
+    print(f"{'t_ms':>7}  {'lang':<8} {'seg':<8} {'readyState':<12} {'DOM nodes'}")
+    for ms, lang, seg, ready, nodes in rows:
+        print(f"{ms:>7}  {lang!r:<8} {seg!r:<8} {ready:<12} {nodes}")
+
+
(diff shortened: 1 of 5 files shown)


ffroliva:
**Correction to this PR's own test plan: the full local sweep did complete.**

I marked it *inconclusive* because the run appeared stalled at ~38% while this machine was saturated by leftover processes from the session. It was slow, not wedged, and finished clean:

```
1663 passed, 3 skipped, 12 warnings in 506.89s (0:08:26)   RC=0
```

Scope: `tests/api tests/features tests/mcp tests/worker -m "not live and not e2e and not smoke"`.

So the third checkbox is met. Recording the correction here rather than silently editing the body — a stale "inconclusive" is a smaller error than the one this branch exists to fix, but it is the same kind.

Issue #657, opened 2026-09-04
video: the output-count tab cascade is unscoped and unverified — a wrong match doubles spend


ffroliva:
Surfaced by the `/gflow:predict` round on the duration-probe work (2026-09-04). Filed separately because the blast radius is larger than the duration path and it deserves its own predict round rather than riding along on a plan scoped to duration.

## What

`_set_output_count` (`src/gflow_cli/api/transports/ui_automation_video.py:1636-1644`) builds the same inline, unscoped text cascade as the duration probe, clicks bare, and never reads back whether the click took effect. Its own comment claims *"the same pre-spend contract as the duration probe (#288)"*.

## Why it is worse than the duration case

A wrong duration match produces a clip of the wrong length — one generation, mis-sized. A wrong **count** match produces the wrong number of generations. `x1` → `x2` is a 2× spend; `x1` → `x4` is 4×. The user is billed for clips they did not ask for.

## Evidence

- The cascade is not scoped to the open settings popover, unlike `VIDEO_ASPECT_TAB_SELECTORS` (`:236-239`), which scopes one of its three candidates.
- `_probe_selector_cascade` takes `.first`, so the first matching element anywhere in scope wins.
- No `aria-selected` read-back after the click, so a click that lands on the wrong element still logs success.
- The count tabs are `x1`–`x4` (renamed from `1x` in #404), which are short, generic strings.

## What is NOT claimed

Nobody has demonstrated a stray element that actually collides here. This is the same LIKELY-not-CONFIRMED status the duration hazard has, and it should be settled the same cheap way before any code: extend `scripts/dev/capture_video_model_capability_matrix.py`'s `_menu_state` to replay the count cascade document-wide and dump matches. $0, no submit. If `strays == []` across models, close this.

## Suggested route

`/gflow:predict` first — it is a transport/selector change, which AGENTS.md gates. The duration plan at `docs/superpowers/plans/2026-09-04-duration-cohort-probe/` has the shape a fix would follow (fail-fast existence probe + per-raise `remediation_hint=`, no `errors.py` diff).

Refs #288, #650

PR #660, opened 2026-09-04
feat(recon): $0 stray-match probe, and stop the collector faking a positive

ffroliva:
## Summary

Two open gates need the same evidence before anyone writes code — **#657** (the count-tab cascade) and **task 7** of `docs/superpowers/plans/2026-09-04-duration-cohort-probe/PLAN.md` (the duration read-back). Both are unscoped `.first` probes over the same five roles, and `/gflow:predict` rated the hazard **LIKELY, not CONFIRMED** for both: nobody has demonstrated a real element that actually collides.

This answers it for $0, on the script that already walks that DOM.

## The probe

`_menu_state` replays the transport's cascade document-wide and reports only matches **outside** the open popover — exactly the elements `.first` could hit instead of the real tab:

```json
"duration_strays": [{"label":"8s","role":"button","tag":"BUTTON","id":"…","text":"…"}],
"count_strays":    []
```

**Kill condition:** empty lists across every model ⇒ scoping and read-back are unnecessary, and both tasks close unstarted. That is the outcome I expect, and it is worth $0 to know rather than assume.

## The bug it also fixes

The collector could manufacture the positive it exists to measure.

`scope = menu || document.body` was harmless when the tab scrape was `[role='tab']` only. #650 widened it to plain buttons — so a popover that failed to open would scrape **the whole page**, and any stray `8s` button would be recorded as a duration tab. This is the instrument that settles capability arguments; it must not be able to invent one.

The tab scrape is now menu-only and yields `[]` when no menu is open, with `menu_present` recording which happened, so a miss can never read as an absence. The text and credit reads keep the body fallback — they are descriptive, not capability claims.

## How to run it

```powershell
$env:PYTHONUTF8=1
.venv\Scripts\python.exe scripts\dev\capture_video_model_capability_matrix.py --profile denon82 --project <id>
```

Navigation and popover reads only, nothing submitted, no credits. Run it on `denon82` — `ffroliva` is fully migrated to `flow.google.com` and exits 36 immediately.

## Test plan

- [x] `pytest tests/scripts/test_capture_video_model_capability_matrix.py` — 3 passed
- [x] `ruff check src tests scripts/dev/capture_video_model_capability_matrix.py` — clean (the file has no pre-existing errors; `scripts/dev/` at large has ~54 that CI does not gate)
- [x] `check_repo_hygiene.py`, `check_doc_links.py`, `check_council_memory.py` — green
- [ ] **Not run against live Flow** — that is the point of the PR, and it needs a maintainer's Chrome session

## On the tests

The DOM scrape runs in the browser and cannot be exercised offline, so the two invariants that decide whether a capture can be **trusted** are pinned as source guards instead. That is weaker than a behavioural test and I would rather say so than dress it up — the pre-existing test covered only `_classify`, which neither this change nor #650 touches.

Refs #657, #650, #288

### DIFF
--- a/scripts/dev/capture_video_model_capability_matrix.py
+++ b/scripts/dev/capture_video_model_capability_matrix.py
@@ -131,16 +131,23 @@ async def _menu_state(page: Any) -> dict[str, Any]:
           // and `x{n}` (count) and takes `.first`. Any VISIBLE match outside the
           // open popover is an element the transport could click by mistake.
           // Empty lists across every model = scoping/read-back is unnecessary.
-          const STRAY_ROLES = [
+          // Mirrors of the transport's two cascades, which are NOT the same.
+          // ui_automation_video.py: duration probes five roles for `{n}s`;
+          // count probes ONLY [role='tab'], for BOTH affix orders `x{n}` and
+          // the #404 legacy `{n}x`. Scanning one shared list under-reported
+          // count (missed `2x`) and over-reported it (flagged buttons the
+          // transport can never click).
+          const DURATION_ROLES = [
             "[role='tab']", "[role='button']", "button",
             "[role='option']", "[role='menuitem']",
           ];
+          const COUNT_ROLES = ["[role='tab']"];
           const visible = (e) => !!(e.offsetParent || e.getClientRects().length);
-          const strayScan = (labels) => {
+          const strayScan = (labels, roles) => {
             const seen = new Set();
             const out = [];
             for (const label of labels) {
-              for (const role of STRAY_ROLES) {
+              for (const role of roles) {
                 for (const el of document.querySelectorAll(role)) {
                   if (menu && menu.contains(el)) continue;
                   if (!visible(el)) continue;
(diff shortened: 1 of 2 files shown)


ffroliva:
## Corrected after review — the probe under-reported

A reviewer caught the failure direction that actually matters here. An empty stray list is a **kill condition**: it closes #657 and task 7 of the duration plan with no code written. So a probe that under-reports silently deletes work that was needed — and the first cut under-reported.

**The two cascades are not the same, and I had treated them as one.** From `ui_automation_video.py`:

| | roles probed | labels |
|---|---|---|
| duration | 5 (`tab`, `button` role, `button` tag, `option`, `menuitem`) | `{n}s` |
| count | **only `[role='tab']`** | `x{n}` **and** `{n}x` (#404 legacy) |

Scanning one shared list missed a visible `2x` outside the popover — clickable by the transport, invisible to the probe — while simultaneously flagging plain buttons the count cascade can never click, which would have *invented* work instead. Both directions wrong, in the same six lines.

**The strays also never reached the per-model rows.** They existed only in the pre-model-select baseline, so this PR's own rule — "empty across every model" — was not answerable from the output it produced. Both row shapes now carry them, including the picker-miss row.

**The tests were vacuous, which is why none of this was caught.** `assert "menu_present:" in src` passed against unmodified `develop`. They now derive the expected label and role sets from the transport source itself, so the probe cannot drift from what it claims to replay.

Verified by mutation rather than assertion:

| Mutation | Result |
|---|---|
| drop the legacy `{n}x` affix | 🔴 `test_count_probe_scans_both_affix_orders` |
| widen count roles to all five | 🔴 `test_count_probe_mirrors_the_transport_exactly` |

Both were **green** against the first cut.

Gates: `pytest tests/scripts` 217 passed · `ruff check`/`format --check` clean · hygiene, doc-links, council-memory green.

Still not run against live Flow — that remains the point of the PR.

PR #669, opened 2026-09-05
# fix(639): the migrated model picker — `veo-lite-lp`, `[Lower Priority]` ambiguity, and the overlay a switch leaves behind

kittinan:
## Summary

Three defects in the migrated `flow.google.com` model picker (#639). The first is the
reported one; the other two were found while fixing it and are the more serious pair —
one bills the wrong tier, one breaks the next run outright.

**1. `--model veo-lite-lp` could not be selected at all.** On an account Google has
already moved, `_select_model` refused with `ConfigurationError` (exit 11, *"not available
on the migrated Flow host"*) before it ever opened the picker. The omission was
deliberate: the migrated map is keyed by product label, and that tier had no captured
label — the 2026-08-14 two-account capability matrix, #650's duration capture and
v0.61.0's refusal A/B all recorded a **picker MISS**. It is now matched the way the labs
driver has since #539: by the `[Lower Priority]` tag alone.

The entry has since been captured — a migrated account renders
`Veo 3.1 - Lite [Lower Priority]`, and its picker was already *defaulted* to that tier,
which is the likely reason earlier captures missed it (they were taken on accounts Flow
was not throttling). Matching stays on the tag rather than the now-known label: one
account's rendering, the labs driver keys off the same tag, and a tag Flow appends to
whichever tier it throttles survives that tier changing.

**2. `--model veo-lite` could silently generate on the throttled tier.** Entries were
matched by case-insensitive substring with `.first`, and the same substring ran against
the picker button's read-back. Against the real button text:

```python
"Veo 3.1 - Lite [Lower Priority] arrow_drop_down".lower().startswith("veo 3.1 - lite")
# -> True
```

So `--model veo-lite` returned *"already selected"*, never opened the menu, and generated
on the throttled tier — which Flow bills. This is the ambiguity #539 fixed on labs.google
(whose selectors carry `:not(:has-text('[Lower Priority]'))`); the port had dropped it.

**3. Switching `--model` broke the next run.** Reported from the field: previous run on
`veo-lite-lp`; run 1 on `omni-flash` dies, run 2 on `omni-flash` works, run 3 on
`veo-lite-lp` dies. Picking from the menu leaves Angular with **two** stacked overlays
(the settings pane and the menu opened over it) and each Escape dismisses exactly one, so
`_close_pane`'s single press left the settings pane covering the composer. `send_prompt`'s
click then failed actionability and surfaced ~5 s later as a bare `TimeoutError` naming
`[contenteditable='true']`, pointing nowhere near the pane. A repeat run binds the model
at the button read-back, never opens the menu, and never stacks the second overlay — which
is exactly why re-running the same model worked.

The click's message blames stability, and that is a red herring. Sampling every 100 ms for
12 s after a switch:

| | composer bounding box | `.cdk-overlay-pane:visible` |
|---|---|---|
| after a switch | **identical for 12 s** | **1**, for the whole window |
| after a repeat | identical | 0 |

`migrated.pane_still_open` *does* fire on the pre-fix source — confirmed by replaying it
against the live host, not assumed. It was a warning the run ignored; only the consequence
was missing, and it is now the failure itself.

### What changed

- `ModelMenuMatcher(contains, excludes)` replaces the bare label lookup and drives **both**
  the button read-back and the menu scan, so the two cannot drift.
- The menu is matched in Python over `all_text_contents()` rather than a `has_text` filter:
  an *exclusion* is not expressible as `has_text`, and more than one hit must **refuse**
  (naming every candidate) instead of resolving `.first`.
- `_close_pane` escapes until no overlay is visible (bounded at 4) and raises pre-submit at
  $0 if one will not close.
- New log events `migrated.model_already_selected` (the read-back short-circuit previously
  emitted nothing at all) and a counted `migrated.pane_still_open`.
- **Routing deliberately unchanged:** `migrated_can_serve` still gates on
  `VIDEO_MODEL_MENU_LABELS`, so `veo-lite-lp` does not pull an *unmoved* account onto the
  new host. A moved account is routed by its URL and never reaches that question.

Refs #639. Not filed as its own issue — reported directly against v0.67.0.

## Lifecycle

- [ ] ~~Issue triaged with `issue-assessment`~~ — **not run.** This arrived as a direct
  field report against v0.67.0, not a GitHub issue. Surfaces: **CLI only** — no MCP or CLI
  flag surface is touched (`--model veo-lite-lp` already existed and already validated).
- [ ] ~~`predict` verdict recorded~~ — **not run**, and this does touch selectors. It is a
  pre-implementation gate and the work began as a one-line map addition; defects 2 and 3
  were then found by measurement rather than design. Running it retroactively would be
  theatre — say the word and I will run it against the diff instead.
- [ ] ~~`scenario` + `plan` under `docs/superpowers/plans/`~~ — **not written**, same
  reason. The investigation is written up instead as
  `docs/superpowers/spikes/2026-09-05-migrated-model-menu-lower-priority.md`, with the
  captures, the measurements and an explicit "recorded, not asserted" section.
- [ ] `check` — **components run individually, not the skill**, and the step 1b CLI↔MCP
  mirror sweep was not run as such. What did run is listed under Validation, plus the full
  MCP suite (`tests/mcp/`, 168 passed).
- [x] ~~MCP twin~~ — **not applicable.** `git diff origin/develop...HEAD -- src/gflow_cli/mcp/
  src/gflow_cli/cli_video.py src/gflow_cli/api/video.py` is **empty**; this changes only
  the migrated transport's internals. No new CLI surface to mirror.
- [x] **Live-verified against real Flow.** Details below, including what could NOT be
  verified.
- [ ] ~~Council review~~ — **not run.** Happy to run `branch-review` before this is looked
  at if you would rather see it first.

### Live verification

Instrument: `scripts/dev/capture_migrated_model_menu.py` (new). Selection happens entirely
before submit — what made v0.61.0's and v0.62.1's refusal matrices free — so all of this
is **$0** except where noted.

`--drive`, the shipped `_select_model` for all five tiers, reading the picker back after
each switch (the only evidence separating "clicked the entry the user asked for" from
"returned success"):

| requested | outcome | button read back after |
|---|---|---|
| `omni_flash` | selected | `Omni 1.1 Flash` |
| `veo_3_1_lite` | selected | `Veo 3.1 - Lite` |
| `veo_3_1_fast` | selected | `Veo 3.1 - Fast` |
| `veo_3_1_quality` | selected | `Veo 3.1 - Quality` |
| `veo_3_1_lite_lower_priority` | selected | `Veo 3.1 - Lite [Lower Priority]` |

Both directions of the exclusion are exercised: plain Lite lands on the plain entry with
the throttled sibling in the same menu, and the throttled tier is reached *through* the
menu.

The reported sequence replayed after the fix — `omni-flash` (switch) / `omni-flash`
(repeat) / `veo-lite-lp` (switch): all three apply settings and type the prompt, with
`visible_overlays_after_settings: 0`.

**Two real generations** (not $0): one on `veo-lite-lp`, and one end-to-end on the switch
path (`veo-lite-lp` → `omni-flash`, `migrated.model_selected` in the timeline) that
submitted, generated and downloaded a 2.6 MB clip, exit 0.

**Could NOT be verified:**

- **One account, one locale (`en`).** Whether Flow serves the `[Lower Priority]` entry to
  accounts it is *not* throttling is untested, and every earlier capture suggests it does
  not. The routing gate is left conservative for exactly this reason.
- **The ambiguity refusal has no live case.** This account's menu yields exactly one hit
  per matcher, so the >1 branch is covered offline only.
- **`_open_pane` cannot be called twice on one page load** — re-opening the settings pane
  leaves a detached overlay that still contains radiogroups and `.last` resolves to it
  ("0 option groups"; reproduced twice while building the probe, which now loads a fresh
  document per tier). Production opens the pane once per run, so no shipped path is known
  to hit it. Recorded in the spike, not filed.

## Validation

- [x] Focused tests added — **9 new**, of which **8 fail against `origin/develop`'s
  source** and pass here. The ninth
  (`test_repeating_the_same_model_never_stacks_the_second_overlay`) passes on both **by
  design**: it pins the working half of the reported asymmetry. The fake DOM now models
  what the live host does — a menu pick leaves an overlay that eats the next Escape,
  `:visible` excludes the detached pane, and a covered composer is not clickable.
- [x] Documentation updated — `CHANGELOG.md`, `KNOWN_ISSUES.md` +
  `website/docs/KNOWN_ISSUES.md`, and the new spike.
- [x] `uv run python scripts/ci/check_doc_links.py` → `All links resolved across 29 files.`
- [x] `uv run ruff check src tests` → `All checks passed!` (plus `ruff format --check`,
  426 files already formatted)
- [x] `uv run pyright src` → `0 errors, 0 warnings, 0 informations`
- [x] Relevant pytest commands:
  - `uv run pytest -q tests/api/transports/test_migrated_composer.py` → **34 passed**
  - `uv run pytest -q tests/mcp/` → **168 passed**
  - full suite → **3820 passed, 12 skipped**

One unrelated pre-existing failure, reproduced on an untouched `origin/main` worktree:
`tests/features/test_auth_steps.py::test_status_of_a_profile_with_a_live_flow_session`
asserts `"experiments" in output`, but Rich wraps the long basetemp path mid-word
(`profile_expe\nriments`). Environment/path-length artifact, not from this branch.

**Worth flagging for review:** both the offline suite and the `--drive` matrix missed
defect 3, because neither touched the composer *after* a real switch. The new switch-path
tests close that gap, but the matrix alone is not sufficient coverage for this driver.

## Contribution Checklist

- [x] Commits use a real Git identity (`Kittinan <tunxedge@gmail.com>`)
- [ ] **`Signed-off-by:` is missing.** This is an external contribution and the three
  commits carry no DCO trailer. Needs
  `git rebase --signoff origin/develop` (or a re-commit with `-s`) before merge.
- [x] No secrets, cookies, tokens or signed URLs. Scanned the new files; what the spike
  and probe do contain is a **profile name, project ID and media IDs** from the capture —
  the same class of identifier existing `LIVE_VERIFICATION_*.md` docs carry, but they are
  account data, so redact them if you would rather they were not in the repo.
- [ ] AI-assisted changes reviewed before submitting — **author to confirm.** These commits
  were written by Claude Code with live measurement; the diff, the spike's claims and the
  two real generations should be read before this is opened.


ffroliva (changes requested):
# PR #669 — Council Review Verdict

**Reviewed at `1901d85` (pinned).** No SHA drift during review. Protocol: `skills/pr-council-review/SKILL.md` v2.1.

## Consensus: 🔴 RED — two must-fix items, both small

Let me say up front that the evidence quality here is the best I have reviewed on this repo: a $0 instrument, a read-back matrix that separates "clicked what was asked" from "returned success", a measured refutation of the stability red herring, and an explicit "could NOT be verified" section. Defects 2 and 3 are real, well-diagnosed and worth more than the reported one. The blockers below are narrow and none of them touch the core fix.

| Dim | Verdict | Headline |
|---|---|---|
| D0 CI-mechanical | 🟢 | All six gates re-run at `1901d85` + ruff/pyright/34 tests. 14/14 GitHub checks green. |
| D1 Correctness | 🟡 | `_close_pane` raising inside `finally` masks the precise error (reproduced). |
| D2 Code quality | 🟡 | `_close_pane` widens a selector that `_open_pane` deliberately narrows. |
| D3 Security | 🟡 | Profile name + project UUID in the spike; the PII gate does not cover that path. |
| D4 Tests | 🟢 | 9 new, 8 red pre-fix. Affected surface exercised. |
| D5 Memory hygiene | 🔴 | Two memory files still record the claim this PR disproves. |
| D6 UI / live-verify | 🟡 | Live evidence strong; `[Lower Priority]` remains an English prose anchor. |
| D9 Docs drift | 🟢 | CHANGELOG + KNOWN_ISSUES + mirror consistent; links and mirror gates pass. |
| D13 Scripts | 🟢 | New probe follows the `_spike_common` sibling convention. |
| D14 YAGNI | 🟢 | One nit: `PANE_CLOSE_ESCAPES = 4` where 2 is the measured need. |
| D15 Surface parity | 🟢 | Your "MCP twin not applicable" claim **verified** — diff on `mcp/`, `cli_video.py`, `api/video.py` is empty. |

---

## Must-fix (2)

### 1. 🔴 D5 — the memory store still says this tier is a picker MISS

`docs/superpowers/memory/video-model-capability-matrix.md` records, as current state:

- `:103` — `| veo_3_1_lite_lower_priority | **PICKER MISS** (still) | — | — |`
- `:144` — *"**What is NOT settled:** … whether `lower_priority`'s miss is selector or absence."*

and `flow-capabilities-are-cohort-dependent.md` carries *"(`lower_priority` missed its picker, so it stays unmeasured)"*.

This PR **settles that open question**: the entry exists, renders as `Veo 3.1 - Lite [Lower Priority]`, and the miss was cohort/absence — not selector drift — because Flow serves it to accounts it is actively throttling. That is exactly the fact the memory store exists to carry, and right now it points the next agent the wrong way. The spike is the capture; memory is the routing layer, and only the spike was updated.

Concretely: update the `:103` row, retire the "not settled" clause at `:144`, and amend the cohort file's "stays unmeasured" parenthetical. `check_council_memory.py` stays green either way (47 files, all cited) — this is a content edit, not a wiring one.

### 2. 🟡→blocking D1 — `_close_pane` raising in `finally` replaces the real error

`apply_video_settings` calls `_close_pane` in a `finally:` (`migrated_composer.py:224-225`). It previously could only warn; it can now raise. When the `try` body already failed, the new raise **discards the actionable diagnostic**. Reproduced against your own fake DOM:

```
RAISED : UiSelectorDriftError — "2 overlay(s) still visible after 4 Escape presses" (exit 23)
CONTEXT: ConfigurationError — "model 'veo_3_1_quality' is not offered on this account's
         migrated Flow host; offered: Omni 1.1 Flash, Veo 3.1 - Fast" (exit 11)
```

The user gets exit **23** ("selectors drifted, file a bug") instead of exit **11** ("pass a different `--model`"), and the offered-models list is gone. `__context__` does not survive into the Problem Details output, so it is invisible. Suggested shape: only raise from `_close_pane` when no exception is in flight — or let `apply_video_settings` log-and-swallow the close failure when the body already raised.

---

## Nice-to-have (3)

1. **`_close_pane` is broader than the bug requires** (`:279-297`). It aborts on *any* visible `.cdk-overlay-pane`, while `_open_pane` (`:236`) deliberately narrows the same selector with `.filter(has=RADIOGROUP)` — the file already knows the class is not unique to the settings pane. `.cdk-overlay-pane` is the generic Angular CDK class (snackbars, tooltips, dialogs), and Escape does not dismiss a snackbar. A transient toast would now hard-abort a run that used to succeed. Cost is $0 and pre-submit, so this is a false-abort risk rather than a credit risk — but scoping the count to panes containing `RADIOGROUP`/`MENU_ITEM` would keep the fix and drop the blast radius.

2. **PII in the spike** — `docs/superpowers/spikes/2026-09-05-…md:3-4` carries profile `kittinansr2` and project `66324c59-…`. You flagged this yourself. Worth noting the gate gap: `check_website_docs_pii.py` scans only `website/docs/` (25 files), so `docs/superpowers/` is unscanned and CI is green regardless. Maintainer's call, but the redaction is two lines.

3. **`PANE_CLOSE_ESCAPES = 4`** where the measurement says 2. The comment calls the headroom deliberate, which is fair — flagging only because D14 asks.

---

## Correcting one thing in the PR body

> **`Signed-off-by:` is missing.** … Needs `git rebase --signoff origin/develop` before merge.

**This is not the case** — all three commits already carry the trailer:

```
1901d85  Signed-off-by: Kittinan <tunxedge@gmail.com>
7baa566  Signed-off-by: Kittinan <tunxedge@gmail.com>
ea1f8c0  Signed-off-by: Kittinan <tunxedge@gmail.com>
```

No rebase needed. Worth fixing the checkbox so it does not cost anyone a round trip.

---

## Confirmed-good

- **D0 green at the reviewed SHA**, run in a detached worktree, not inherited from your log: `ruff check` (all passed), `ruff format --check` (428 files), `pyright src` (0/0/0), `check_doc_links` (29 files), `check_repo_hygiene` (947 files), `check_website_docs_pii`, `check_council_memory` (47 files), `generate_website_docs --check` (mirror in sync), and `pytest tests/api/transports/test_migrated_composer.py` → **34 passed**.
- **The `.first` → refuse change is right, and matches labs.** `ui_automation_video.py:265-291` reasons through this exact hazard and reaches the same conclusion; the port now carries the `:not(:has-text('[Lower Priority]'))` semantics it had dropped, on **both** the menu scan and the button read-back. Driving both from one `ModelMenuMatcher` is the correct structural fix — the two paths drifting is what caused defect 2.
- **Matching on the tag rather than the newly-captured label is the right call**, and for the reason you give. `VEO_3_1_LITE_LOWER_PRIORITY` keys off the same tag on labs (`ui_automation_video.py:291`), so the two drivers stay symmetric.
- **Routing left alone** is the conservative choice and is well argued: `migrated_can_serve` still gates on `VIDEO_MODEL_MENU_LABELS`, so one account's capture cannot pull an unmoved account onto the new host.
- **`migrated.model_already_selected` closes a real observability hole** — a silent short-circuit on a billing-relevant path was worth a log line.
- **The test fake now models the mechanism, not the symptom**: one overlay dismissed per Escape, `:visible` excluding the detached pane, and a covered composer that is genuinely unclickable. That is why 8 of the 9 fail pre-fix.

## Note on gates

`SonarCloud analysis` is **`skipped`**, not green — expected on a fork PR, so it is maintainer-checked here rather than a signal from this PR. Per AGENTS.md it cannot be called merge-ready on the Sonar axis until checked from a maintainer branch.

## On the skipped lifecycle phases

Your accounting of what was and was not run is honest and I would not ask for retroactive `predict`/`scenario` theatre. The one gap with teeth is `check` step 1b — but D15 above verifies the mirror axes hold by inspection, so it is satisfied in substance.

---
_Generated by [Claude Code](https://claude.ai/code)_

ffroliva:
**Re-review at `2731f4d` — RED cleared. Merged `develop` onto your branch to resolve the CHANGELOG conflict.**

Both must-fix items and all three nice-to-haves from the earlier council are addressed, verified against the SHA rather than the commit message:

| Finding | State at `2731f4d` |
|---|---|
| D5 memory rows still said PICKER MISS | `video-model-capability-matrix.md` rows at :17/:106 now read "absence, not a broken selector; settled 2026-09-05", the "NOT settled" clause is struck, and the cohort file carries the generalisation. |
| D1 `_close_pane` raising in `finally` masked the real error | `apply_video_settings` now catches `BaseException`, closes with `strict=False` (warn only) and re-raises; `strict=True` only on the success path. |
| D2 selector broader than `_open_pane` | `_blocking_overlays` filters `.cdk-overlay-pane:visible` to panes containing a radiogroup or menuitem. |
| D14 `PANE_CLOSE_ESCAPES = 4` | Now 2. |
| D3 PII in the spike | Profile and project id removed; the header says so. |

Mechanical gate re-run in a detached worktree at `2731f4d`: ruff, format, doc-links, hygiene, website PII, council-memory, website mirror, `pyright src` (0/0/0), `test_migrated_composer.py` → 37 passed. After the merge with `develop`: 39 passed, all doc gates green.

**What I pushed to your branch** (`9da2502`, using the maintainer-edit permission you granted): a merge of `origin/develop` — v0.68.0 shipped an hour ago and its back-merge emptied `[Unreleased]`, so your `### Added` and `### Fixed` entries are re-homed under the new `[Unreleased]` section and `[0.68.0]` keeps only what shipped. No code changes beyond git's clean auto-merge of `migrated_composer.py` and its tests (your `_close_pane` work and #672's submit-enable poll coexist). If you had local commits on top of `2731f4d`, pull before pushing again.

**Evidence bar, forward-looking only.** This PR carries live-verify evidence (two Veo generations, the $0 read-back matrix) and meets the current bar. Open PR #675 will additionally require a `tests/e2e/` test, run or written, for every behavior change. It is not applied retroactively here; flagged for the maintainer, not held against this PR.

Merging once CI reports on `9da2502`. SonarCloud stays skipped on fork PRs and is maintainer-checked.

Issue #3, opened 2026-05-12
HTTP 401 on batchGenerateImages


gaidehaha:
The CLI can successfully create projects (labs.google), but image generation (googleapis.com) returns 401.

**Log:**
- Creating project... (Success)
- Generating image... (Authentication expired: HTTP 401)

**Observation:**
It seems the `aisandbox-pa.googleapis.com` endpoint now strictly requires the `Authorization: Bearer` header, which is missing in the current `FlowApiClient` implementation (it only sends cookies). Even with `Origin` and `Referer` headers patched, the 401 persists.

**Environment:**
- gflow-cli v0.4.0a2
- Windows / Playwright

ffroliva:
Hey @gaidehaha — thanks for the detailed report and the spot-on diagnosis. Your read on the `aisandbox-pa.googleapis.com` 401 is essentially correct: v0.4.0a2's `FlowApiClient` sent cookies only and did not attach the `Authorization: Bearer` header that the `batchGenerateImages` endpoint expects.

The good news: **v0.5.0a1 just shipped today** and resolves this. The release is on PyPI now:

```bash
pip install --upgrade gflow-cli==0.5.0a1
```

### What changed

The image-generation path is now backed by a new pluggable transport layer (`gflow_cli.api.transports`). v0.5.0a1 ships four strategies; we investigated each end-to-end against real Flow and made `ui_automation` the default:

| Strategy | Approach | v0.5.0a1 status |
|---|---|---|
| `ui_automation` (default) | Playwright drives Flow's editor on a logged-in profile — Google's own JS attaches the auth | **Production-validated**, ships as default |
| `evaluate_fetch` | `page.evaluate(fetch(...))` from a logged-in Playwright page; browser session handles auth | experimental — works against current Flow, kept gated |
| `bearer` | Capture Bearer token once via Playwright, then pure httpx with TTL refresh | experimental — Bearer alone produces 401/403; the API also validates per-request fields a live browser session generates |
| `sapisidhash` | SAPISID cookie + SHA1 HMAC, pure httpx | experimental — Chromium 80+ stores SAPISID encrypted, so reading it from a Chromium profile dir typically fails |

The three experimental strategies stay in `transports/experimental/` for research. CLI `--transport` only exposes `ui_automation` by default; set `GFLOW_CLI_EXPERIMENTAL_TRANSPORTS=1` to surface the others, or call them directly via the Python `make_transport()` factory.

### Why not just add the Bearer header?

[...]

Issue #626, opened 2026-09-02
Feature Request: Support Initial and End Frames with Omni 1.1 Flash


cmaohuang:
Hi, thank you for building and maintaining gflow-cli! I really appreciate the work that has gone into keeping it compatible with Google Flow’s rapidly changing interface.

I noticed that v0.62.1 added support for selecting the renamed Omni 1.1 Flash model. However, the CLI still appears to reject --model omni-flash when both --initial-frame and --end-frame are provided.

Since Google Flow now officially supports first-and-last-frame generation with Omni 1.1 Flash, may I ask whether there are any plans to support the following combination?

- Omni 1.1 Flash
- Initial frame
- End frame

Thank you again for your time and for maintaining this project!

ffroliva:
Thank you for the kind words — and for a genuinely well-researched report. You separated the two axes correctly, which saved real time: v0.62.1 fixed the **model picker selector** (Flow renamed the entry `Omni Flash` → `Omni 1.1 Flash`, which broke our `has-text` match), and that is completely independent of the **end-frame capability gate** you were hitting. Two different things; fixing one never touched the other.

**You were right, and it's fixed.** The change is in #627, shipping in **v0.64.0**.

### What was actually going on

`--model omni-flash --end-frame` was rejected up front with exit 17 (`ModelModeIncompatibilityError`). That guard was deliberate, not a bug — and it existed for a specific reason. On 2026-05-30 we captured Flow *accepting* omni-flash frame bindings in the UI and then **silently dropping them at submit**, routing the request to text-to-video and billing a credit for a clip that ignored the frames entirely. Since gflow spends your own Google credits, we fail closed on any model/mode combination we can't prove on the wire.

Your premise checked out: Flow shipped first-and-last frames for Omni 1.1 Flash, so the "coming soon" note our guard cited had gone stale. What was still missing was our own wire-level proof — so we went and got it.

### The evidence

We ran a route-abort probe that binds both frames through the real production code path, then kills the submit XHR in the browser before it reaches Veo. **Zero credits.** On two different Google accounts:

| Account | Captured route | startImage | endImage |
|---|---|---|---|
| A | `video:batchAsyncGenerateVideoStartAndEndImage` | ✅ non-null | ✅ non-null |
| B | `video:batchAsyncGenerateVideoStartAndEndImage` | ✅ non-null | ✅ non-null |

Two accounts, so this isn't a single-account artifact or a visibly staged rollout.

[...]

PR #683, opened 2026-09-06
feat(639): r2v on the migrated host

kittinan:
## Summary

**`gflow video r2v` from local `--ref` files now runs on the migrated `flow.google.com`
host.** On a moved account it exited 36: the composer took t2v, and since your i2v slice a
local start frame, and refused everything else.

Refs #639. Not filed as its own issue — reported directly against v0.67.0 by a user whose
account has been moved.

Each `--ref` file is uploaded through **your `_upload_via_toolbar`**, so the app's own
`maseQ` reply names the media id, and is then attached as an **`@` mention** in the prompt.
References are not a chip slot on this host — that is the one structural difference from
i2v, and it drives everything else.

Three details had to be measured, not assumed:

| | |
|---|---|
| submit rpc | **`MZZa6b`** — t2v is `YhhmEf`, i2v `eb1hJf` |
| commit gesture | **Enter**. A typed query alone leaves the picker open and inserts nothing |
| typing | mentions need real key events; prompt text keeps `insert_text` so a newline cannot submit early — the two cannot share a path, hence `send_prompt(append=)` |

Watching only `YhhmEf`/`eb1hJf` is why this looked, through several rounds of recon, like
r2v never submitted at all.

**Safety mirrors your i2v backstop, because the failure mode is identical and expensive.**
`_r2v_body_problem` asserts the submit body carries a `_r2v_` model key and **every**
uploaded id, and a run whose references have not all attached is refused *before* submit —
the alternative being a full-price clip with none of the user's references on it.

**Gate:** local files only, exactly as i2v is local-frame only. A reference by `@Name` or a
character entity has no media id in the picker's DOM to anchor on, so it stays on labs.
Your `test_run_video_rejects_modes_not_yet_ported_with_exit_36` asserted r2v was refused
wholesale; it now asserts the two forms that still are, and says why.

Full capture, including the rounds that failed and why:
`docs/superpowers/spikes/2026-09-05-migrated-r2v-attach-surface.md`.

### A note on the rebuild

This branch had a working r2v implementation before `develop` moved 55 commits under it.
Your i2v slice landed an upload path strictly better than the one it had grown — mine
polled an `Uploading<name>` label and never learned the media id at all — so the merge
resolves `migrated_composer.py` / `client.py` to **develop wholesale**, drops the feature,
and re-applies it on your infrastructure in the next commit. Two upload implementations in
one file, one of them the weaker, is not a merge.

## Lifecycle

- [ ] ~~Issue triaged with `issue-assessment`~~ — **not run.** A direct field report against
  v0.67.0, not a GitHub issue. Surfaces: **CLI only** — no MCP or CLI flag surface is
  touched; `--ref` already existed and already validated.
- [ ] ~~`predict` verdict recorded~~ — **not run**, and this is a transport/selector change,
  so the gate applies. It is a *pre*-implementation gate, and the work began as recon into
  why exit 36 happened; the design fell out of measurement rather than being chosen up
  front. Running it retroactively would be theatre. Say the word and I will run it against
  the diff instead.
- [ ] ~~`scenario` + `plan` under `docs/superpowers/plans/`~~ — **not written**, same
  reason. The investigation is written up as the spike above, which carries the
  measurements and an explicit "what this does NOT settle" section.
- [ ] `check` — **components run individually, not the skill**, and the step 1b CLI↔MCP
  mirror sweep was not run as such. What did run is under Validation, plus `tests/mcp/`.
- [x] ~~MCP twin~~ — **not applicable.** `git diff origin/develop...HEAD -- src/gflow_cli/mcp/
  src/gflow_cli/cli_video.py src/gflow_cli/api/video.py` is **empty**. No new CLI surface to
  mirror; this changes the migrated transport's internals only.
- [x] **E2E evidence** —
  `tests/e2e/test_migrated_host_e2e.py::test_e2e_r2v_binds_local_references_on_the_migrated_host`,
  added by this PR (the `e2e_video` twin of your t2v test):

  ```
  GFLOW_CLI_E2E_PROFILE=… GFLOW_CLI_E2E_PROJECT=… \
    uv run pytest -q -m e2e_video \
    tests/e2e/test_migrated_host_e2e.py::test_e2e_r2v_binds_local_references_on_the_migrated_host
  1 passed in 88.06s (0:01:28)      # one clip billed
  ```

  It pins that `migrated.references_attached` reports as many media ids as references
  given, so a run that quietly attached one of two fails there rather than in the field.
  Its fixture is a real 1024×1024 PNG, not a 1×1: the uploader answered `maseQ` **200 with
  no media id** for a 64×64 synthetic one (`MediaUploadRejectedError`). That attempt failed
  at upload and cost nothing.
- [x] **Live-verified against real Flow.** Details and limits below.
- [ ] ~~Council review~~ — **not run.** Happy to run `branch-review` before you look, if you
  would rather see it first.

### Live verification

The reported command, exactly — two local refs, `--model veo-lite-lp`:

```
migrated.frame_uploaded        media_id=ee23b02b-…      status=200
migrated.frame_uploaded        media_id=8ab6c3c6-…      status=200
migrated.references_attached   count=2  media_ids=[ee23b02b-…, 8ab6c3c6-…]
migrated.submit_observed       rpc=MZZa6b  workflow_id=278e2797-…
migrated.result                done=true  bytes=3434223
Saved: manual-test/7912c0a5-….mp4                        exit 0
```

The body assertion is armed inside `submit_and_observe`, so reaching `migrated.result`
without it firing is itself the evidence that both ids rode the submit.

**Semantically confirmed by the account owner**, which is the layer no log reaches: in the
output the **presenter is the person from the first reference** and the **product is the
one from the second**. Both references are genuinely *bound*, in the order given — not
merely accepted. Same distinction that caught a silently dropped end frame on labs in
v0.64.0.

**Could NOT be verified:**

- **The avatar wire slot.** A `likeness` chip was produced during recon but never
  submitted; only `media` has been exercised end to end.
- **Character entities on this host.** The attach works — a chip with a real `entity_id` —
  but the gate routes them to labs and no entity run has been submitted.
- **More than two references, and the per-model caps** (omni 7, veo lite/fast/lite_lp 3).
  Two is the most ever attached.
- **One account, one locale (`en`).**

## Validation

- [x] Focused tests added — **12 new**: 6 in `tests/api/transports/test_migrated_r2v.py`
  (routing, the submit rpcs, `_r2v_body_problem`'s branches, chip-per-mention, the refusal
  when one does not attach, prompt-append), plus the e2e above. The unit doubles are
  purpose-built rather than bolted onto `test_migrated_composer.py`'s fake — that one models
  the settings pane and model menu; this path needs a mention picker and chips.
- [x] Documentation updated — `CHANGELOG.md`, `KNOWN_ISSUES.md`, the spike, and
  `website/docs/` **regenerated** with `scripts/ci/generate_website_docs.py` (I hand-edited
  the mirror first and `test_committed_mirror_is_in_sync` caught it).
- [x] `uv run python scripts/ci/check_doc_links.py` → `All links resolved across 29 files.`
- [x] `uv run ruff check src tests` → `All checks passed!` (plus `ruff format --check`,
  443 files already formatted)
- [x] `uv run pyright src` → `0 errors, 0 warnings, 0 informations`
- [x] Relevant pytest commands:
  - `uv run pytest -q tests/api/transports/test_migrated_r2v.py` → **6 passed**
  - `uv run pytest -q tests/api/transports/test_migrated_dispatch.py` → **20 passed**
  - `uv run pytest -q tests/mcp/` → **168 passed**
  - full suite → **3950 passed, 12 skipped**
  - `uv run python scripts/ci/check_website_docs_pii.py` → clean across 25 files

One unrelated pre-existing failure, reproduced on an untouched `origin/main` worktree:
`tests/features/test_auth_steps.py::test_status_of_a_profile_with_a_live_flow_session`
asserts `"experiments" in output`, but Rich wraps the long basetemp path mid-word
(`profile_expe\nriments`). Environment/path-length artifact, not from this branch.

## Contribution Checklist

- [x] Targets `develop`
- [x] Commits use a real Git identity
- [x] `Signed-off-by:` on every commit — verified per-commit rather than assumed, because
  `git rebase --signoff` silently skips a commit completed via `--continue` after a
  conflict (it bit this fork on #669)
- [x] No secrets, cookies, tokens, signed URLs, or private captured data. The spike carries
  no account identifiers: the profile name and project id were removed after your review on
  #669, and what mattered was never the identity but that Flow was throttling the account.
- [ ] AI-assisted changes reviewed before submitting — **author to confirm.** Written by
  Claude Code against live measurement; the diff, the spike's claims and the e2e result
  should be read before this is opened.


ffroliva:
Heads-up in case it saves you some debugging on this PR — **there is a bug on `develop` that makes the e2e suite go red specifically on migrated accounts**, which is exactly the surface this PR works on. It is not caused by anything here.

`FlowApiClient.health_check()` accepted only the old host:

```python
return hostname.endswith(".google") or hostname == "google.com"
```

A migrated account's page sits on `flow.google.com`, which ends in `.com` and is not `google.com` — so `health_check()` returned `False` for a perfectly healthy session while `verify_flow_profile` correctly reported `AUTHENTICATED`.

Two e2e tests fail because of it on any migrated profile:

- `tests/e2e/test_auth_verification_e2e.py::test_e2e_verify_flow_profile_authenticated`
- `tests/e2e/test_transports_e2e.py::test_e2e_health_check_returns_true_when_active`

Filed as #690, fix is up as #693 (one line, plus a parametrized boundary test). Verified against a live migrated profile: the auth e2e goes 1 failed → 3 passed, and the transports health-check pair passes.

So if you have been seeing either of those red, it is not yours. Once #693 lands, a rebase should clear them.

No file overlap with this PR — #693 only touches `api/client.py` and `tests/api/test_client_image.py`, and this PR touches `migrated_composer.py` and its own tests. Nothing to reconcile, purely a signal-quality thing.

ffroliva:
## 🏛️ Council review — PR #683 @ `33d1b8b`

10-dimension council (6 baseline + D6/D9/D13/D15 adaptive). **D0 mechanical gate GREEN** — `ruff check`, `ruff format --check`, `check_doc_links.py`, `check_repo_hygiene.py` all pass at the reviewed SHA, and CI is green on 3.11/3.12/3.13.

Every finding below was **re-verified by hand** against `33d1b8b` before posting; commands and line numbers are from that SHA.

> ⚠️ D2 (code quality) is still running and will be posted as a follow-up. Everything else is complete.

### Verdict: 🔴 RED — one real must-fix, then this is close

Let me lead with what stands out: **the live-verification evidence here is the best tier this project has.** A real e2e run (`1 passed in 88.06s`), genuine `migrated.*` structlog lines, a saved `.mp4`, one clip billed, *and* a semantic cross-check that the two references bound in the right order. That last part — distinguishing "accepted" from "bound" — is the layer most people skip. The spike doc's explicit "What this does NOT settle" section is exactly the PROVEN/ASSUMED discipline this repo asks for. Thank you for that.

The RED is a single wiring bug plus doc drift, not a design problem.

---

## Must-fix

**1. 🔴 The r2v wire-safety check never runs.** `migrated_composer.py:1037`

`on_request` was correctly extended to branch on `expect_reference_ids` (line 974), but the listener registration was not:

```python
1037:        if expect_media_id is not None:
1038:            page.on("request", on_request)
```

`media_id` is set only under `Mode.I2V` (line 1261), so on the r2v path `expect_media_id` is always `None` → the listener is never attached → `_r2v_body_problem` never executes in a live run. It is unit-tested in isolation, which is why every gate stayed green.

```python
if expect_media_id is not None or expect_reference_ids:
```

...plus the mirrored cleanup at line 1105.

**Consequence for the PR body:** it states the assertion is *"armed inside `submit_and_observe`, so reaching `migrated.result` without it firing is itself the evidence that both ids rode the submit."* That inference does not hold — the check is inert for r2v. The live run genuinely worked and the references genuinely bound, but *despite* the assertion, not because of it. Worth correcting so the claim doesn't outlive the review.

Mitigating: the pre-submit chip-count guard (lines 1267-1276) independently refuses to spend credits when the picker inserted nothing, so this is a defence-in-depth hole rather than an unguarded path. Residual risk is a reference attaching as a chip but binding the wrong id in the RPC body.

**2. 🔴 `MODEL_KEY` never matches an r2v key.** `migrated_composer.py:99`

```python
MODEL_KEY = re.compile(r"[a-z0-9]+(?:_[a-z0-9]+)*_(?:t2v|i2v)_[a-z0-9_]+")
```

Verified: `veo_3_1_r2v_lite_low_priority` → no match. So in `_r2v_body_problem`, `key_text` is always `"no model key"` — the diagnostic is factually wrong for the one mode it was added to serve. Add `r2v` to the alternation.

**3. 🟠 Stale "r2v exits 36" claims — the #626 pattern.** `_unported_form` (line 205-215) now returns `None` for r2v with local `reference_images`, but these still tell users and agents otherwise:

| File | Line | Says |
|---|---|---|
| `src/gflow_cli/mcp/tools.py` | 1018-1019 | "a UUID frame, an end frame and 'r2v' return the exit-36-equivalent envelope" |
| `docs/MCP.md` | 86 | identical claim |
| `skills/gflow-cli/SKILL.md` | 278 | "only t2v and i2v … are ported" |
| `docs/USAGE.md` | 1747 | exit-36 table omits r2v |
| `docs/USAGE.md` | ~665 | `gflow video r2v` has no migrated-host callout (i2v has one at 617-624) |
| `KNOWN_ISSUES.md` / `website/docs/` | 19 | summary still lists r2v as "not ported yet and still exit 36" — contradicted by your own new paragraph 40 lines below |
| `docs/CONFIGURATION.md` | 364, 368 | lists r2v among modes that exit 36 |
| `docs/INDEX.md` | 152 | "everything except t2v/i2v exit 36" |

`mcp/tools.py:1018` is the important one — it's the literal text an agent reads before calling the tool, and AGENTS.md's second law was written from exactly this failure.

**4. 🟠 Capture scripts default output into the tracked tree.** Flagged independently by two dimensions.

`capture_migrated_attach_rpcs.py:221`, `capture_migrated_mention_gestures.py:154`, `capture_migrated_r2v_submit_payload.py:281`, `capture_migrated_upload_flow.py:277` each default `--out` to a bare filename written to CWD. Verified none is gitignored, while `scripts/dev/_spike_out/` **is** (`.gitignore:67`) and `_spike_common.default_out_path()` exists for this — your own sibling `capture_r2v_ingredient_lifecycle.py:240` already uses it. These reports carry decoded batchexecute payloads: prompt text, project ids, media/entity ids. No credential capture found (grepped clean), so this is accidental-commit hygiene, not a credential leak.

**5. 🟠 `CHANGELOG.md` has two consecutive `### Added` sections** under `## [Unreleased]` (lines 10 and 33). Found independently by two dimensions. Merge the r2v bullet into the existing section.

**6. 🟡 `_mention_by_name` has no retry on a picker miss.** `migrated_composer.py:806-830`

The module's own `_pick_frame_by_name` documents that this picker's search index can miss a fresh upload on first query and retries (`FRAME_SEARCH_ATTEMPTS`). `_mention_by_name` does a single query behind fixed `wait_for_timeout`s with no retry and no `wait_for(state="visible")` — the same failure mode you already proved happens, unguarded on the new path.

## Nice-to-have

1. `migrated_composer.py:116` — `MENTION_CHIP` is dead; `read_chips` hardcodes `'.mention-chip'` inline.
2. `capture_migrated_r2v_submit_payload.py:53` — `QUERY = "Me"` is unused; `_pick_first_reference` never filters, it just takes the highlighted option, so the comment misdescribes what the script verifies.
3. Spike doc "What this does NOT settle" retains bullets superseded by Rounds 6-7 in the same file (local-file upload, multiple references).
4. Spike doc states "not the `veo_3_1_lite_lower_priority` a t2v run sends" as settled, but no capture in the repo observed a t2v body at that tier — label it inferred.
5. Consider an `r2v (slice 2)` entry in the migrated-host wire-lessons ledger so the next session doesn't re-mine a 405-line spike.
6. ~7.8s of fixed `wait_for_timeout` per reference could be state-based waits.

## Confirmed good

- New selectors are Tier-1 structural (`INGREDIENTS_LIGATURE` icon ligature, `.mention-chip` class); picker text is used for error diagnostics only, never for selection — **no locale-invariance violation**.
- MCP parity axis A verified: `reference_images` round-trips `mcp/tools.py` → `worker/codec.py:268` on the same key, and MCP reaches the same `run_video` dispatch, so migrated r2v is inherited automatically. **No `mcp/tools.py` code change was needed** — only its docstring.
- `KNOWN_ISSUES.md` and its `website/docs/` mirror are word-for-word consistent.
- Typed errors with correct exit codes throughout; no bare exceptions. No regression to t2v/i2v traced.
- +2123 is mostly spike doc and recon instruments, following the repo's own precedent (13 existing `capture_*.py` paired to spikes) — not scope creep.
- All 26 new unit tests pass in 1.18s; no hardcoded account/profile/paths anywhere.

---

Only #1 and #2 are code. Everything else is text. Happy to push the fixes myself if you'd rather not context-switch — just say.

ffroliva:
### Addendum — `develop` moved under this PR, and one finding gets sharper

`develop` is now `9fba962`. Three merges landed since this PR opened, and two of them change what the review above means for you. Verified against `origin/develop`, not taken on trust.

**1. My must-fix #3 survived, at new line numbers.** #687 (`ea55f1b`) edited `src/gflow_cli/mcp/tools.py`, but the stale claim is still there — it just moved from **1018-1019 → 1055-1056**:

> "There the ported modes are 't2v' and 'i2v' with a local `initial_frame` and no `end_frame`; a UUID frame, an end frame and **'r2v' return the exit-36-equivalent envelope**."

Still false once this PR lands. Use the new line numbers after you rebase.

**2. New: entity-only r2v is now reachable over MCP, which sharpens the docstring fix.** `ea55f1b` relaxed the MCP r2v guard (`tools.py:694`) — it previously demanded `reference_images`, and now accepts `reference_entities` too, because `GenerateVideoRequest`'s own rule is "reference_images, ref_names, OR reference_entities" and the guard was stricter than the domain rule it fronts (#689).

Consequence for this PR: an MCP `r2v` request carrying **only** an entity id now passes the guard and reaches the transport, where your `_unported_form` (lines 205-215) returns `"character references"` → exit 36.

That behaviour is **correct** — character references genuinely aren't ported. But it's a path that could not be reached before `ea55f1b`, so the docstring fix has to draw the line precisely rather than just deleting "r2v":

- r2v with local `reference_images` → **now served** on the migrated host (this PR)
- r2v with `reference_entities` / `@Name` → **still exit 36**

Same distinction needed in `docs/MCP.md`, `skills/gflow-cli/SKILL.md`, and `docs/USAGE.md`. Credit for spotting the reconciliation goes to a parallel session reviewing the same area.

**3. `9039983` (#693) may clear an unrelated red for you.** `health_check()` matched only `.google`, so it returned `False` for every `flow.google.com` account — two e2e tests failed on any migrated profile. Fixed on develop. If you saw `test_auth_verification_e2e` or `test_transports_e2e::test_e2e_health_check_returns_true_when_active` red, that was never yours.

**4. Heads-up on a new CI gate.** `ea55f1b` also made `tests/mcp/test_cli_parity.py` check CLI **options** against MCP tool params, not just leaves. This PR adds no new CLI option, so I expect it to pass — but I have not run it against a rebased tree, so treat that as unverified rather than assured.

Rebasing onto `9fba962` before the fixes will save you doing the doc edits twice.

ffroliva:
### D2 (code quality) — the last dimension, now in

Verdict **🔴 RED**, which does not change the council's overall verdict but adds two findings and strengthens one.

**Independent confirmation of must-fix #1.** D2 reached the dead-listener bug at `migrated_composer.py:1037` by its own route and verified there is no other `media_id` assignment site. Two dimensions arriving at it separately, plus my own check of lines 974 / 1037 / 1261, makes that finding as solid as this council gets. It calls it *"the PR's headline safety mechanism, unreachable by construction"* — which is a fair description.

**New must-fix — `_unported_form:218` is now dead, and its message regressed.**

```python
if request.mode is Mode.R2V:
    ...            # every path returns
    return None
if request.mode is not Mode.I2V:   # <- unreachable
    return "this mode"
```

Verified: `Mode` is a 3-value `StrEnum` (`T2V`/`I2V`/`R2V`, `api/video.py:29`). With the T2V branch returning above and the new R2V branch always returning, control can only reach line 218 as `I2V` — so the branch can never fire.

It was reachable before this PR, and the diff also changed what it said:

```diff
-        return "reference-to-video"
+        return "this mode"
```

So the line went from "reachable, names the mode specifically" to "unreachable, generic". Either delete it, or keep it as a genuine fallback with the specific noun restored — as it stands it is strictly worse than what it replaced.

> This partly overlaps D14, which defended the `"this mode"` wording as non-speculative given there is no fourth mode. Both readings hold: D14 was judging the wording, D2 the reachability. D2's is the sharper point and supersedes it.

**Nice-to-have — two conventions for one job.** In `on_request` (lines 972-985) the r2v branch settles `route_error` with `.set_exception(WireFormatError(...))` while the i2v branch three lines below uses `.set_result(WireFormatError(...))` on the same `Future`, and the two use different `route=` formats (bare `rpcid` vs `f"batchexecute:{rpcid}"` — the latter is what existing tests assert on). Worth picking one before this hardens. Related: `_r2v_body_problem` (312) and `_i2v_body_problem` (285) are structurally near-identical and could share a helper, though not blocking at this size.

**Confirmed good:** no `has-text(...)` or locale-dependent selectors, no raw `print()`/`import logging` in `src/`, no duplication of the new mention/chip helpers elsewhere in `transports/`, new functions appropriately typed and sized.

---

**Council complete — 10 of 10 dimensions.** Consolidated code must-fixes:

1. `migrated_composer.py:1037` — arm the request listener for r2v (`or expect_reference_ids`), plus the mirrored cleanup at 1105.
2. `migrated_composer.py:99` — `MODEL_KEY` regex must include `r2v`.
3. `migrated_composer.py:218` — delete the dead branch or restore its specific message.

Everything else on the list is text: the stale "r2v exits 36" claims (now at `mcp/tools.py:1055-1056` after the rebase), the capture-script `--out` defaults, and the duplicate `### Added` in the CHANGELOG.

Offer stands — happy to push these three myself if you'd rather take just the doc edits.

ffroliva:
### Rebase is mechanically safe — verified on the merged tree, not just this branch

A parallel session closed the one thing I explicitly left unverified. Worth recording precisely, because it is a different claim from either green check on this PR:

- My D0 gate was **#683 alone** at `33d1b8b`.
- This PR's own CI green is from **00:17Z — about 11 hours stale**, and the branch is 21 commits behind; the three merges landed 11:03-11:29Z.
- Neither of those tested **#683 merged with current `develop`**. That has now been done: temp worktree at `33d1b8b`, `git merge origin/develop` (clean, no conflicts), then the full gate set on the *merged* tree:

| | |
|---|---|
| `tests/mcp/test_cli_parity.py` | **7 passed**, incl. `test_every_cli_option_reaches_its_mcp_tool` |
| ruff / format / doc links / hygiene / website mirror | green |
| `pyright src` | 0 errors |
| full offline suite | **4006 passed**, 24 skipped |

So the new option-level parity gate is a **non-event** for this PR — it adds no unmirrored CLI option. Rebase onto `9fba962` with confidence.

---

### The pattern worth naming: three instances of one shape, all today

Credit to that session for spotting this. Your bug is not an isolated slip — it is the third occurrence today of a failure mode this codebase is structurally blind to:

| | Both halves correct | Wiring wrong |
|---|---|---|
| **#683 (this PR)** | `_r2v_body_problem` is written and unit-tested | the listener is gated on `expect_media_id`, so it never registers on the r2v path |
| **#689** (fixed in `ea55f1b`) | `worker/codec.py` *read* `reference_entities` off the payload | no tool ever *wrote* it |
| **MCP r2v guard** (also `ea55f1b`) | `GenerateVideoRequest` accepts entities | the guard fronting it demanded `reference_images` — stricter than the rule it guards |

**All three pass every gate**, because in each case both halves are individually correct and only the join between them is wrong. A unit test on either half stays green. Type checking stays green. Lint stays green. That is precisely why your PR arrived with 26 passing tests and a clean CI.

What catches this class is a **round-trip assertion**: drive the real entry point and assert the far end observed it. Concretely for must-fix #1 — the regression test should drive an r2v `submit_and_observe` with a bad body and assert `_r2v_body_problem` **actually fired**, rather than calling `_r2v_body_problem` directly. The current test proves the function is correct; it cannot prove the function runs.

That is the test worth adding alongside the one-line fix, and it is the one that would have caught this before review.

ffroliva:
### All six must-fixes verified, and I resolved the merge conflict for you

Re-reviewed at `128234b` and checked each item against the code rather than the commit subjects:

| Must-fix | Status |
|---|---|
| Listener armed for r2v | ✅ `or expect_reference_ids` at 1155 + 1223 |
| `MODEL_KEY` matches r2v keys | ✅ |
| Dead `_unported_form` branch | ✅ removed |
| Stale "r2v exits 36" claims | ✅ `mcp/tools.py`, `docs/MCP.md`, `skills/`, `USAGE`, `CONFIGURATION`, `INDEX`, `KNOWN_ISSUES` all corrected |
| Capture-script `--out` defaults | ✅ `default=None` → the gitignored spike dir |
| Duplicate `### Added` | ✅ resolved |

**The round-trip test is the part worth calling out.** `test_submit_arms_the_body_assertion_on_the_r2v_path` drives the real `submit_and_observe` with a body that dropped a reference and asserts `_r2v_body_problem` actually **fired** — rather than calling the helper directly. That is the test that catches the whole bug class, not just the one instance. Exactly right, and better than a narrower fix would have been.

I also appreciate the CHANGELOG keeping the correction visible ("the submit-body assertion was written and unit-tested but never registered") instead of quietly rewriting history. That is the more useful record.

### The conflict was mine, so I fixed it rather than bouncing it back

`develop` moved under you again when **my** #694 merged at 13:47Z, and both conflicting files — `CHANGELOG.md` and `tests/e2e/test_migrated_host_e2e.py` — were ones I had touched. Since `maintainerCanModify` is on, I merged `develop` in and resolved it: one `### Added` carrying your r2v entry plus my diagnostic entry, my `### Fixed` above it, and the e2e bail assertion left accepting **either** site (`mint_recaptcha_token` or `mint_recaptcha_token_after_failure`) so neither of our changes rejects the other's path.

Pushed to your branch as `b518003`. Nothing of yours was altered — it is a merge commit only.

### Verified on the merged tree

| | |
|---|---|
| full offline suite | **4023 passed**, 24 skipped |
| `ruff` / `ruff format --check` | clean |
| `uv run pyright src` | 0 errors |
| doc links / repo hygiene / website mirror | green |

Merging once CI confirms. Thanks for the thorough turnaround on this — six items including a genuinely subtle one, plus the right test for it.

PR #710, opened 2026-09-06
release: v0.70.0

ffroliva:
## v0.70.0

**`gflow character create` works on the migrated `flow.google.com` host.** It was never broken there — the driver wasn't driving.

### The headline

The character editor is fully present on `flow.google.com`, on the **same** labs tRPC + aisandbox backend. Only the view layer differs — labs is React + **Slate**, migrated is Angular + **ProseMirror** — so seven selectors missed, a 20 s readiness gate timed out, and the absence of a *match* was recorded as the absence of the *feature*. v0.69.0 then hardened that misreading into a guard that aborted **before** probing the DOM, which made the claim unfalsifiable.

Verified live on a moved account, confirmed by independent read-back: a complete character — name, personality, voice, **portrait and body** — with two distinct workflow ids, two distinct media ids and both images on disk.

### Also in this release

- **`--model` is deterministic.** The picker was best-effort and could silently generate on a tier you did not ask for; it now verifies by re-reading the chip and raises rather than proceeding. Refusing is free — it runs before submit.
- **A failed `character create` no longer strands an "Untitled Character."** A/B controlled live: 1→2 entities without the fix, 2→2 with it. The rollback asks the backend first, because an empty local workflow list means only that *this process* failed to read a result.
- **Each slot takes its own `primaryMediaId`**, read from the project listing — reusing the thumbnail corrupted the body workflow through `commit_workflow`.
- **`-o <existing directory>` no longer costs you a clip** — exits 2 in 0.8 s instead of exit 1 after ~2 min with a billed, orphaned mp4.
- **Spike becomes Phase 0** of the workflow (`/gflow:spike`), with the CDP/HAR harness vendored in and CONTRIBUTING telling contributors how to produce redacted evidence.

### Evidence

`docs/LIVE_VERIFICATION_v0.70.0.md`, which also records what was **not** verified: the labs.google path for all seven changed anchors, because every account available here is migrated.

### Test plan

- [x] Full Impeccable Routine on develop: **4056 passed, 7 skipped, 92% coverage**, ruff clean, pyright 0 errors, all six doc/hygiene gates
- [x] Version bumped in `pyproject.toml`, `.codex-plugin/plugin.json`, `__init__.py`
- [x] CHANGELOG `[Unreleased]` migrated to `[0.70.0]` with its compare link
- [x] 84 version/changelog/release gate tests pass
- [x] Live-verified on `ci-probe` — see the ledger


ffroliva:
Re-opening from `chore/release-v0.70.0` — the base guard requires that prefix, not `release/`. Same commit, plus the two release-artifact gates fixed (PROJECT_STATUS for v0.70.0, and the ledger's Not-verified entries now name a blocker and a tracking issue).

PR #725, opened 2026-09-07
fix: five defects a live e2e sweep exposed, plus the credits misdiagnosis

ffroliva:
## Summary

A full `pytest -m "e2e and not e2e_video"` sweep on a **funded, migrated** account (`denon82`, 291 credits) returned **27 failed / 32 passed / 12 skipped**. 22 failures were migration-structural — the world moved to `flow.google.com`, the tests did not. This PR fixes the rest, plus a credits misdiagnosis and a log-label leak found alongside them.

Every fix was **watched failing first**. Where a fix touches a Flow surface it was **re-run against live Flow** and the result is quoted below.

### 1. An empty wallet was diagnosed as a moved frontend (`48f870b`)

On the migrated host an account out of Veo credits produced `UiSelectorDriftError` (exit 23) — *"Google may have updated their frontend… file a bug"*. Flow does not **disable** the submit control when the wallet is empty, it **replaces** it: `arrow_forward` disappears and a `prompt-warning-button` carrying `aria-label='Insufficient credits warning'` takes its place. The anchor's absence tracks the **wallet**, not the frontend.

Measured by A/B (`scripts/dev/spike_migrated_submit_anchor.py`, from #717) — same probe, same host, same code, ~60 s apart; drained account had `arrow_forward` **absent** and the warning **present**, funded account the mirror image.

New `InsufficientCreditsError`, **exit 37**. Every path that gives up on the submit control now checks the wallet before naming a culprit. A missing anchor with **no** warning beside it is still 23 — the guard must not swallow real drift, and a test pins that.

Beyond the wrong message, the old behaviour manufactured frontend-drift bug reports that no code change could ever fix.

### 2–5. Four defects the sweep exposed, and a log-label leak (`abb6920`)

- **`e2e_env` handed its subprocess the isolated home.** `_isolate_settings` is autouse and points `GFLOW_CLI_HOME` at an empty tmp dir; `os.environ.copy()` inherited it, so the child could not see *any* real profile and exited 2 with *"No session for profile"* while that profile was in active use by the rest of the run. Six e2e modules each carried a private copy of the workaround; it now lives in the shared fixture, once.
- **Four agentic-count tests timed out blind.** They drive raw Playwright at a labs-only panel with no migration guard, so on a moved account each spent ~40 s failing on a locator with nothing saying the *host* was the reason. Two guards were needed, not one: the handoff **cancels** the navigation, so the `goto` itself can die with `net::ERR_ABORTED` before any post-goto check runs.
- **The character "double-spend" was a test asserting an unimplemented guarantee.** It ran two ordinary, fully-completing creates and asserted the second reused the first's entity — but a completed run records `SUCCEEDED` and `find_incomplete_character` only matches `status='started'`, so the reuse branch was unreachable and a second entity was minted every time, on every host. Written **2026-07-06** — red by construction for two months, unseen because `e2e_character` is opt-in and the nightly canary runs only `e2e_auth`. Its docstring claimed a mid-saga kill *"cannot be reliably injected from a black-box subprocess test"*; it can — the test already holds a DB handle, so flipping the row back to `'started'` exercises the resume path against true state. Same-name idempotency was never promised and is **not** implemented here.
- **The MCP ambiguous-name failure was the tool working correctly.** The fixture project held two characters sharing a display name — pollution from the defect above — so `show(name=…)` refused. The test now picks a name unique in the project.
- **`run_with_handlers` bound `cli_command` and never unbound it.** Only `cli.main()` clears contextvars, and long-lived processes (MCP server, `gflow serve`, the worker, a pytest session) run many commands without passing back through it. Six captures in the sweep log read `cli_command="project rename"` while their request bodies show `project.createProject`. A wrong label is worse than a missing one — it sends whoever is debugging to the wrong command.

## Test plan

- [x] TDD throughout — each fix watched **red** first, with the failure message recorded
- [x] Anti-over-reach guards included: real drift still reports 23; `cli_command` still bound *during* the run
- [x] Offline suite: **3998 passed, 7 skipped, exit 0**
- [x] `ruff check` / `ruff format` / `pyright src` clean (453 formatted, 0 errors)
- [x] repo hygiene, doc links, website PII, website mirror, council memory — all exit 0
- [x] **E2E re-run on the affected surface** (`denon82`, funded + migrated):
  - `test_e2e_env_isolation.py` — 2 passed
  - `test_agentic_count_enforcement_e2e.py` — **4 skipped in 44 s** (was ~160 s of blind timeouts)
  - `test_character_mcp_parity_e2e.py` — 3 passed, $0
  - `test_migrated_i2v_e2e.py` — 2 passed

## Named blockers — not hidden, not claimed

- **The exit-37 e2e test has not been run.** It is committed and gated on its own `GFLOW_CLI_E2E_DRAINED_PROFILE`, deliberately *not* the usual e2e profile — that one is funded and would make the test pass by never reaching the branch. `denon82` has 291 credits and I will not drain an account to test this. A drained profile exists elsewhere and the run is handed to whoever holds it.
- **`--format-prompt` selector drift is confirmed and left red.** Reproduced live: `ui_automation.format_button_not_found` fires while the whole create succeeds. The CLI help states the contract plainly — *"Best-effort: if the button is not found the prompt is submitted as typed"* — so the code is correct and does warn on stderr. What is broken is the `personal_recommendations` cascade no longer matching Flow's Format button. Fixing it needs a live DOM probe; guessing a new cascade from a log line is how the two-month-dead saga test above happened. The test is a drift canary and it fired correctly, so it stays red until the probe runs.
- **The migrated i2v upload failure did not reproduce** (2 passed, 85 s idle). `FRAME_PICKER_OPEN_S` is **not** widened — that would be a speculative fix for a failure I could not observe.

## Note on the "flaky" suite

Two single-test failures earlier today, different tests each time, both while a 34-minute e2e sweep and a browser fleet competed for the machine. Three consecutive full runs on an idle box: **3994 / 3994 / 3994 passed**. Contention, not flakiness.

## Coordination

Overlaps #717 on `migrated_composer.py` (different regions, auto-merges) and `CHANGELOG.md` (**will conflict** — #717 should land first and I will resolve).


ffroliva:
## Council review — findings and what was done about them

A full council ran on `abb6920`: **D1 GREEN · D3 GREEN · D14 GREEN · D15 GREEN · D4 YELLOW · D6 YELLOW**. Both YELLOWs were mine and both are now closed by a run rather than a caveat. Commit `6735fee` is the result.

### D4 — the saga rewrite had never been run, and running it proved it wrong

The reviewer noticed something the Test plan hid: `test_character_create_partial_saga_recoverable` was rewritten from code analysis, listed in neither the live re-run evidence nor the named blockers, and gated behind an opt-in the offline suite never reaches. It could have been run the whole time — so it was an omission with a false paper trail attached, which is exactly what the Iron Law singles out.

Run live, it **failed**:

```
AssertionError: the interrupted saga was not resumed: the re-run minted a NEW
entity and re-spent the face quota. '7f8d31c6-…' != '0d261bc5-…'
```

The instinct was to file a product defect in the resume path. That would have been wrong. **The injection was wrong, not the product.** `record_character_completed` REPLACES `metadata_json` with a freshly built `{workflow_ids, primary_media_ids, …}`, dropping the `{entity_id, name}` that `record_character_started` wrote — so a *completed* row flipped to `'started'` carries no `$.name`, and `find_incomplete_character` matches on exactly that key. A real crash never leaves that shape: it dies before completion with both fields intact. **Status alone is not the crash; the metadata is.**

Corrected, then re-run live on `denon82`:

| run | injected state | result | wall-clock |
|---|---|---|---|
| first | status only | **FAILED** — fresh entity minted | 7 m 09 s |
| corrected | status **+** metadata | **1 passed** | **3 m 44 s** |

The clock corroborates the assertion independently: roughly half the time, because the resume skipped face regeneration instead of generating twice.

### D6 — the shipped selector had never touched Playwright's real engine

`CREDITS_WARNING` was exercised only against a hand-built `FakeLocator` that matches on **string equality**, and the A/B that measured the warning button used raw `document.querySelectorAll`. Neither calls `page.locator(CREDITS_WARNING)` — the API the shipped code actually uses. Two verified halves, unverified join.

Probed live against a real migrated project page:

```
selector under test: button.prompt-warning-button, [aria-label*='Insufficient credits']
landed on: https://flow.google.com/project/7702600d-…
page.locator(CREDITS_WARNING).count() = 0
  ok  'button.prompt-warning-button' -> 0
  ok  "[aria-label*='Insufficient credits']" -> 0
control: arrow_forward submit buttons = 1
```

This rules out the failure mode that no test could have caught: a **malformed selector** would make the guard silently never fire, and every unit test would still pass because `FakeLocator` never consults a real engine. The control resolving (`= 1`) means a flat zero cannot be mistaken for "the engine works."

**It does not prove the positive match.** That still needs a drained wallet and remains a named blocker below.

### Cleared without change

- **D14** enumerated every reuse candidate for the new class. `RateLimitError` was the tempting one and is **retryable** — reusing it would have made the CLI, the MCP server and the worker all retry an empty wallet forever. `InsufficientCreditsError` is correctly absent from `RETRYABLE_ERRORS`.
- **D15** confirmed an MCP caller gets the same problem-type URI, class name and remediation text as a CLI caller with **no MCP change needed** — the error funnel is generic over `GFlowError`. `docs/MCP.md` needed no edit (its only enumeration is the retryable whitelist plus a terminal catch-all).
- **D3** traced `GFLOW_CLI_HOME` end-to-end: popping it opens no catalog or output risk, and the raw `UPDATE operations SET status='started'` can only reach the isolated tmp DB.
- **D1** verified the red/green split empirically — it reverted the production code and re-ran the six new unit tests: 4 failed as expected, and the 2 that passed either way are the anti-over-reach guards.

### Recorded, pre-existing, not fixed here

`incidents/` has no env-var override the way `GFLOW_CLI_DB_PATH` and `GFLOW_CLI_OUTPUT_DIR` do, so an `e2e_env`-driven subprocess that hits an operational failure writes incident bundles into the developer's **real** `<home>/incidents/`. This predates the PR and is shared by every module using the same pattern. Noting it so it stops being invisible.

## Named blockers — unchanged, still honest

- **The exit-37 e2e has not been run.** It needs a drained account; `denon82` has 291 credits and draining one to test this is refused. Handed to whoever holds a drained profile.
- **The selector's positive match is unproven.** The settling run is one line — `page.locator(migrated_composer.CREDITS_WARNING).count()` on a drained account, $0, no submit.
- **No spike finding doc** exists in `docs/superpowers/spikes/` for the credits A/B. The measurement belongs to the session that took it, so the doc should carry their data rather than my retelling.

## Final verification on `6735fee`

- offline suite: **3998 passed, 7 skipped, exit 0**
- `ruff check` clean · `ruff format` **453 files formatted** (one file was caught un-formatted by the final gate and fixed — cosmetic only, no semantic change to the live-verified test)
- `pyright src`: 0 errors · doc links resolved · repo hygiene 1008 files clean · website mirror in sync

Issue #727, opened 2026-09-07
character create --format-prompt silently skips Flow's Format button — the personal_recommendations anchor has drifted


ffroliva:
## What happens

`gflow character create --format-prompt` no longer clicks Flow's Format button. The prompt is submitted as typed, the command exits **0**, and the only sign is a warning line on stderr.

Reproduced live on 2026-09-07 against `denon82` (migrated host, funded, idle machine). The create itself succeeded end to end — `ui_automation.character_model_selected`, `ui_automation.prompt_input_found`, `ui_automation.prompt_submitted`, `character_create.entity_patched`, `character_create.face_done` all fired. The one thing missing:

```
ui_automation.format_button_not_found
```

and `ui_automation.prompt_formatted` never appears.

## This is not a code bug — the selector has drifted

The contract is explicit and the code honours it. `src/gflow_cli/cli_character.py`, `--format-prompt` help:

> Click Flow's in-editor Format button before submitting… **Best-effort: if the button is not found the prompt is submitted as typed.**

`format_character_prompt` returns `False` and logs `ui_automation.format_button_not_found` (`src/gflow_cli/api/transports/ui_automation.py:1656`); both callers (`:1697`, `:1765`) proceed to submit, which is what "best-effort" means. The warning does reach stderr, so this is degraded-and-noisy rather than silent.

What is actually broken is the anchor. `PROMPT_FORMAT_SELECTORS` (`ui_automation.py:264-268`):

```python
"button:has(i.google-symbols:text-is('personal_recommendations'))",
"button:has(i:text-is('personal_recommendations'))",
"button:has(span:text-is('Format'))",
```

None of the three match any more. Note the third is a display-label fallback and is banned as a primary anchor by the locale-invariance rule in AGENTS.md, so the fix cannot simply lean on it.

## Impact

Anyone passing `--format-prompt` silently gets an unformatted prompt. Character creation consumes image quota, so the run is paid for and the requested prompt-engineering step never happened. The feature has been inert for an unknown period — `e2e_character` is opt-in (`GFLOW_CLI_E2E_RUN_CHARACTER=1`) and the nightly canary runs only `e2e_auth`, so nothing was watching.

## Test status: deliberately RED

[...]

Issue #691, opened 2026-09-06
mcp: port gflow character create — the exemption reason was never a reason


ffroliva:
`_MCP_EXEMPT` carried `"character create": "character mutations — not yet ported"`. "Mutations" is a category, not a justification, and it does not survive contact with the rest of the table.

## Why the stated reason doesn't hold

The implied argument is that a mutation which spends quota needs human consent. But `gflow_generate_image` is exposed over MCP and spends the same per-model daily quota. There is no principle that admits one and refuses the other — creating a character generates reference images, which is the thing the tool already does.

Compare the exemptions that **do** carry real reasons:

- `video extend`: long-running billed chain, not worker-enqueued, would block a client's tool call past its timeout, and its cost confirmation has no MCP equivalent (#481). Three independent reasons.
- `auth login`: genuinely interactive browser flow.
- `data prune`: irreversible destruction of local data.

`character create` has none of that shape. It is simply unported.

## Scope

`services/character_create.py::character_create` already orchestrates the persist-before-spend saga, so the MCP tool is an adapter over it, not new logic. CLI surface to mirror: `--project`, `--name`, `--face-prompt`, the body-image options, `--voice`, `--personality`, and the prompt-formatting flag.

Two things to settle in the port rather than assume:

1. **Duration.** It generates one or two images, so it is closer to `gflow_generate_image`'s latency than to a catalog read. Decide whether it enqueues through the worker like the generate tools, or runs inline — #481 is the relevant hazard.
2. **Consent.** If it should not spend without a human, say so explicitly and gate it behind `--no-spend` semantics like the generate tools already honour (`GFLOW_MCP_NO_SPEND`), rather than leaving it unreachable.

Blocking neither #687 nor anything else — filed so the exemption stops being a shrug.

Refs #689.

Issue #481, opened 2026-08-12
investigate: clamp long MCP operations to the client tool-timeout (or confirm enqueue already covers it)


ffroliva:
linkedin-mcp expresses inner waits as a fraction of the client tool-timeout, because MCP clients time out around 60s while some operations run longer.

Current state (gflow): only static timeouts/credit budgets (config.py:483-484); nothing reads the client-supplied tool timeout. BUT gflow MCP tools appear to enqueue to a worker rather than block, which may already avoid the long-blocking-call problem.

Investigation: confirm whether any MCP tool actually blocks longer than a typical client timeout. If yes, express those waits relative to the client timeout or return a poll handle early. If no, close as not-needed. Size: medium; needs a look first. Tier 3.

ffroliva:
**Investigated against the v0.57.0 tree. The premise is inverted — the tools do _not_ already avoid this, so it should not be closed as not-needed.**

## Do any MCP tools block longer than a typical client timeout? Yes — both generate tools, by default.

`gflow_generate_image` and `gflow_generate_video` both declare `wait: bool = True` (`mcp/tools.py:736`, `:941`). With `wait=True`, `_run_generation_task` (`:300-420`) does **not** hand off to the daemon. It:

1. enqueues the task,
2. **claims it and runs it in-process** — `await worker.process_task(claimed)` (`:363-369`), a full browser generation,
3. reads the terminal state back and returns.

So the call blocks for the entire generation. The governing budget is `timeout_seconds: int = Field(default=600, ge=1, le=3600)` (`config.py:483`) — a **10-minute default, 60-minute ceiling**, against MCP clients that typically give up near 60 s.

The "tools enqueue rather than block" assumption in the issue body describes the `wait=False` branch (`:308-313`), which is not the default.

## The non-blocking path already exists — but it is opt-in and undiscoverable

`wait=False` returns a task handle immediately via `_enqueue_generation_task`, and the SEP-2663 Tasks extension (#409, v0.50.0) serves `tasks/get` + `tasks/cancel` (`mcp/tasks_extension.py:38-48`) so a client can poll it to completion. That is exactly the "return a poll handle early" design this issue asks for, and it is already built.

[...]

PR #729, opened 2026-09-07
fix(character): re-anchor the Format button on its custom element, not a tag that moved

ffroliva:
## Summary

> ### Both faults are fixed and live-verified
>
> `--format-prompt` had **two** independent faults. The anchor had drifted (the button was never found), and — once found — gflow clicked it and submitted **before Flow's server-side rewrite arrived**, so the reshaped prompt was discarded on every run behind a green exit code.
>
> Both are fixed here. Live e2e: **`1 passed in 406.20s`** on `denon82`, asserting the **observed rewrite** (`prompt_formatted`, now emitted only after the composer text actually changes) plus `format_button_clicked`, and rejecting `format_not_observed`, `format_click_failed` and `format_button_not_found`.
>
> Measured cost of making it work: **406s vs a 210s control** on the same account and face prompt. Partly the ~4-5s rewrite; mostly Flow generating from a 608-character elaboration instead of the 102 characters typed. That is the feature's actual price — it was free only while it did nothing — and it is why `_CREATE_TIMEOUT_S` moved 300 → 480.

## Lifecycle

- [x] Issue triaged — reproduces on **CLI only**; `character create` is an explicit `_MCP_EXEMPT` in `tests/mcp/test_cli_parity.py`, so there is no MCP twin to drift
- [x] `spike` run **before** any claim — `skills/spike/SKILL.md`, finding in `docs/superpowers/spikes/2026-09-07-character-format-button-anchor.md`
- [ ] ~~`predict`~~ — a selector re-anchor backed by a live DOM capture, not a transport/auth/schema change
- [ ] ~~`scenario` + `plan`~~ — single-constant fix with an existing drift canary already in place
- [x] `check` green, including the step 1b mirror sweep (all six axes below)
- [x] MCP twin: **reasoned exemption already recorded** for `character create`; zero hits for `format_prompt` under `src/gflow_cli/mcp/`
- [x] **E2E evidence** — `tests/e2e/test_character_create_e2e.py::test_character_create_format_prompt_clicks_format_button`, **`1 passed in 219.90s`** on `denon82`, real generation, zero credits. This test was RED before the fix and is green **by a live run** — the assertion was never relaxed.
- [x] Live-verified against real Flow; what could NOT be verified is stated below
- [ ] Council review — running next

## Validation

- [x] Two new unit tests pin the contract that was missing, both **watched red** against the old constant first:
  - `test_cascade_covers_both_ligature_carriers` — `AssertionError: cascade must carry the migrated-host <mat-icon> carrier`
  - `test_cascade_carries_no_display_label_anchor` — `AssertionError: "button:has(span:text-is('"'"'Format'"'"'))" anchors on a localised display label`
  - plus `test_clicks_ligature_selector_first`, updated to the new primary anchor
- [x] `uv run python scripts/ci/check_repo_hygiene.py` — 1017 files, no violations
- [x] `uv run python scripts/ci/check_doc_links.py` — all links resolved across 29 files
- [x] `uv run python scripts/ci/check_website_docs_pii.py` — no private identifiers, 25 files
- [x] `uv run python scripts/ci/generate_website_docs.py --check` — mirror in sync
- [x] `uv run python scripts/ci/check_council_memory.py` — 48 memory files, all cited
- [x] `uv run ruff check src tests` / `ruff format --check src tests`
- [x] `uv run pyright src` — 0 errors
- [x] `uv run python -m pytest -q --cov=gflow_cli` — **4009 passed, 7 skipped, 92% coverage**

### Step 1b blast radius — six axes

Grepped `format-prompt` / `format_prompt` / `personal_recommendations` across `src/gflow_cli/mcp/`, `docs/`, `website/docs/`, `README.md`, `KNOWN_ISSUES.md`, `skills/`.

| Axis | Verdict |
|---|---|
| A. Execution paths (CLI / MCP direct / MCP queued) | **N/A** — no option, param or payload key changed; a selector constant only |
| B. MCP surface truth | **N/A** — 0 hits in `src/gflow_cli/mcp/`; `character create` exempt |
| C. Agent-facing surfaces | **N/A** — 0 hits in `skills/`, `README.md`, `docs/INDEX.md`, AGENTS.md command surface |
| D. Error / exit-code surface | **N/A** — behaviour stays best-effort; no raise site, no `EXIT_CODE_MAP` entry |
| E. Declarative / template surfaces | **N/A** |
| F. Docs + published mirror | CHANGELOG entry added; mirror verified in sync. Remaining hits are dated historical records (`LIVE_VERIFICATION_v0.45.0.md`, `PROJECT_STATUS.md`, the 2026-07-26 PLAN) that correctly describe what was true then |

## What this does NOT prove

- **The labs frontend.** No unmoved account exists here, so the two `<i>` entries are carried forward on the 2026-07-27 capture alone. They were **not** re-verified today and are not asserted to still work.
- **The content of Flow'"'"'s rewrite.** The e2e asserts the button was found, enabled and clicked. Whether Flow'"'"'s reshaped prompt is *better* stays a human read.
- **That this was the last missed constant.** #703'"'"'s carrier sweep demonstrably skipped one; this PR sweeps one more.

## Open question, deliberately not acted on here

`--format-prompt` stays **best-effort**, which is what its help text promises and what #727 says is correct. But that is *why* this went unnoticed for an unknown period: a paid run silently skipped the step the user asked for behind exit 0, and only an opt-in e2e (`GFLOW_CLI_E2E_RUN_CHARACTER=1`, not in the nightly canary) could catch it. Whether a missed button should be loud — or fatal under an explicit `--format-prompt` — is a real design change and belongs in its own issue.

## Contribution Checklist

- [x] Targets `develop`
- [x] Real Git identity
- [x] No secrets, cookies, tokens, signed URLs or raw captures — the capture stays gitignored under `scripts/dev/_spike_out/`
- [x] Reviewed before submitting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,15 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   degraded to a no-op: `ui_automation.format_button_not_found`, exit **0**, prompt submitted as
   typed, and the image quota spent on a run whose requested prompt-engineering step never
   happened. The button was never absent. Measured on 2026-09-07 with
-  `scripts/dev/spike_character_prompt_format.py` (`$0` — DOM read only, with a control anchor
-  so a flat zero could not be mistaken for absence): it renders as
+  `scripts/dev/spike_character_prompt_format.py` (`$0` — navigation, one free `createEntity`,
+  DOM reads and typing; two control anchors, so a flat zero could not be mistaken for
+  absence): it renders as
   `<flow-format-prompt-button>`, and the Angular frontend carries the unchanged
   `personal_recommendations` ligature in **`<mat-icon>`** rather than `<i>` — the same carrier
   split fixed for `add_2` / `arrow_drop_down` / `accessibility_new` in #703, which this constant
   was not swept with. The EN `span:text-is('Format')` fallback missed too and is **deleted**
   rather than translated: the button's own `aria-label` reads `"Formatar"` on a pt account, and
   display labels are banned as anchors. The cascade now leads with the custom element and covers
-  both carriers. ([#727](https://github.com/ffroliva/gflow-cli/issues/727))
+  both carriers. ([#727](https://github.com/ffroliva/gflow-cli/issues/727),
+  [spike](docs/superpowers/spikes/2026-09-07-character-format-button-anchor.md))
 
 - **`video t2v --reference-entity` no longer bills a clip that ignores the entity.** On the
   migrated `flow.google.com` host the "character references are not ported" refusal lived inside
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`character create --format-prompt` clicks Flow's Format button again.** On the migrated
+  `flow.google.com` host all three entries of `PROMPT_FORMAT_SELECTORS` missed, so the flag
+  degraded to a no-op: `ui_automation.format_button_not_found`, exit **0**, prompt submitted as
+  typed, and the image quota spent on a run whose requested prompt-engineering step never
+  happened. The button was never absent. Measured on 2026-09-07 with
+  `scripts/dev/spike_character_prompt_format.py` (`$0` — DOM read only, with a control anchor
+  so a flat zero could not be mistaken for absence): it renders as
+  `<flow-format-prompt-button>`, and the Angular frontend carries the unchanged
+  `personal_recommendations` ligature in **`<mat-icon>`** rather than `<i>` — the same carrier
+  split fixed for `add_2` / `arrow_drop_down` / `accessibility_new` in #703, which this constant
+  was not swept with. The EN `span:text-is('Format')` fallback missed too and is **deleted**
+  rather than translated: the button's own `aria-label` reads `"Formatar"` on a pt account, and
+  display labels are banned as anchors. The cascade now leads with the custom element and covers
+  both carriers. ([#727](https://github.com/ffroliva/gflow-cli/issues/727))
+
 - **`video t2v --reference-entity` no longer bills a clip that ignores the entity.** On the
   migrated `flow.google.com` host the "character references are not ported" refusal lived inside
   the r2v branch of the routing gate, so a t2v request returned from that gate before its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **`character create --format-prompt` clicks Flow's Format button again.** On the migrated
-  `flow.google.com` host all three entries of `PROMPT_FORMAT_SELECTORS` missed, so the flag
-  degraded to a no-op: `ui_automation.format_button_not_found`, exit **0**, prompt submitted as
+- **`character create --format-prompt` finds and clicks Flow's Format button again.** Scoped
+  deliberately: this fixes the *anchor*, and **`--format-prompt` still does not deliver a
+  reshaped prompt** — see the Known-issues note below. On the migrated `flow.google.com` host
+  all three entries of `PROMPT_FORMAT_SELECTORS` missed, so the flag degraded to a no-op:
+  `ui_automation.format_button_not_found`, exit **0**, prompt submitted as
   typed, and the image quota spent on a run whose requested prompt-engineering step never
   happened. The button was never absent. Measured on 2026-09-07 with
   `scripts/dev/spike_character_prompt_format.py` (`$0` — navigation, one free `createEntity`,
@@ -86,6 +88,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a dead cookie.
 - `scripts/dev/spike_migrated_composer_mode_axis.py` — hard-error probe that opens the migrated
   composer's settings overlay and enumerates its radiogroups.
+- `scripts/dev/spike_format_prompt_effect.py` — measures whether clicking Format actually
+  rewrites the prompt, and when. `$0`, never submits.
+
+### Known issues
+
+- **`character create --format-prompt` still submits the prompt as typed, even when the button
+  is found and clicked.** The anchor fix above is necessary and not sufficient. Flow rewrites
+  **server-side** — a `batchexecute` round trip whose result reaches the composer at **~5.4s**
+  (measured, `denon82`, `$0`) — while `format_character_prompt` waits `_jitter_ms(500)` and
+  `_send_prompt` submits on the next line. The reshaped text arrives after the submit has
+  already gone.
+
+  **The defect is not that the wait is too short. It is that gflow reports success without
+  observing it**: `ui_automation.prompt_formatted` is emitted on the *click*, so the absence of
+  a completion inside a window we chose is recorded as a completion. That is the same class as a
+  20 s selector timeout read as "the feature is absent" and a bundle photographed after teardown
+  read as "the DOM was gone" — a measurement that could not be taken, recorded as a measurement.
+  This one is the worst of them because it fails toward **success**: a premature green looks
+  exactly like the feature working, and nobody investigates a pass. The e2e is green for this
+  reason and is not wrong to be — it asserts the click, which was the only observable when it
+  was written.
+
+  A `/gflow:predict` on the fix returned **GO with conditions** (8.2/10): poll the composer
+  until its text differs from the string gflow inserted, guarded by a length delta rather than
+  raw inequality, and emit `prompt_formatted` on the observed rewrite. Not gating on the
+  `batchexecute` response — it arrives **4.2s before** the DOM settles, so a wire-level gate
+  reproduces the same early submit.
+  ([#727](https://github.com/ffroliva/gflow-cli/issues/727),
+  [spike](docs/superpowers/spikes/2026-09-07-format-click-is-not-a-format.md))
 
 ## [0.70.0] — 2026-09-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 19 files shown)


ffroliva:
## ⚠️ Correction to this PR's claims — the fix is necessary but NOT sufficient

A Phase-0 spike run after the council (`scripts/dev/spike_format_prompt_effect.py`, `$0`, denon82) measured what happens **after** the click this PR fixed:

```
baseline        'girl, red hair, sad'                       19 chars
format_clicked  flow-format-prompt-button button
box_text  t=0.011   chars=19    'girl, red hair, sad'      <- unchanged
box_text  t=5.355   chars=651   'Medium studio shot of a young girl with…'
verdict   changed=True  changed_at_s=5.355  within_current_wait=False
          requests_after_click=1
t=1.191   POST .../batchexecute?rpcids=eAenfb    <- the rewrite, server-side
```

**Flow's rewrite lands at ~5.4s. `format_character_prompt` waits `_jitter_ms(500)` and returns, and `_send_prompt` submits on the next line.**

So `--format-prompt` still submits the prompt the user typed, discarding the reshaped one — on every run, *including* runs where the button is found, enabled, clicked, and `ui_automation.prompt_formatted` is emitted.

### What that means for the green e2e in this PR

`test_character_create_format_prompt_clicks_format_button` asserts `prompt_formatted` present and the two failure events absent. All three genuinely hold — the run was real and I am not withdrawing it. But the assertion set proves **the click**, not **the rewrite**, so the test is green while the feature does not work. It asserts the click because at the time it was written the click was the only observable.

**This PR's scope is unchanged and still correct**: #727 is "the anchor has drifted", and the anchor is fixed, verified live three times. What I am correcting is the *implication* a reader would draw from "e2e green" — that `--format-prompt` now does something. It does not, for a second and independent reason.

### Why the fix is not folded in here

The settle signal is available (the box text changes in one discrete step, so it is a condition to poll, not a duration to sleep), and anchoring on "the text differs from what we inserted" is locale-invariant by construction. But choosing a timeout needs a distribution and I have **one** observation; the failure modes of `eAenfb` were never provoked; and labs was not measured at all. That is a behaviour change on a paid path, so it goes through `/gflow:predict` first.

Evidence: `docs/superpowers/spikes/2026-09-07-format-click-is-not-a-format.md` (added in this PR).

ffroliva:
# `/gflow:predict` — wait for the observed rewrite in `format_character_prompt`

**Verdict: 🟢 GO with conditions** · confidence **8.2 / 10** · no STOP conditions.

| Persona | Score | Verdict | Headline |
|---|---|---|---|
| P1 Architect | 8 | GO | Fits the existing settle-gate idiom; invent nothing |
| P2 Security | 7 | GO* | No auth surface; the fingerprint moves *toward* human — but do not log the rewritten text |
| P3 Performance | 6 | CAUTION | Design is right; the timeout must not be reverse-engineered from n=1 |
| P4 CLI/MCP UX | 8 | GO | No MCP surface to break; hazards are doc/telemetry edits, not design risk |
| P5 Devil's Advocate | 7 | CAUTION | Withdrew its own duplication objection; wants #731 landed first |

Average 7.2, +1 for unanimous agreement on the approach. The two CAUTIONs are conditions on the *number* and the *ordering*, not objections to the design.

## The one measurement that decides the mechanism

> The `eAenfb` response arrived at **t=1.191s**. The DOM settled at **t=5.355s**.

P1 and P3 independently rejected gating on `page.expect_response(rpcids=eAenfb)`, and P3 gave the killer reason: **the response precedes the DOM by 4.2s, so a wire-level gate re-creates the same early submit.** It also hardcodes an undocumented `batchexecute` id into `ui_automation`, against the structural-anchor rule, on a host whose labs twin was never measured. Keep the rpcid as a logged diagnostic; never as the gate.

## Conditions carried into the plan

**Unanimous (3+ personas independently):**
1. **Length delta, not `!=`.** Slate/ProseMirror re-renders and normalises whitespace; raw inequality against the inserted string false-positives and fires `prompt_formatted` on a non-rewrite — today's bug with better telemetry. Observed signal was 19 → 651 chars.
2. **Match the existing settle-gate idiom** — `time.monotonic()` deadline + `page.wait_for_timeout(poll)`, as `_locate_body_prompt_box` and five siblings already do. No new helper, no `wait_for_function` (locators over `evaluate`), no `expect()` (appears nowhere in this file).
3. **Stay best-effort.** No new exit code. Promoting the miss to an exception lands it in `EXIT_CODE_MAP`, the MCP schema and the docs at once — and the body-prompt call site runs *after* the face image spent quota, so failing there aborts a saga mid-spend.

**P1 — Architect:**
4. The baseline is a **parameter, not instance state**. Both callers already hold the string they inserted. Stashing it on `self` gives a page-reused transport hidden temporal coupling across generations.
5. Keep `-> bool`; change its *meaning*, do not widen it. Both callers discard it, so enriching it builds a channel with no consumer.
6. Module constants, no env var (`_BODY_SLOT_MOUNT_TIMEOUT_S` precedent). No config nobody sets.

**P2 — Security:**
7. **Never log the rewritten text, and never a truncated excerpt.** Flow *elaborates* a terse person description into a detailed physical one — the output is more PII-dense than the input, and `GFLOW_CLI_HISTORY_PROMPTS` governs the DB, not structlog. Log `prompt_len_before` / `prompt_len_after` + the 8-char `_prompt_hash_stable`, matching same-file precedent. A 70-char head is exactly where the physical description lands.
8. **Jitter the poll interval** — a fixed 250ms cadence is itself a deterministic signature, and `_jitter_ms` exists in this module for that reason.
9. **Monotonic deadline inside the existing try/except**, so an `eAenfb` hang cannot leave the page held.

**P4 — CLI/MCP UX:**
10. **Redefine `prompt_formatted` to mean the observed rewrite** (it already claims to); add `format_button_clicked` (old semantics) and `format_not_observed` (timeout). CHANGELOG it as a semantics change and fix `docs/LIVE_VERIFICATION_v0.45.0.md:106`, which asserts the old meaning.
11. **The `--help` is currently false** — the button *is* found and the prompt *is* still submitted as typed. Rewrite it to name both failure modes and disclose the ~5-15s cost.
12. Expose `prompt_formatted: false` in `--json` so agents detect the degrade without parsing stderr.
13. **No MCP surface exists** (`mcp/tools.py`, `worker/codec.py`, `docs/MCP.md` all clean; `character create` is `_MCP_EXEMPT` for GAP #691). The exemption's reason still holds and this *strengthens* it — a longer inline call is direct input to the worker-enqueue decision (#481). Record that in the #691 port note.
14. **Strengthen the e2e in the same PR**, or it re-certifies the click.

## Unresolved conflict — surfaced, not suppressed

**The timeout value.** P1 says ~15s (labs is unmeasured; there a no-rewrite pays the full budget before submitting). P3 says ~30s (costs are asymmetric — too long is bounded latency on an opt-in flag, too short *silently reintroduces the exact bug*, invisibly). P4 says ~20s.

**Resolution: take the generous end, and make the number stop mattering.** P5's framing is the strongest: the fallback on expiry *is* today's behaviour, so a generous cap means n=1 is no longer load-bearing. P1's labs concern is real but bounded, and only on an opt-in flag. **~30s cap, 250ms jittered poll, and log `elapsed_ms` on every success so the distribution accrues in production** rather than being demanded before shipping.

P3 also notes the evidence is *weaker* than n=1: the 250ms cadence bounds the true settle only to `(5.105, 5.355]`.

## Sequencing

P5 argues **#731 first**: `docs/CHARACTER.md` prints stale selectors marked **VERIFIED**, mirrored to the published website. A false claim marked verified actively misleads; this proposal fixes a flag with four months of zero reports. Then this change. Then **#730 as its own predicted sweep** — not another hand-fix.

## Cheapest experiment, if a tighter number is ever wanted

`scripts/dev/spike_format_prompt_effect.py` already does one run at `$0`, never submitting. Loop it: 5 prompts of varying length, one already-formatted (idempotency), one under DevTools network throttling (the error/empty path). ~15 minutes, no quota. Labs remains a **named blocker** — no unmoved account exists here.

ffroliva:
## Live e2e green — both faults fixed

```
tests/e2e/test_character_create_e2e.py::test_character_create_format_prompt_clicks_format_button
1 passed in 406.20s (0:06:46)      profile denon82, migrated host
```

Transport events from a direct run of the same path:

```
16:52:21.960  ui_automation.format_button_clicked   selector='flow-format-prompt-button button'
16:52:26.138  ui_automation.prompt_formatted        elapsed_ms=4178
                                                    prompt_len_before=102  prompt_len_after=608
                                                    prompt_hash=f20df23c
16:52:30.495  ui_automation.prompt_submitted
```

Click → **observed rewrite** → submit, in that order. The inversion of those last two is the entire bug. Telemetry carries lengths and a stable hash and no prompt text, per the predict's security condition.

## The control, because a result without one is a coincidence with formatting

| run | flag | outcome |
|---|---|---|
| e2e, before the gate | `--format-prompt` | passed in **219.90s** — but asserted only the click |
| e2e, after the gate | `--format-prompt` | **TimeoutExpired at 480s** |
| direct diagnostic | `--format-prompt` | OOM-killed at ~5min mid-poll — *not* evidence of a stall |
| **control** | **none** | **completed in 210s**, image on disk |
| e2e, re-run | `--format-prompt` | **passed in 406.20s** |

Generation on this account was never broken. The overrun tracks the flag, and only ~9s of it is the gate — the rest is Flow generating from a 608-character elaboration instead of 102 characters. **406s vs 210s, roughly double.**

That is not a regression. It is what the feature costs, and it was free only while it was doing nothing. It is also why the old 300s budget failed: that budget was fitted to a flag that did not work.

## Claims I had to retract

An hour ago I wrote, in four places, that this flag does not work. All four are now false and all four were mine — a CHANGELOG "Known issues" block, the `--help` string, a spike-doc title, and that doc's conclusion. Swept by grep rather than recall, since the one that matters is always the one nobody thought to look at.

The spike doc keeps its past-tense title under a RESOLVED banner rather than being deleted: it is a dated measurement, and it is the evidence that makes the fix defensible.

## Also fixed, found by the overrun

`_run_gflow` timed out and then **hung** — `subprocess.run(timeout=)` kills the child, then re-enters `communicate()` to drain pipes that surviving Chrome grandchildren still hold open. 32 minutes at near-zero CPU, indistinguishable from slow progress. My first fix bounded the drain, stopped the hang, and **discarded the output** — trading a hang for a blind failure that reported a command and a duration and not one transport event. Redirecting to files removes the class: nothing blocks on a reader, and the evidence survives the timeout path.

Five sibling e2e modules carry their own pipe-based `_run_gflow`. Named in the docstring, not changed — I cannot exercise the timeout path in the others.

Issue #731, opened 2026-09-07
docs/CHARACTER.md marks stale <i>-only selectors VERIFIED; docs/USAGE.md never documents --format-prompt


ffroliva:
## What

Two docs assert things about the character editor that the shipped code contradicts. Both predate #727/#729 — they are fallout from **#703**, which changed the selectors and did not sweep the prose.

### 1. `docs/CHARACTER.md` marks stale selectors as VERIFIED

Lines ~442-451 print a table of selectors marked **VERIFIED**, e.g.:

```
button:has(i.google-symbols:text-is('accessibility_new'))
```

The shipped code carries `mat-icon` variants for all three of those anchors (`src/gflow_cli/api/transports/ui_automation.py`, lines 3417 / 3443 / 3475 at `7c1d4c6d`). Line ~423 still prescribes `i.google-symbols:text('<lig>')` as *the* pattern — the exact single-carrier rule that caused #727.

The `website/docs/` mirror carries the same rows, so this is published.

**VERIFIED is the load-bearing word.** A stale row marked verified is worse than no row: it is the thing a reader trusts instead of reading the code.

### 2. `docs/USAGE.md` never documents `--format-prompt`

```
grep -n "format-prompt" docs/USAGE.md
(no output)
```

The `gflow character create` options block lists 8 flags; `--format-prompt` shipped in **v0.45.0** and is not among them. Silence rather than a false claim, so lower severity — but `docs/USAGE.md` is specifically the file the repo's own `correct-by-grep-not-recall` lesson names as "the one users read", and this is the flag that just spent a release being a silent no-op.

## Suggested fix

- Re-derive the CHARACTER.md selector rows from the code rather than editing them in place, and either drop the VERIFIED column or date each row against the capture that verified it.
- Add `--format-prompt` to the `character create` block in `docs/USAGE.md`, describing the **best-effort** contract accurately (a missed button warns on stderr and still submits, exit 0).
- Regenerate `website/docs/` (`uv run python scripts/ci/generate_website_docs.py`).

## Note

`scripts/ci/check_doc_links.py` uses an explicit allowlist (`FILES`, line 19) with **zero** `spikes/` entries, so its green run does not cover files under `docs/superpowers/spikes/`. Worth deciding whether that is intentional.

Related: #703, #727, #729, #730.

Issue #730, opened 2026-09-07
The ligature-carrier sweep is unfinished: 55 single-carrier selectors, and nothing watches them


ffroliva:
## What

`#703` swept three selector constants for the labs/migrated **ligature-carrier split** (`<i class="google-symbols">` vs `<mat-icon>`) — `add_2`, `arrow_drop_down`, `accessibility_new`. It missed `PROMPT_FORMAT_SELECTORS`, which became a silent no-op and shipped as #727.

**Nothing proves that was the last one.** Measured on 2026-09-07 at `7c1d4c6d`:

```
git grep "i\.google-symbols" src/gflow_cli --include=*.py | grep -v "mat-icon" | wc -l
55
```

55 single-carrier literals across `ui_automation.py`, `ui_automation_video.py`, `mode_control.py`, `diagnostics.py` and `drivers/{agentic,factory}.py`. Two consecutive PRs have now fixed one of these **by hand**, which is the signal that this is a sweep problem, not a selector problem.

Named candidates from the #729 council (all pre-existing, none introduced there):

| Constant | Location | Exposure |
|---|---|---|
| `IMAGE_MODEL_PICKER_TRIGGER` | `ui_automation.py:104` | **Sharpest.** Sole selector, no `mat-icon` twin — while its sibling `_CHARACTER_MODEL_PICKER_TRIGGER_SELECTORS` in the same file carries both *and* a comment recording that omitting mat-icon logged `model-picker trigger not found` on the migrated host. Likely masked today because `_select_image_model` only runs after a panel that warn-returns first — **not verified on the migrated host**. |
| `ui_automation_video.py:251` | same shape | same class |
| `OVERLAY_CLOSE_BUTTON_SELECTORS` | `ui_automation.py:490-494` | `clear` / `close` are `<i>`-only. Degraded, not dead — the `[role=dialog]:has(a[href*=changelog])`, `button[data-dismiss]` and Escape fallbacks remain. |
| `NEW_PROJECT_SELECTORS` | `ui_automation.py:329-331` | `add_2` tier is `<i>`-only; falls back to a regex + 14 localised labels. |

## Why nothing caught it, and why nothing will catch the next one

`.github/workflows/selector-probe.yml` drives `run_probe.py --surface editor` against `src/gflow_cli/flow_selectors/registry.py`. That registry has **zero** entries for the Format button:

```
grep -rn "personal_recommendations\|format" src/gflow_cli/flow_selectors/ --include=*.py
[...]

PR #735, opened 2026-09-07
fix(selectors): de-blind the diagnostics, and fix the two selectors hiding behind them

ffroliva:
## Summary

Scoped down from #730's "sweep 24 constants" after `/gflow:predict` returned **CAUTION 7.3/10** from three personas — unanimous: *right change, wrong blast radius*. Four changes ship, every one backed by a measurement.

Refs #730 (the full sweep stays open; see "Not touched" below)

## The one that matters: our diagnostics were blind

```
ui_automation.py   _gsQuery         = 'i.google-symbols, span.google-symbols'
diagnostics.py     STRUCTURAL_DOM_JS  querySelectorAll('i.google-symbols')
```

On `flow.google.com`, where every Material Symbols ligature rides a `<mat-icon>`, **both returned zero.**

The instrument we reach for to diagnose selector drift was blind to the host the drift lives on. That is why #727 and #731 stayed invisible for weeks, and why an incident bundle from a migrated user would have reported **no ligatures at all**. Both queries are now class-only.

## Two selectors were broken behind it

| Constant | State |
|---|---|
| `SUBMIT_BUTTON_SELECTORS` | **Working by luck.** Entries 0–1 matched nothing on migrated; the submit only landed because entry 2 matches the `<mat-icon>`'s *text*. Live: `prompt_submitted via="button:has-text('arrow_forward')"` — after paying two misses per submit |
| `IMAGE_MODEL_PICKER_TRIGGER` | No twin, no fallback → total silent miss. Best-effort, so generation proceeded on **whatever model tier the editor opened at** |

Both now anchor on the `google-symbols` **class**, which sits on the labs `<i>` and the migrated `<mat-icon>` alike.

## The premise is now observed, not asserted

The whole proposal rested on "`.google-symbols` matches both carriers" — CSS semantics, never checked. `tests/api/transports/test_ligature_carrier.py` proves it against a **real CSS engine** with a two-carrier fixture: no account, no network, no credits. If it ever fails, the premise is wrong and this reverts.

**It immediately caught a defect in my own fix: `matched 3 of 2`.** `SUBMIT_BUTTON_SELECTORS` used `:text`, a *substring* match, which also accepts `arrow_forward_ios` — a real Material Symbol. The `<i>` qualifier had been containing that; dropping it for the class-only carrier made the over-match reachable on the load-bearing submit path. Now `:text-is`.

So my own "one character each" framing was **wrong**: dropping a tag qualifier can expose a latent substring over-match, and each site needs its exactness checked too.

## Not touched, each for a measured reason

- **The negation sites** (`button:not(:has(i.google-symbols))`, and a `has_not` filter). Broadening **inverts** them — they would exclude *more* buttons, and on a dialog whose buttons carry mat-icons the iconless set collapses to zero → no include click → a **120s hang**. They are also already broken the other way on migrated.
- **The 10 mode-detection selectors.** They feed a classifier that calls `raise_if_migrated` every tick, so they never execute on the host this fixes. Zero upside, nonzero risk on labs — which is unmeasured.
- **The `add_2` constants.** A ligature **name** drift, not a carrier one, and `_CHARACTER_SLOT_ADD_LIGATURE` feeds a JS exact-match heuristic where a string swap is not the fix.

## Validation

- [x] `pytest -q --cov` — **4043 passed**, 7 skipped
- [x] `pyright src` — 0 errors · `ruff check` / `format --check` clean
- [x] hygiene 1030 files · doc links · website mirror · memory round trip · PII scan
- [x] New offline browser fixture, 6 tests, no credits

**E2E:** the changed selectors sit on the submit and model-picker paths, exercised by the character-create e2e. Not re-run here — the previous live run on this branch's parent predates these selector edits, so I am **not** claiming live coverage for them. The offline two-carrier fixture is what backs this PR; a live run belongs with whoever takes the remaining sweep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Flow's own diagnostics were blind on the migrated host, and two selectors were broken behind
+  it.** `gflow`'s incident-bundle DOM dump queried `i.google-symbols, span.google-symbols`, and
+  `diagnostics.py`'s `STRUCTURAL_DOM_JS` queried `i.google-symbols` — so on `flow.google.com`,
+  where every Material Symbols ligature rides a `<mat-icon>`, **both reported zero ligatures**.
+  The instrument used to diagnose selector drift was blind to the host the drift lives on. That
+  is why [#727](https://github.com/ffroliva/gflow-cli/issues/727) and
+  [#731](https://github.com/ffroliva/gflow-cli/issues/731) stayed invisible, and why an incident
+  bundle from a migrated user named no ligatures at all. Both queries are now class-only.
+
+  Two selectors were broken behind that blindness. `SUBMIT_BUTTON_SELECTORS` was **working by
+  luck**: its two tag-qualified entries matched nothing on the migrated host and the submit only
+  landed because a third `has-text` entry matches the `<mat-icon>`'s *text* — observed live as
+  `prompt_submitted via="button:has-text('arrow_forward')"`, after paying two misses per submit.
+  `IMAGE_MODEL_PICKER_TRIGGER` had neither twin nor fallback, so its miss was total and silent:
+  the picker is best-effort, so generation simply proceeded on whatever model tier the editor
+  opened at. Both now anchor on the `google-symbols` **class**, which sits on the labs `<i>` and
+  the migrated `<mat-icon>` alike — verified against a real CSS engine, offline, in
+  `tests/api/transports/test_ligature_carrier.py`.
+
+  `SUBMIT_BUTTON_SELECTORS` also moves from `:text` to **`:text-is`**. `:text` is a substring
+  match and accepts `arrow_forward_ios`, a real Material Symbol; the `<i>` qualifier had been
+  containing that, and dropping it for the class-only carrier made the over-match reachable on
+  the submit path. The two-carrier fixture caught it as `matched 3 of 2` before it ran live.
+  ([#730](https://github.com/ffroliva/gflow-cli/issues/730),
+  [spike](docs/superpowers/spikes/2026-09-07-ligature-carrier-and-name-drift.md))
+
 - **`character create --format-prompt` works again — both halves of it.** The flag had two
   independent faults, and fixing only the first would have left it just as useless. On the
   migrated `flow.google.com` host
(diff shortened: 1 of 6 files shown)


ffroliva:
## Live verification — the E2E gap is closed

**E2E:** `tests/e2e/test_character_create_e2e.py::test_character_create_format_prompt_clicks_format_button` → **`1 passed in 219.19s`** on `denon82`. That drives `_click_submit` with the changed `SUBMIT_BUTTON_SELECTORS` through a real generation.

But pytest swallows stderr on a pass, so it cannot show *which* cascade entry fired — and "still works" was never the claim. "Works via the fixed selector rather than the accidental text fallback" is. So the sweep spike now counts the shipped constant against the form it replaced, on both surfaces:

| surface | anchor | shipped | pre-#730 |
|---|---|---|---|
| composer | submit | **1** | **0** |
| composer | image model picker | 0 | 0 |
| composer | `STRUCTURAL_DOM_JS` | **47 ligatures, 17 distinct** | would be 0 |
| character editor | submit | **1** | **0** |
| character editor | image model picker | **1** | **0** |
| character editor | `STRUCTURAL_DOM_JS` | **14 ligatures, 13 distinct** | would be 0 |

**`pre_730=0` everywhere is the point.** The old anchors matched *nothing* on this host; the new ones resolve. And the diagnostics went from blind to 47 ligatures — that is the `ui.json` a migrated user's incident bundle would have carried, previously empty.

## The probe caught itself

The first verification run reported **zero ligatures of any kind** on the composer, where an earlier run on the same URL saw 8. The page had not rendered its icons at the fixed 3s mark.

I had given the character editor a control and left the composer without one — so those zeros were unreadable, and I nearly reported them as findings. That is precisely the failure this week has been about, reproduced in my own instrument. **One control on one surface certifies only that surface.**

Fixed twice over: every surface now carries its own control and hard-errors *after* the dump, and the composer waits for `.google-symbols` to attach rather than sleeping a fixed interval — the thing being counted is the thing to wait for. Same URL now yields 17 hits.

That also promotes `image_model_picker=0` on the composer from an unreadable zero to a **meaningful** one: the control passed, so that button genuinely is not on that surface — consistent with the migrated composer having no image mode.

## Still not measured

**labs.** No unmoved account exists here. The `<i class="google-symbols">` half of the class-only claim is proven by the offline two-carrier fixture against a real CSS engine, not by a live labs run. Named, not hidden.

Issue #723, opened 2026-09-07
Character entity mentions DO work on the migrated composer — the gesture is @ + query + Enter; port it, and make entity outrank media


ffroliva:
## Summary

`gflow video r2v/t2v --reference-entity` and `@Name` exit 36 on the migrated host as "not ported". **The host takes them fine.** Measured 2026-09-07, $0, profile `ffroliva`: typing `@`, then the character's name, then `Enter` in the migrated composer inserts a real character chip.

```html
<span class="mention-chip"
      data-mention-id="75cd2b39-c345-4b96-ae1b-09360cc44825"
      data-reference-type="entity"
      data-entity-id="75cd2b39-c345-4b96-ae1b-09360cc44825"
      contenteditable="false">Kael</span>
```

The `entity_id` is the character created by `gflow character create` minutes earlier. So this is a **porting gap in the driver**, not a capability gap in Flow — and it is now the last thing standing between a migrated account and multi-shot character consistency.

## The gesture matrix

`scripts/dev/capture_migrated_mention_gestures.py ffroliva <project> --query Kael`, one browser session, composer cleared between attempts:

| gesture | chip | reference_type |
|---|---|---|
| `arrow_enter` | — | picker dismissed, `@` cleared |
| `query` (type `@Kael`, stop) | — | picker open, nothing committed |
| **`query_enter`** | **`Kael`** | **`entity`** ✅ |
| `query_arrow_enter` | `kael_ref.jpg` | `media` |
| `slow_query` | — | picker open, nothing committed |

Two things follow.

### 1. The working gesture is `@` + query + Enter

`@` must be typed with real keystrokes — `page.keyboard.type("@", delay=120)`. `insert_text("@")` dispatches input events without keystrokes, the ProseMirror mention plugin opens its picker with no query to track, and every later gesture operates on a picker with no state behind it. That was already established in `docs/superpowers/spikes/2026-09-05-migrated-r2v-attach-surface.md`; this run confirms it and adds that **Enter without ArrowDown** is what commits the *entity*.

### 2. The picker does not rank entities above media — and it must

`query_arrow_enter` returned a **media** chip for the same query: a JPEG that happened to be named `kael_ref.jpg`. Same name, same keystrokes but for one arrow, a completely different anchor and no `entity_id` at all.

[...]

Issue #722, opened 2026-09-07
Incident bundles on the migrated video path are captured after the page is parked, so they ship an all-zero DOM and a blank screenshot


ffroliva:
## Summary

On the migrated host's video path, the incident bundle is captured **after** the pooled page has been parked at `about:blank`. So every such bundle ships:

```json
"tag_counts": {"div": 0, "button": 0, "input": 0, "textarea": 0,
               "dialog": 0, "iframe": 0, "video": 0, "img": 0},
"ligatures": [], "ligature_count": 0, "overlays": [],
"title": {"category": "other", "length": 0},
"url": {"host_category": "other", "route": "other"}
```

and a blank white screenshot.

## Why this is worse than a missing artifact

The bundle is not *empty*, it is **confidently empty**. It looks like a successful capture of a dead page, so a reader concludes "the page was gone" — when the truth is "we blanked it before looking". A missing artifact prompts you to go and measure. A zeroed one answers the question, wrongly, and stops the investigation.

This is the same failure family the project has been correcting all week: a measurement that could not be taken, recorded as a measurement of absence. Here the diagnostics themselves manufacture it, on every incident.

**It has already cost.** #692's reporter attached a bundle with `ligature_count: 0` and every element count at zero, and that was read as evidence about the migrated DOM. It was evidence about capture ordering.

## Mechanism

`_generate_video_locked` wraps the migrated composer call in a `try/finally`, and the `finally` parks the pooled page:

```python
finally:
    # The pooled page would otherwise stay on flow.google.com/project/<id>, and the
    # NEXT request on this client would be routed by that URL instead of by its own shape
    await page.goto("about:blank", wait_until="commit", timeout=5_000)
```

The park is correct and should stay — it prevents the next request being routed by a stale URL. The problem is ordering: incident capture happens at the CLI layer (`run_with_handlers`), which is **outside** the transport, so it runs after the `finally` has already blanked the page it is about to photograph.

## Measured today

Two `UiSelectorDriftError` incidents from `gflow video t2v` on the migrated host, both with all element counts zero and a blank screenshot:

- `20260907T101449Z-3beb72d8-…`
- `20260907T101604Z-d3f87d23-…`

Both were undiagnosable from their bundles. The actual cause — the account was out of Veo credits, and Flow had replaced the submit control with a `prompt-warning-button` (#721) — was only findable by writing a probe that inspects the page **while it is still alive**. The information was on screen at failure time and the bundle threw it away.

## Suggested fix

[...]

PR #736, opened 2026-09-07
test(character): verify --voice and --personality actually attach

ffroliva:
## Summary

`--voice` had **never been exercised end-to-end**. A repo-wide grep for `--voice` across `tests/e2e/` matched nothing.

Every voice test in the suite is a unit test of the hardcoded `VOICES` constant — list length, capitalisation, the sample-URL string pattern, dataclass frozen-ness — and not one reaches Flow. The single test that *looks* live (`chars[0].voice == "gacrux"`) parses a fixture. A voice that silently failed to attach was therefore invisible to the whole suite while `character create` exited 0, and `test_character_create_binds_parent_entity` stops one field short of the voice.

This adds the missing coverage and corrects two documents with the evidence it produced.

## What the run proved

```
[voice-case-evidence] sent='Charon' stored='Charon' identical=True
1 passed in 227.92s (0:03:47)
```

Profile `ffroliva`, project `c5550ed7`, one image generation, zero credits.

| Question | Answer |
|---|---|
| Does `--voice` attach at all? | **Yes** — Flow's read-back reports the voice |
| Which case does the wire store? | **Capitalized, unchanged** |
| Does `--personality` attach? | **Yes** — asserted hard, no `if` guard |

## Docs corrected

Our own two documents disagreed, and one was wrong:

- `docs/CHARACTER.md` called the Capitalized name canonical while flagging the question **UNVERIFIED**.
- `docs/CHARACTER_RECON.md` recorded `presetVoiceId: "gacrux"` and stated *"preset id is lowercased name"*.

The caveat is now the measurement, and it names what remains **untested**: whether a lowercase id is also accepted (gflow never sends one), and — separately and more importantly — whether a bound voice is actually **applied to generated video audio**. Attachment is proven; application is not.

Both docs also now record that **`personalityNotes` is Agent-scoped**. Flow's own character editor states:

> "The Flow agent can use this information to help craft scenes with your character."

It is not a control on the audio engine and not a documented input to a direct composer generation. Nothing in our docs said this, which invites expecting it to steer a `t2v` / `r2v` performance — put that direction in the prompt instead.

## Design notes

- **Voice case is compared case-insensitively on purpose.** The defect worth catching is a voice that does not attach AT ALL; pinning either spelling would go red for a reason that is not that defect. The run prints the spelling Flow returned, so it settles the contradiction as evidence rather than opinion.
- **Personality is asserted hard.** `test_character_personality_utf8` guards it as `if shown_personality:` — a soft check that passes vacuously when the read-back omits the field, which is exactly the failure worth catching.

## Test plan

- [x] `ruff check src tests` — All checks passed
- [x] `ruff format --check src tests` — 456 files already formatted
- [x] `check_repo_hygiene.py` — 1028 files, no violations
- [x] `check_doc_links.py` — 29 files resolved
- [x] `generate_website_docs.py --check` — mirror in sync (regenerated)
- [x] `check_website_docs_pii.py` — clean
- [x] `check_council_memory.py` — 49 files, all cited
- [x] Character unit tests — 66 passed
- [x] **e2e, live** — `test_character_create_attaches_voice_and_personality` passed against `ffroliva`

## Known follow-ups (not in this PR)

- The test leaves a `voice-attach-<hex>` entity behind per run; the file's other tests share that property.
- Whether a bound voice reaches rendered video audio is still unproven — the open question this test deliberately does not claim.

### DIFF
--- a/docs/CHARACTER.md
+++ b/docs/CHARACTER.md
@@ -262,12 +262,28 @@ token that drives the sample URL, so a canonical id round-trips to a playable sa
 `gflow character create --voice <Name>` validates the value **case-insensitively** (so `charon` and `Charon`
 both normalize to the canonical `Charon`) and sets `audioReferences[].presetVoiceId` via the entity PATCH.
 
-> **Wire-case caveat (open):** a prior live run sent a **lowercase** id (`"charon"`) and Flow persisted it,
-> but whether Flow applies the voice from a lowercase id vs the Capitalized canonical form is **UNVERIFIED**.
-> gflow adopts the Capitalized form as canonical per the UI. The voice list is currently a **hardcoded
-> constant**; fetching the live list from Flow's voice API and confirming the `presetVoiceId` wire-case are
-> tracked in the [Backlog](#14-backlog--not-yet-implemented). A "create new voice" flow ("Criar nova voz")
-> exists in the UI but is **not yet implemented**.
+> **Wire-case: Capitalized round-trips — VERIFIED 2026-09-07.** A live e2e
+> (`tests/e2e/test_character_create_e2e.py::test_character_create_attaches_voice_and_personality`,
+> profile `ffroliva`) created a character with `--voice Charon` and read it back from Flow:
+> `sent='Charon' stored='Charon' identical=True`. The Capitalized canonical form is stored
+> unchanged, so `CHARACTER_RECON.md`'s older note that "the preset id is the lowercased name"
+> does **not** describe what the wire does today.
+>
+> **Still untested:** whether a *lowercase* id is also accepted (gflow never sends one — it
+> normalizes to Capitalized before the PATCH), and — separately and more importantly —
+> whether a bound voice is actually **applied to generated video audio**. Attachment to the
+> entity is proven; application at render time is not.
+>
+> **`personalityNotes` is Agent-scoped.** Flow's own character editor labels the field
+> "Character info (optional) — Describe how your character acts…" and states underneath:
+> *"The Flow agent can use this information to help craft scenes with your character."*
+> It is an input to Flow's **agent** when it writes scenes, not a control on the audio engine
+> and not a documented input to a direct composer generation. Do not expect it to steer
+> performance on a `video t2v` / `r2v` call — put the direction in the prompt instead.
+>
+> The voice list is currently a **hardcoded constant**; fetching the live list from Flow's
+> voice API is tracked in the [Backlog](#14-backlog--not-yet-implemented). A "create new voice"
+> flow ("Criar nova voz") exists in the UI but is **not yet implemented**.
 
 ## 8. CLI surface (shipped v0.12.0)
 
--- a/docs/CHARACTER_RECON.md
+++ b/docs/CHARACTER_RECON.md
@@ -28,8 +28,13 @@ entity {
     entityType: "CHARACTER"
     displayName                  # "Denidra" (default "Untitled Character")
     characterInfo {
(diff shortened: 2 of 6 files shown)


ffroliva:
## Second commit added: a diagnostic spike for #723

`5a2be11` adds `scripts/dev/spike_entity_submit_rpcs.py`. Flagging it explicitly because it is **adjacent scope** — this PR is the attachment e2e, and the spike answers a different question. It rides along because it came directly out of trying to use what this PR verifies: `--voice` attaches to a character, and then the character cannot be attached to a generation.

Split it out if you'd rather keep this PR to the e2e; the file is standalone and has no dependency on the rest of the branch.

### What it establishes

`_unported_form` refuses character references, saying the submit *"never produces a YhhmEf/eb1hJf/MZZa6b reply"*, cause unknown, and undiagnosable from an incident bundle because of #722.

Reading the wire live shows that note is wrong on the key point:

```
MZZa6b -> [["wrb.fr","MZZa6b",null,null,null,[5],"generic"]]   null result + error slot
WuwhI  -> []
jwpduf -> the generation IS in the project, prompt prefixed with the entity name
```

It replies. The run enqueues and then **fails server-side** — Flow's UI: *"Failed… You have not been charged."* So there is nothing to add to `SUBMIT_RPCS`, the guard is correct, and only its stated reason needs updating. **#722 turns out not to be on this path at all**, since reading the wire live sidesteps the empty bundle.

### One thing the commit records against itself

The first run of the probe was invalid: it skipped `apply_video_settings`, the stage that selects the **Ingredients** submode, and the composer's own comment already warns that a character chip submitted from under Frames gets no reply. Re-run with settings applied, the outcome is identical — so the finding stands, but the first measurement did not, and the commit message says so.

ffroliva:
## Branch now carries four commits — and one of them retracts another

Flagging the shape of this branch, because it grew past its title and one commit corrects an earlier one on the same branch.

| commit | what it does |
|---|---|
| `38f0a1cf` | **the PR's actual subject** — e2e proving `--voice` and `--personality` attach, plus the `CHARACTER.md` / `CHARACTER_RECON.md` corrections |
| `5a2be115` | spike that reads the wire live for #723 |
| `1958b5af` | ⚠️ recorded "the backend rejects entity-bound video" on the `_unported_form` guard — **wrong** |
| `276d5e02` | **retracts `1958b5af`** with evidence |
| `40ccf9c0` | read-only project status probe (#741) |

### Why `1958b5af` was wrong

It concluded a backend refusal from the submit reply. Entity-bound generations are actually **accepted and queued** — Flow derives an `abra_r2v_8s` model key and they proceed. What fails is the observer: `MZZa6b` returns a null payload, `submit_and_observe` never sees a media id, and `SUBMIT_REPLY_BUDGET_S = 60.0` expires — a value calibrated where *"the submit reply arrived 4.0–4.6 s after the click"*, against an idle queue.

The account owner disproved my conclusion by opening the project and seeing **Queued**, after I had asserted the opposite three times.

`276d5e02` leaves the guard in place but with an honest reason, since until the observer is fixed the CLI would report a timeout on a video that is actually rendering — worse for a user than an explicit refusal.

### Reviewer note

`1958b5af` and `276d5e02` could be squashed, but I'd rather leave both: a wrong diagnosis that was committed, published to an issue, and then retracted is worth being visible in history — particularly since the *original* note on that guard was the same class of error, and reading only the final state would hide that it has now happened three times to three different people.

Happy to split the spikes out of this PR if you'd prefer it kept to the e2e; they're standalone files with no dependency on the rest of the branch.

Issue #749, opened 2026-09-08
Migrated host (flow.google.com): settings trigger button relocated to agent-footer-actions, requires composer interaction first


lopezdu53:
## Summary

On an account migrated to `flow.google.com` (the "agentic" Flow UI cohort mentioned in #183/#174), `READY_ANCHOR = ".settings-trigger-button"` resolves to a `<button>` that is permanently `hidden` in the DOM. The real settings button now lives elsewhere in the page and only becomes interactive after the composer has been focused/used.

## Environment

- gflow-cli version: 0.71.0
- OS: Windows 11
- Commands affected: `gflow video t2v`, `gflow image t2i`
- Account: migrated to flow.google.com (agentic UI cohort)

## What I found

**1. The old selector target is dead, not missing:**

```html
<button hidden="" matbutton="" flow-button="" cdkoverlayorigin=""
  aria-label="Settings trigger"
  class="mdc-button mat-mdc-button-base settings-trigger-button mdc-button--unelevated
         mat-mdc-unelevated-button mat-unthemed flow-button-primary flow-button-small">
  …
</button>
```

It's present in the DOM (57 locator matches in the wait_for retry loop) but `hidden` is set directly on the element — it will never become visible, so `Locator.wait_for(state="visible")` always times out at 30s.

**2. The real settings button lives in `agent-footer-actions`, next to the prompt composer:**

```html
<button _ngcontent-ng-c1385090464="" flow-icon-button maticonbutton
  class="mdc-icon-button mat-mdc-icon-button mat-mdc-button-base mat-mdc-tooltip-trigger
         agent-action-button mat-unthemed flow-icon-button-transparent flow-button-small"
  mat-ripple-loader-class-name="mat-mdc-button-ripple" mat-ripple-loader-centered
  aria-label="Configuración">
  <span class="mat-mdc-button-persistent-ripple mdc-icon-button__ripple"></span>
  <mat-icon role="img" class="mat-icon notranslate flow-icon-m google-symbols mat-icon-no-color"
    aria-hidden="true" data-mat-icon-type="font">tune</mat-icon>
  …
</button>
```

Confirmed via `playwright codegen` against a real logged-in session — recording a manual click on this button produced:

```python
page.get_by_role("button", name="Configuración", exact=True).click()
```

(Account locale is Spanish; presumably `"Settings"` in English locales — worth confirming across locales the way #24 handled locale-agnostic selectors.)

**3. The new button is not available until the composer is touched.**

[...]

Issue #719, opened 2026-09-07
migrated host: an upload request leaves the page and Flow never answers (intermittent, ~1 in 4 on a consented account)


ffroliva:
## Summary

On the migrated `flow.google.com` host, every local-file upload fails with `MediaUploadRejectedError` (exit 27) — `"no maseQ reply within 60s of choosing the file"`. This blocks **both** attach surfaces: `video i2v --initial-frame <local>` and `video r2v --ref <local>`, i.e. the two forms `KNOWN_ISSUES.md` currently lists as the ported ones.

Measured today, profile `ci-probe`, project `1e4efe0d-…`, gflow-cli 0.70.0.

## The file is ruled out

The error's remediation hint says to re-encode the image to strip metadata. That points at the wrong thing. Three different files fail **identically**:

| File | Origin | Size | Result |
|---|---|---|---|
| `kael_ref.png` | generated by Flow's own character editor | 734 KB | exit 27, no maseQ reply |
| `kael_ref.jpg` | `ffmpeg -map_metadata -1 -q:v 2` re-encode of the above | 206 KB | exit 27, no maseQ reply |
| `tiny.png` | `ffmpeg -f lavfi -i color=c=0x8B5A2B:s=1280x720` — a flat brown rectangle, no metadata at all | **4.3 KB** | exit 27, no maseQ reply |

A 4.3 KB synthetic solid colour with no metadata is about as benign as an upload gets. Whatever is failing is not the file, and the hint sends the reporter on a detour. Worth rewording regardless of the root cause.

## Both attach surfaces, same wall

- `video r2v --ref <local .png>` → exit 27
- `video r2v --ref <local .jpg>` → exit 27
- `video i2v --initial-frame <local .jpg>` → exit 27
- `video i2v --initial-frame <tiny .png>` → exit 27

`video t2v` on the same profile, same project, same session **works** — so this is not auth, not the project, and not the composer generally. Settings are applied successfully immediately before the failure.

## What the log shows, and what it does not

```
migrated.dispatch → migrated.navigate → migrated.editor_ready
→ migrated.model_already_selected → migrated.settings_applied → error_raised
```

There is **no upload-attempt telemetry between `settings_applied` and the failure.** No event for the attach affordance being found, clicked, or for a file chooser firing. So from the logs alone it is not possible to say whether:

1. the attach control was never located,
2. it was clicked but no file chooser opened,
3. the chooser opened and the file was set but no request went out, or
[...]

Issue #716, opened 2026-09-07
video t2v silently drops --reference-entity on the migrated host and bills the generation


ffroliva:
## Summary

On the migrated `flow.google.com` host, `gflow video t2v --reference-entity <id>` (and the `@Name` form, which resolves to the same field) appears to **submit with the entity silently dropped, and bill the generation**. The user gets a clip of a stranger and no warning.

R2V has an explicit guard for this. T2V does not.

## Where

`src/gflow_cli/api/transports/migrated_composer.py`, `_unported_form()`:

```python
if request.mode is Mode.T2V:
    return None                      # <-- never inspects request.reference_entities
if request.mode is Mode.R2V:
    ...
    if request.reference_entities:
        return "character references"   # <-- R2V is guarded
```

`run_video()` calls `_unported_form(request)` first and raises `FlowHostMigratedError` when it returns a noun. For T2V it returns `None`, so the run proceeds — and nothing downstream attaches an entity. `attach_start_frame` is I2V-only, `attach_references` is R2V-only, `read_chips` verification is R2V-only. There is no T2V entity path and no T2V entity check.

## Why `migrated_can_serve` does not save it

`migrated_can_serve()` **does** refuse on `reference_entities`, but it only feeds `prefer_migrated`. For an account Google has already moved, `migrated_route()` decides from the URL:

```python
kind = flow_host_kind(url)
if kind == "migrated":
    return "blocked" if flow_host == "labs.google" else "migrated"
```

`prefer_migrated` is not consulted on that branch — as `migrated_can_serve`'s own docstring says: *"An account Flow has already moved is routed by its URL and never reaches this question."* So the refusal is unreachable for exactly the accounts that need it.

The sequence on a moved account: the pooled page is parked at `about:blank`, so the first route call returns `labs`; `_enter_editor` navigates to labs, which hops to `flow.google.com`; the re-check now reads `migrated`; `run_video` takes it; T2V passes `_unported_form`; the entity is gone.

## Impact

Worse than an exit 36. A refusal costs nothing and tells the truth. This spends credits and returns a plausible clip with the wrong person in it, and the whole point of a character entity is that the person is not supposed to change. It also defeats the labs path's own protection — that path intercepts the submit and **raises** if the expected `referenceEntities` are missing (`ui_automation_video.py:2964-2994`).

## NOT RUN — this is a code-read finding

[...]

Issue #673, opened 2026-09-05
Moved account: `image` commands die with RecaptchaError (exit 1) before the exit-36 guard fires


ffroliva:
Split out of #639, where two independent accounts reported the same outcome on the image path after Google moved them to `flow.google.com`.

## Symptom

`gflow image t2i` / `image i2i` on a fully migrated account exits **1 `RecaptchaError`** — not 36 `FlowHostMigratedError`, which is what `docs/USAGE.md` and `KNOWN_ISSUES.md` say a moved account gets for the not-yet-ported commands. Zero credits, no submit, but the error names the wrong cause and its remediation points at reCAPTCHA, not at the migration.

## Two shapes, two accounts

1. **Stranded on the labs.google logged-out landing page** — henricurti, v0.66.3, `image i2i --ui-mode classic`, fresh `auth login` minutes before: exit 1 after ~24 s; the incident bundle's network records are the old Next.js frontend plus `/aitestkitchen/website/flow/landing_page/…`, only 4 requests touch the migrated origin, `generation_requests: 0`. Cookie harvest collapsed from `all_cookies: 9 / flow_domain: 7` to `2 / 1` after the login — the #644 harvest keyed on `labs.google` handing a logged-out jar to the driven browser. https://github.com/ffroliva/gflow-cli/issues/639#issuecomment-5548973773
2. **Landed on the flow.google.com project grid, never the editor** — ChandraLiuswanto, v0.67.0, `image t2i --project <id>`: exit 1 in ~6 s; `ui.json` shows `host_category: flow_app`, `route: /` (38 buttons, 11 imgs). The run never went to labs.google. https://github.com/ffroliva/gflow-cli/issues/639#issuecomment-5552103393

Shape 2 is the one 0.67.0 should already handle: `raise_if_migrated` (`src/gflow_cli/api/transports/_common.py:112`) exists precisely so a page on the migrated origin exits 36 before any probe. A run that reaches the migrated **root** rather than `/project/<id>` evidently hits reCAPTCHA site-key discovery first.

## Also on the new host

The 640×576 non-modal `role=dialog` seen in earlier bundles is Flow's changelog modal on `flow.google.com/` ("Faster loading, simpler login, and more!", buttons "View all changelogs" / "Get started"), shown once per account after the handoff — same class as #593 on labs. The migrated composer will need the same dismissal precondition.

## To triage (read-only, $0)

[...]

PR #750, opened 2026-09-08
fix: support migrated Flow image generation

### LINKED ISSUE #692: RecaptchaError on migrated Flow host with gflow-cli 0.69.0 (image i2i)
I'm still hitting the original RecaptchaError path for `image i2i` on a migrated Flow account, even after upgrading to gflow-cli v0.69.0.

Environment:
- OS: Windows 10/11 (********)
- gflow-cli: 0.69.0
- Python: 3.14.0 (from incident report)
- Fresh `gflow auth login --profile *********** --browser chrome` succeeded
- Flow session was verified successfully (`[OK] Flow session verified`)
- Command: `gflow image i2i "<prompt text>" --ref ********** --project ********* --aspect 16:9 -o ***********`

Observed:
- `ui.json` reports:
  - `host_category: flow_app`
  - `route: /`
  - `ligature_count: 0`
  - `button/input/textarea/dialog: 0`
- The command still exits with:
  - `RecaptchaError`
  - exit code 1
  - phase/route: `image_generation / -`
- Incident bundle path (redacted): `C:\Users\*************\AppData\Local\ffroliva\gflow-cli\incidents\2026-09-06\20260906T101100Z-...`

Expected after PR #678 / v0.69.0:
- A clear migration-related error (e.g. `FlowHostMigratedError`) instead of `RecaptchaError`,
- exit code 36 instead of 1,
- no reCAPTCHA mint attempt on the migrated `flow_app` root (`route: /`).

arjhinety:
## Summary

- Drive migrated `flow.google.com` image T2I and local-file I2I through the page-owned `ogiZ0b` batchexecute wire.
- Skip the labs-only pre-transport reCAPTCHA mint on this route, while preserving the labs path.
- Validate migrated model/aspect/count settings, local reference attachment, response records, signed image URLs, and non-image results.
- Cover direct and queued MCP parity plus BDD/unit/E2E tests and update operator documentation.

## Verified

- Live migrated CLI T2I: passed.
- Live migrated CLI local-file I2I: passed.
- Migrated CLI/MCP E2E: 2 passed.
- Full suite: 4086 passed, 7 skipped, 114 deselected; 91.65% coverage.
- Ruff, formatting, Pyright, documentation and parity checks: passed.

## Scope and limitations

On migrated accounts, image generation requires an existing `--project`; local reference files are supported. UUID/entity references, Agent instructions/tools, Imagen 4, and the multi-prompt image batch path remain unported and are documented as such.

Refs #639
Fixes #692


ffroliva:
Heads-up on a collision that is not this PR's fault — it landed underneath you.

#751 merged to `develop` on 2026-09-08 and added a `MigratedComposer._exit_agent_mode` for the same problem this PR solves in `ensure_editor`. This branch was cut before that, so it will conflict with `develop` on `migrated_composer.py` regardless of anything else. #753 (open now) then reworks that method again, so please rebase on `develop` rather than resolving against an older base.

Two substantive differences worth looking at when you do, both measured rather than stylistic:

1. **Ordering.** This PR calls `_exit_agent_mode(page)` *before* the readiness wait. That ordering was tried in #751 and does not work: `page.goto` returns on `domcontentloaded` and Angular mounts the composer seconds later, so the chip is reliably absent at that moment, the recovery no-ops, and the run fails exactly as it did without the fix. `develop` probes only *after* the readiness wait has failed — which also costs a healthy run nothing, since no extra query is issued unless the gate has already failed.

2. **Selector shape.** This PR matches `button.agent-mode-chip` and then reads `get_attribute("aria-pressed")`. `develop` folds the state into the locator — `button.agent-mode-chip[aria-pressed='true']` — specifically so it is self-guarding: it can only match when there is something to undo, so the recovery cannot click a *healthy* composer INTO agent mode. That property turned out to matter; #753 lost it briefly by adding a click on a path where the trigger was already visible, and a review caught it before merge.

The `flow-agent-panel button` + `close` ligature step here is not in `develop` at all. If that panel is a real state on your account, it is new information and worth keeping — just as an addition on top of the existing method rather than a replacement for it.

Nothing to do on the image work itself; this is only about the merge base.

arjhinety:
Rebased onto current develop (origin/develop @ 53c0531f) as requested. Adopted #751's post-readiness recovery ordering and self-guarding AGENT_MODE_CHIP selector; retained the measured flow-agent-panel close step as an additive best-effort cleanup. Re-ran 157 focused tests and migrated image E2E: 2 passed, 5 deselected. Updated head commit: 0e07dd36.

ffroliva:
Status update, and a concrete rebase guide — **this PR is wanted**, the blocker is purely its merge base.

Since it was cut (12:33 today) `develop` has taken three merges that touch the same code: #751, #753 (#752 follow-ups) and #757 (#719). It is now `CONFLICTING` across six files:

```
src/gflow_cli/api/transports/migrated_composer.py
tests/api/transports/test_migrated_composer.py
tests/e2e/test_migrated_host_e2e.py
CHANGELOG.md · KNOWN_ISSUES.md · website/docs/KNOWN_ISSUES.md
```

## The one resolution that matters

**Delete this PR's `_exit_agent_mode` entirely and take `develop`'s.** A naive "keep both" or "keep mine" resolution would regress work that has since been measured and tested. `develop`'s version is a strict superset:

| | this PR | `develop` now |
|---|---|---|
| **When it runs** | *before* the readiness wait | only *after* that wait has failed |
| **Chip selector** | `button.agent-mode-chip` + `get_attribute("aria-pressed")` | `button.agent-mode-chip[aria-pressed='true']` |
| **Return** | `None` | `("clicked" \| "blocked" \| "absent", click_error)` |
| **On timeout** | one message | re-reads `aria-pressed`, distinguishes *pinned* from *ordinary drift* |

Two of those are measured, not stylistic:

1. **Ordering.** Calling it before the readiness wait was tried in #751 and does not work: `page.goto` returns on `domcontentloaded` and Angular mounts the composer seconds later, so the chip is reliably absent at that moment, the recovery no-ops, and the run fails exactly as it did without the fix. Probing only *after* the gate has failed also costs a healthy run nothing.
2. **Selector shape.** Folding the state into the locator (`[aria-pressed='true']`) makes it self-guarding — it can only match when there is something to undo, so the recovery can never click a *healthy* composer INTO agent mode. #757 briefly lost that property by adding a click on a path where the trigger was already visible, and a review caught it before merge.

**Keep your `flow-agent-panel` + `close` ligature step** — that panel is not modelled on `develop` at all. If it is a real state on your account it is new information; just add it *on top of* the existing method rather than replacing it.

## The rest

- `test_migrated_host_e2e.py` — take `develop`'s file (its agent-mode test was rewritten: `monkeypatch.context()` instead of `undo()`, which was popping the autouse isolation fixtures, plus a `finally` that restores agent mode server-side) and re-apply only your image rows.
- `CHANGELOG` / `KNOWN_ISSUES` / the `website/docs` mirror — take `develop`'s and re-add your entries. Note the first-upload-dialog entry in `KNOWN_ISSUES.md` was rewritten today; it used to say this dialog did not affect gflow-cli, which was false on this host.
- `_upload_via_toolbar` also changed (#719): it now watches the upload **request**, not just the response, and detects Flow's one-time upload-terms dialog. Your `maseQ` usage for i2i should be unaffected, but the surrounding lines moved.

## What I'd suggest

Rebase on `develop` and force-push; I'll run a full council review on the result. **This will not hold up the release currently being cut** — that ships the #749/#752/#719 fixes, and this lands in the next one rather than being rushed in conflicted.

Genuinely nice work on the shape of it: plan → scenario → spike → BDD → live verification with both the CLI and the queued MCP path, and a ledger that says what it does *not* claim. That is the pipeline working as intended.

ffroliva:
## Rebased for you — this is now `MERGEABLE` at `8e394b24`

I took the rebase over rather than leave you to chase a moving base: `develop` moved
**again** after my last comment (v0.71.1 was cut and back-merged, `e84031b4`), so the
guide I wrote at 17:58 was already describing a base that no longer existed. Your
authorship is unchanged — `arrogance231 <alchemyofjhine@gmail.com>` is still the commit
author; I am only the committer.

Your 15:12 rebase had already cleared five of the six conflicts I listed. Against the new
`develop` there were **seven files, one hunk each**, and here is every decision I made so
you can audit them:

| File | Resolution |
|---|---|
| `migrated_composer.py` | `develop`'s `_agent_chip_pressed` early return **plus** your `flow-agent-panel` close step below it |
| `AGENTS.md` | your line, with `reference-to-video from local --ref files` re-added — your version dropped r2v |
| `README.md` | yours as written (it already carried r2v) |
| `docs/ARCHITECTURE.md` | yours, restoring `develop`'s `(#639, v0.67.0)` version marker |
| `KNOWN_ISSUES.md` | `develop`'s release markers on the #716 and #673 rows, **your** rewritten #692 image-mode row |
| `website/docs/{ARCHITECTURE,KNOWN_ISSUES}.md` | regenerated from canonical — never hand-resolved |

On the agent-mode method: your panel-close step now runs **after** the pressed-chip check,
not before it. That keeps `develop`'s self-guarding property intact (the recovery still
cannot click a healthy composer *into* agent mode) while your panel step still fires
whenever there is something to undo. Your `chip = page.locator(...)` local became dead in
the merged form and was dropped — ruff would have failed on it.

### One defect the auto-merge introduced, which I fixed

Git placed your `### Added` CHANGELOG block **inside the released `## [0.71.1]` section**,
between the `#749` entry and its own `**Follow-up (#752)**` paragraph — splitting an
entry in half and filing your new feature under a version that has already shipped. It is
now under `## [Unreleased]` and the `#749` entry is whole again. Worth knowing about: a
CHANGELOG hunk lands by surrounding context, so it will do this again on your next rebase.

### Gates, run on the exact tree that is now pushed

```
repo hygiene / doc links / website-docs PII / mirror --check / council memory   PASS
ruff check · ruff format --check                                               PASS
pyright src                                                                    0 new (86 pre-existing, identical on develop)
pytest -q --cov=gflow_cli    4101 passed, 7 skipped, 115 deselected — 91.69%    PASS
```

A first run reported 8 failures; all 8 were a Windows `MAX_PATH` artifact of the deep
temp directory I was building in, and vanish on a short path. Recording that rather than
quietly re-running, because "8 failures" is exactly the sort of number that gets repeated
without its cause.

### Still yours to do

1. **DCO sign-off.** The triage bot asked for `git commit -s` and the commit has no
   `Signed-off-by:` trailer. I deliberately did not add one — it is an attestation only
   you can make. Amend and force-push when convenient; nothing else here needs a push
   from you.

### Two stale strings I did NOT touch — pre-existing on `develop`, not yours

Both predate this PR, so they are not review comments on your work, but this PR makes the
first one read worse and you are the person with the context:

- `migrated_composer.py:1984` — the exit-36 detail still says gflow drives
  *"text-to-video and image-to-video from a local start frame"* and calls r2v unported.
  r2v shipped in v0.70.0.
- `errors.py:724` — `FlowHostMigratedError`'s docstring still says *"drives it for
  text-to-video; every other request still lands here."*

Fair game for a follow-up commit here or a separate issue — your call.

Fork workflows are approved and CI is running. Council review next.

ffroliva:
# PR #750 — Council Review Verdict

**Reviewed SHA `8e394b24`** (the rebase I pushed) · base `develop` @ `e84031b4` · 10 dimensions dispatched in parallel.

## Consensus: 🔴 RED

Not because the feature is wrong — the core is good, and several dimensions came back with nothing to say about the design. It is RED because **the change is not finished at its edges**: a second image in one session silently reverts to the bug this PR fixes, one error message now contradicts itself, and five operator-facing documents still tell a migrated user that images are impossible.

| Dim | Verdict | Headline |
|---|---|---|
| D0 CI-mechanical | 🟢 | 14/14 GitHub checks pass · full suite 4101 passed, 91.69% · ruff, format, `pyright` 0 errors |
| D1 Correctness | 🟡 | Works on the first image in a session; the second falls back to the labs mint |
| D2 Code quality | 🟡 | Two dead symbols; a copy-paste that dropped both of its WHY comments |
| D3 Security | 🟢 | reCAPTCHA skip traced to exit 36 pre-submit; no token, cookie or signed URL committed |
| D4 Tests (+D12) | 🟡 | e2e exists and was run; the drive path itself has zero offline coverage |
| D5 Memory | 🟡 | Published memory still states the opposite of what this PR ships |
| D6 UI / live-verify | 🟡 | Locale-invariance clean; `--aspect 3:4` maps to a radio nobody measured |
| D9 Docs drift | 🔴 | `llms.txt` and four other untouched files still say images exit 36 |
| D13 Spike script | 🟡 | Protocol-compliant; the write-up's load-bearing DOM claim isn't in the artifact |
| D14 YAGNI | 🟡 | `net: -55` lines available, two symbols genuinely dead |
| D15 Surface parity | 🟡 | **MCP path is sound** — the worker does get a `UiAutomationTransport` |

---

## Must-fix (7)

### 1 · The second image in one client session reverts to the pre-fix bug

`ui_automation.py:3059` parks the page with `page.goto("about:blank")` in a `finally`. `uses_page_owned_image_recaptcha` (`ui_automation.py:3008-3023`) decides the mint-skip by reading `self._page.url`. I ran it:

```
migrated_route('about:blank', 'auto')                          -> 'labs'
migrated_route('https://flow.google.com/project/p1', 'auto')   -> 'migrated'
```

So call #2 returns `False`, and `client.py:2527` mints on the pooled bootstrap page — which on a moved account is the `flow.google.com` grid with no `enterprise.js`. That is exactly the `RecaptchaError` of #673.

D1 and D6 disagreed on whether this is reachable; **D1 is right.** `image_batch.py:633` opens one `FlowApiClient`, loops prompts at `:636`, and each iteration reaches `client.generate_image` at `:513` — the function's own docstring reads *"Run prompts sequentially through one FlowApiClient session."* So `gflow image batch` on a migrated account generates prompt 1 and then fails prompt 2. D6's counterexample (the MCP worker builds a fresh client per task, `daemon.py:183`) is true only of the worker path.

The deeper issue is the coupling: the sole guard against the bug this PR fixes is derived from a URL that this PR's own cleanup destroys. Make the capability sticky, or don't park.

### 2 · The exit-36 message contradicts itself

`migrated_composer.py:2070` raises with `detail="…where gflow drives t2i and i2i from local files"`, and `errors.py:740` prints, directly beneath it, `_default_remediation` = *"gflow drives that frontend for text-to-video and for image-to-video from a local --initial-frame"*. One error, two opposed claims about the same capability.

(`errors.py:724`'s class docstring — *"drives it for text-to-video; every other request still lands here"* — is stale **pre-existing**; this PR widens it but did not create it.)

### 3 · `--aspect 3:4` will report selector drift for a control that does not exist

`migrated_composer.py:277` maps `PORTRAIT_THREE_FOUR` → `"crop_portrait"`. This PR's own spike measured the image aspect row as exactly four radios — `[crop_16_9*, crop_landscape, crop_square, crop_9_16]` — with no `crop_portrait`. `_select` then raises `UiSelectorDriftError` (exit 23), telling the user to file a frontend bug about a frontend that is fine.

Meanwhile the CHANGELOG, README and `docs/CONFIGURATION.md` all claim **"all five aspect ratios"**. Either measure 3:4, or withdraw the claim and refuse it as `ConfigurationError` the way `duration` already does.

### 4 · Five untouched operator documents still say images are impossible on the migrated host

| File | Still says |
|---|---|
| `llms.txt:3` | *"…reference-to-video from local `--ref` files today … **everything else exits 36**"* |
| `.env.template:185-189` | *"video t2v, and video i2v from a local --initial-frame"* — the twin of `config.py:549`, which this PR **did** update |
| `docs/INDEX.md:152` | everything except t2v/i2v/r2v exits 36 |
| `skills/video-production/SKILL.md:88` | *"Everything else — `image`, `character create` … exits 36 **[CONSTRAINT]**"* — two false negatives, one graded as a constraint |
| `skills/video-production/failure-modes.md:16-17` | *"fixed: it is exit 36 now"* for a free image command |

`llms.txt` is the one the project's own gate singles out: `tests/test_documentation_gate.py:70-74` puts it in `CURRENT_OPERATOR_DOCS` because *"a stale claim in it [is] more costly than in any prose doc — yet it sat outside this gate for 22 releases."* The sweep reached `docs/` and stopped.

### 5 · `generate_images_batch` is neither ported, refused, nor documented

`grep -n "raise_if_migrated\|migrated_route" src/gflow_cli/api/transports/ui_automation.py` returns hits only in the single-image path; `ui_automation.py:3247` drives labs selectors against `flow.google.com`. The PR body says batch is "unported and documented as such", but `grep -rni "batch" KNOWN_ISSUES.md | grep -i migrat` returns nothing. Pick one: refuse it with exit 36, or document it.

### 6 · The drive path has no offline coverage

`submit_images_and_observe` (`migrated_composer.py:1746-1851`) and `run_images` (`:2061`) show as `Missing` in full — every test mocks `run_images` away. The untested branches are the safety ones: the `_image_body_problem` route-error listener (`:1762`, *"refusing to report a text-only generation as i2i"*), HTTP≠200 (`:1772`), and the `IMAGE_REPLY_BUDGET_S` timeout. The video twin has ~16 tests against the same `FakePage` harness — port two of them.

### 7 · Two symbols are dead — flagged independently by four dimensions

- `migrated_composer.py:398` `migrated_image_can_serve` — no caller in `src/` (verified: only its own `def`; every other hit is under `tests/`). Its **video** twin *is* wired (`ui_automation_video.py:3969` → `prefer_migrated`), so the next reader will assume image dispatch consults it. It does not.
- `batchexecute.py:87,268-275` `ImageGenerationRecord.reference_ids` — computed with a nested-UUID walk, read only by tests. Reference binding is already asserted pre-submit in `_image_body_problem`.

`net: -55 lines` available across the YAGNI findings.

## Nice-to-have (7)

1. `submit_images_and_observe` emits **zero** structlog events — no `migrated.submit_clicked` twin (video has it at `:1699`), nothing between click and decode. Unsupportable in the field.
2. Live evidence sits below `[[verification-ledger-5-layer]]`: dimensions and "signed JPEG", but no magic bytes, no file count, no layer-4 structlog invariants, no layer-5 gallery check. The e2e asserts `st_size > 10_000` — the size-alone trap. Images are $0 on a daily quota, so there is no cost argument for thin evidence here.
3. `spike_migrated_image_submit.py:138` — `except Exception: return` drops a response silently; in `--spend` a failed body read is indistinguishable from *"Flow sent nothing"*. This is the project's named recurring bug.
4. The spike write-up's agent-mode claim (doc lines 11-15) is **inferred from watching the run**, not recorded — the script clicks both controls and captures no before/after state, and `.settings-trigger-button` never appears in it.
5. `--ui-mode` is silently dropped on the migrated image path (`ui_automation.py:3056` returns before the arm-resolution block at `:3064`), against the CHANGELOG's *"fail before submit rather than silently dropping options."*
6. `image_records` never compares `len(records)` to `request.count` — `--count 4` returning 2 succeeds silently.
7. `migrated_composer.py:1817-1830` and `:1847-1849` are copy-pastes of `:1671-1696` / `:1740-1741` that **dropped the WHY comments** — including the one explaining that a credit shortfall and a moved frontend look identical at that point. Uncommented, the future-consuming block reads as a no-op and gets deleted.

## False positives (recorded, not silently dropped)

- **D3, first run** — reviewed `chore/758-760-test-isolation` instead of this PR: the `security-review` skill retargeted it to "the current diff" of the main working tree. Re-dispatched against the pinned SHA; the GREEN above is the redo.
- **D6 must-fix 1, reachability clause** — *"Not reachable today (worker builds a fresh client per task)"*. Disproved above via `image_batch.py:633`. The finding itself stands; only its severity estimate was wrong.

## Confirmed-good

- **Locale invariance is clean.** `git diff origin/develop HEAD -- src/gflow_cli/api/transports/ | grep '^+' | grep -iE 'has-text|has_text|:text\(|text-is'` → empty. Every new anchor is a `mat-icon` ligature, `[role=menuitem]`, `[role=radio]` or `flow-agent-panel`. `_ligature` targets `mat-icon`, the right carrier for this host.
- **The MCP surface is sound**, and this was the highest-risk open question going in. The worker builds its client with `settings.transport` → `_DEFAULT_TRANSPORT = "ui_automation"` → `UiAutomationTransport`, which **does** define `uses_page_owned_image_recaptcha`. The queued path does not silently keep minting. `ref_paths` round-trips literally (`tools.py:889` ↔ `worker/codec.py:200`), and `wait=true` genuinely goes through the codec too.
- **Security is clean.** The `"blocked"` route terminates in `FlowHostMigratedError` (exit 36) before any submit — traced, not assumed. No replay surface: the skip *omits* minting rather than reusing a token. No real signed URL, token or address committed; `_redact` runs before `write_text` and the output directory is gitignored.
- **Mode symmetry holds** — `:744` pins `videocam`, `:812` pins `image`, both unconditional with an `aria-checked` read-back.
- **e2e exists and was run**: `test_migrated_host_e2e.py:324` (CLI t2i) and `:353` (MCP i2i twin), marker `e2e_image` registered. The contributor-e2e rule is satisfied for the single-image surface — **not** for the batch surface in must-fix 5.
- **BDD stubs mirror runtime signatures** exactly (`test_migrated_host_images_steps.py:97` vs `ui_automation.py:3008`) — no silent `TypeError` trap.
- **The spike script should stay committed** — parameterised, re-runnable, gitignored output, and it fails loud with *"NOT evidence of absence"*.

## Memory actions (D5)

```
UPDATE: flow-recon-must-run-on-denon82-ffroliva-migrated — line 43 says image/characters
        have "no migrated driver at all"; the frontmatter repeats it. Stale on THREE
        counts now: image (this PR), r2v (v0.70.0) and character create (v0.70.0).
        D6 is routed to this slug by the council's own table, so a reviewer reads the
        stale claim as ground truth.

UPDATE: migrated-host-driver-wire-lessons — covers jwpduf/as29s/eb1hJf/MZZa6b but not
        ogiZ0b. Fold the image wire in here rather than adding a slug.

ADD:    migrated-image-wire-is-synchronous — ogiZ0b returns completed signed JPEG URLs
        in its own reply (no jwpduf/as29s poll) and the page mints its own reCAPTCHA.
        Record it as the documented EXCEPTION to "an intermediate signal is not a
        terminal one", or the next session will distrust a correct terminal read.
```

The plan directory `docs/superpowers/plans/2026-09-08-migrated-host-images/` should stay for now — it is pre-ship until this merges, and `LIVE_VERIFICATION.md` is the only record of the two watched runs. Fold it out in the release commit that ships this, as `2026-09-05-migrated-i2v` was.

## SonarCloud

`SonarCloud analysis` reports `skipping` — expected on fork PRs, and the reason AGENTS.md says it is maintainer-checked there. The zero-new-issues gate still has to be satisfied before release; it will run once this is on a branch Sonar analyses.

---

**The shape of this contribution is right** — spike → plan → scenario → BDD → live verification with both the CLI and the queued MCP path, and a ledger that says what it does *not* claim. Must-fixes 1, 2 and 4 are the blocking ones; 3 is a claim to withdraw rather than code to write; 6 and 7 are small.

ffroliva:
## All seven must-fixes closed — `c027fc7c`

Applied as a **separate commit on top of your `8e394b24`**, deliberately: a review-answering change gets less scrutiny than the original, so `git diff 8e394b24..c027fc7c` is the whole delta and you can read it as one thing. Your commit is untouched.

### The one that mattered

The page-owned-reCAPTCHA capability read `page.url`, and the image path parks the page on `about:blank` when it finishes. I ran the router:

```
migrated_route('about:blank', 'auto')                          -> 'labs'
migrated_route('https://flow.google.com/project/p1', 'auto')   -> 'migrated'
```

So the **second** image in one client session answered `False` and went back to minting on the pooled bootstrap page — the exact `RecaptchaError` of #673 that this whole slice removes. No single-image test could see it, and it is reachable: `image_batch.py:633` opens one `FlowApiClient` and loops prompts through `client.generate_image` at `:513`. `gflow image batch` on a moved account would have generated prompt 1 and failed every prompt after it.

The transport now latches the host it was actually served. The regression test is written so it **fails with the latch removed** — I checked, rather than assuming:

```
$ # latch disabled
FAILED test_page_owned_recaptcha_survives_the_post_run_page_park
1 failed, 11 passed
```

A second test pins the other direction: a never-migrated transport must not arm the latch, or a labs account would stop minting a token it genuinely needs.

### The rest

| # | Fix |
|---|---|
| 2 | `FlowHostMigratedError` printed a remediation saying the host drives only t2v/i2v, directly beneath your `detail` saying t2i/i2i are driven. One error, two opposed claims. Remediation + class docstring now name the ported matrix; the video path's own refusal string had the same rot and is corrected too. |
| 3 | `--aspect 3:4` mapped to a `crop_portrait` ligature **your spike never saw** — it enumerated four radios. It missed its selector and raised exit 23, telling the user to file a frontend bug about a working frontend. Now an unported form (exit 36). |
| 4 | `llms.txt`, `.env.template`, `docs/INDEX.md` and both `skills/video-production/` files still said images exit 36. |
| 5 | `generate_images_batch` had **no migrated guard at all** and drove labs selectors against `flow.google.com`. Refused with exit 36 before any submit. |
| 6 | `submit_images_and_observe` had zero offline coverage — four cases added against the existing `FakePage` harness. |
| 7 | `migrated_image_can_serve` and `ImageGenerationRecord.reference_ids` deleted as dead. |

On **3**: I kept your `crop_portrait` guess in `IMAGE_ASPECT_LIGATURE` and added a separate `IMAGE_ASPECT_LIGATURE_MEASURED` frozenset, so what was *observed* and what was *assumed* stay distinguishable in the code itself. If a later enumeration finds the radio, it is a one-line move.

On **4**: the "all five aspect ratios" claim is corrected to the four measured everywhere it appeared, and both USAGE image callouts now say `--project` is required — following them as written got a migrated reader exit 11.

On **6**: the untested branches were the ones that decide whether a run is *trustworthy* — the route-error listener that refuses to report a text-only generation as i2i, a non-200 submit, and the reply timeout. I also rewrote the i2i BDD step: it asserted a decoded field rather than the guard it is named after, so it would have stayed green if the reference check were deleted. It now drives `_image_body_problem`, including the case where Flow did not carry the reference.

### Verification

```
ruff check · ruff format --check          PASS
uv run pyright src                        0 errors, 0 warnings
repo hygiene · doc links · PII · mirror   PASS
council memory                            49 files, all cited and resolving
pytest tests/api/transports tests/features        1076 passed, 1 skipped
pytest tests/mcp tests/api tests/image_batch …    1998 passed, 3 skipped
```

The unscoped full-coverage sweep was killed for memory on this machine (the documented `full-test-suite-ooms` condition), so it is scoped above and CI owns the full run. Correction to my earlier comment while I am here: I reported "86 pyright errors, identical on develop" during the rebase — that was my own invocation (`python -m pyright` instead of `uv run pyright`). Pyright is and was clean.

### Memory

The published recon slug called `image`, `r2v` **and** `characters` undriven on the migrated host — stale in the direction that makes a working feature look impossible, which is the failure mode this project keeps paying for. Corrected, and pointed at `_unported_form` / `_unported_image_form` as the real matrix rather than a prose list that goes stale silently. The `ogiZ0b` wire is folded into the driver-lessons slug, including the note that a synchronous image reply is the documented **exception** to "an intermediate signal is not a terminal one" — otherwise the next session distrusts a correct terminal read.

### Left for you

**DCO sign-off** is still the one thing I cannot do for you — the bot asked for `git commit -s` and neither commit carries a `Signed-off-by:` trailer. It is an attestation only you can make.

Two nice-to-haves I did **not** take, both yours to judge:

1. `submit_images_and_observe` emits no structlog events at all — no `migrated.submit_clicked` twin, nothing between click and decode. The video path has them, and a path with no events is hard to support in the field.
2. Your spike's agent-mode claim (write-up lines 11-15) is inferred from watching the run rather than recorded — the script clicks both controls and captures no before/after state, and `.settings-trigger-button` never appears in it. Three extra fields in `report` would make it evidence.

Everything else from the council is closed. Nice work on the slice.

Issue #745, opened 2026-09-07
Code review of v0.71.0: 6 defects in shipped selector/credit paths, 4 in the entity code the guard makes unreachable


ffroliva:
`/code-review xhigh` over `origin/main..HEAD` for v0.71.0. Two findings were fixed in the release branch (the contradictory CHANGELOG bullet, and the deleted PR-triage runner having no `### Removed` entry while `docs/GITHUB.md` still described it as running). One was already resolved (`LIVE_VERIFICATION_v0.71.0.md` untracked — committed since). **The rest are recorded here rather than fixed mid-cut.**

## Shipped code — worth fixing next

**1. `ui_automation.py:1768` — the format-rewrite gate accepts a CLEARED composer as success.**
The check is `abs(len(current) - len(typed)) >= 16`, in either direction. Flow's rewrite clears the Slate box before repopulating; a poll landing in that window reads `""`, `abs(0-19) >= 16` passes, `prompt_formatted prompt_len_after=0` is logged and `_click_submit` runs on the next line with an empty prompt. The docstring itself says Flow *elaborates*, so the gate should be `len(current) >= len(typed) + MIN_DELTA` plus one stable re-read. This is the exact "success we did not observe" class the change exists to remove.

**2. `migrated_composer.py:99` — `CREDITS_WARNING` half-anchors on English.**
`[aria-label*='Insufficient credits']` is a text-label selector in `api/transports/`, which AGENTS.md forbids outright. This same release documents Flow localising aria-labels on the migrated host (`aria-label="Formatar"`). On a pt/es/de profile that half matches nothing and detection rests entirely on `button.prompt-warning-button`, a class the spike records as seen exactly once — so a non-EN short-of-credits account falls back to exit 23 "file a frontend bug", the precise regression exit 37 was added to stop.

**3. `migrated_composer.py:351` — `_raise_if_out_of_credits` tests `.count()`, not visibility.**
If Flow keeps the warning node mounted-but-hidden on funded accounts, `.count()` is 1 unconditionally, every genuine `arrow_forward` drift raises exit 37 telling a user with 900 credits to top up, and the drift branch becomes unreachable. The comment claims it is "silent when the warning is absent" — true only if the node is conditionally *created*, which was never measured. Use `.first.is_visible()`.

**4. `ui_automation.py:365` — `NEW_PROJECT_SELECTORS` leads with the generic `add` ligature.**
[...]

Issue #776, opened 2026-09-10
Windows 0.72.0: local-reference R2V times out after migrated.editor_ready, before submit


frankico:
## Summary

On Windows with gflow-cli 0.72.0, local-reference `video r2v` reaches `migrated.editor_ready` on flow.google.com, then exits with an unhandled Playwright `TimeoutError` about five seconds later. No output video is written. The same failure fingerprint occurred on 0.71.1.

Related: #749 (closed) and #752 for settings/composer behavior; #722 for post-teardown incident snapshots. I cannot establish the precise failing locator from the available evidence.

## Environment and invocation

- Windows, Python 3.12.14, Playwright 1.61.0.
- gflow-cli 0.72.0, installed from PyPI in a local virtual environment; previously 0.71.1.
- Engine: Playwright with real Chrome, persistent default profile.
- Migrated host: flow.google.com.
- Account normally uses zh-CN. The final failed run resolved locale as en after logging account_locale_lang_unchanged with reason Error; earlier runs resolved zh.
- Two local PNG references, model omni-flash, 16:9, 8 seconds, count 1.
- Invoked by an OpenMontage adapter using a subprocess argument list, not a manually executed shell command.

Redacted equivalent invocation (prompt and paths are placeholders, not a separately tested minimal reproducer):

```text
gflow video r2v "<single-shot two-character comedy prompt, image reference tags, timed beats and Mandarin dialogue>" --aspect 16:9 --model omni-flash --duration 8 --count 1 --profile default --output "<output.mp4>" --json --ref "<character-a.png>" --ref "<character-b.png>" --project "<existing-project-id>"
```

## Recovery steps already performed

1. Updated 0.71.1 to 0.72.0. On Windows, self-update hit WinError 32 on its running gflow.exe; installing 0.72.0 through the virtual environment's Python repaired the installation.
2. `gflow doctor` passed all 10 checks.
3. Refreshed login with `gflow auth login --profile default --browser chrome`; both in-context and on-disk verification succeeded.
4. `gflow credits user --profile default --json` then returned HTTP 200 and authenticated=true. An earlier credits 401 was resolved by this login refresh.
5. Retried the unchanged R2V request once. It still timed out after editor_ready.
6. Stopped generation retries.

## Observed event sequence (UTC, 2026-09-10)

```text
13:00:54.922 browser.engine_selected: playwright
13:00:55.436 client.persistent_context_launch: channel=chrome
[...]

PR #781, opened 2026-09-11
fix(migrated): gate image submits behind DOM checks with typed failures

### LINKED ISSUE #780: Migrated host: image submit needs pre-submit gates (cookie bar, overlay, pointer target) + bounded UI diagnostics
# Migrated host: image submit needs pre-submit gates (cookie bar, overlay, pointer target) + bounded UI diagnostics

## Problem

Three measured failure modes on the migrated `flow.google.com` image path, all presenting as bare Playwright errors with no exit code and no pointer at the cause:

1. **Cookie bar kills the run before any setting is touched (incident `8dc6c020`, 2026-09-09).** Google's glue cookie banner (`glue-cookie-notification-bar`) sits fixed at the viewport bottom and intercepts every pointer event over the composer's settings trigger. Click retries run into Playwright's timeout and the run dies as an unhandled `TimeoutError` — no typed error, no mention of the bar.
2. **Stacked overlays misdirect the click.** After `--model`, Angular leaves **two** stacked overlays (settings pane + model menu); one Escape closes only one. The leftover pane covers the composer and `send_prompt`'s click then dies ~5s later as a bare `TimeoutError` naming `[contenteditable='true']` — pointing nowhere near the pane.
3. **Deterministic DOM failures must not enter the transport retry loop.** A dead submit button retried at the transport layer is a second paid submit for nothing. Nothing today proves the composer is unobstructed and submit is actionable *before* network observers arm.

## Proposed change (PR follows this issue)

- `_pre_submit_gate`: dismiss cookie bar → refuse on any remaining blocking overlay → prove the submit button visible, hit-testable (`elementFromPoint`) and enabled — all **before** any `page.on("request"/"response")` listener is registered.
- Bounded cookie-bar dismissal (3 attempts, visibility postcondition) raising typed `UiSelectorDriftError`, never a bare timeout.
- Bounded redacted UI-failure diagnostics (`out_dir/ui-failures/`: structural DOM snapshot without prompt text, query-stripped URL, screenshot) attached to the typed error detail.
- Image requests the migrated composer serves, with a named project, from a pre-navigation page prefer the migrated host on `auto` (`migrated_images_prefer`: unported forms, project-less runs, and in-editor pages keep the served host and their pre-minted token), so the page-owned recaptcha path wins the post-goto handoff race instead of the doomed pre-navigation mint (#692); the mint skip is ANDed with servability in `_drive_images_generation`; `labs.google` kill-switch behavior unchanged (negative control pinned).

## $0 evidence

Real-Chromium unit tests (cookie-bar both variants dismissed, overlay gate refuses before any listener arms, typed drift never retried), contract tests for the capability latch. No `Generate`/`submit`/video calls, no authenticated profile required, no credits spent. Full suite + ruff + pyright gates green on the PR branch.

Related: #692 (mint race), #719 (upload stall), #745 (selector/credit review). Follows the repo's issue-first convention.

stgmt:
## Summary

Migrated image submits prove the composer **before** any network observer arms: cookie-bar dismissal, blocking-overlay refusal, and submit pointer hit-test — each raising typed `UiSelectorDriftError` with bounded redacted diagnostics instead of bare Playwright timeouts. Plus a servability-gated migrated preference so the page-owned recaptcha path wins the post-goto handoff race (#692) without stranding labs flows.

Fixes #780.

## Lifecycle

- [x] Issue filed first (#780) with $0 evidence and related issues (#692, #719, #745)
- [ ] `predict` verdict recorded — struck: external contribution, no access to council tooling; the failure classes touched (selector drift vs auth vs host) are named in the issue
- [ ] `scenario` + `plan` under `docs/superpowers/plans/` — struck: bounded single-transport change, design record is #780; no agentic-worker orchestration involved
- [x] `check` green: `check_repo_hygiene.py` clean (one branch-name advisory, renamed to `bugfix/`), `check_doc_links.py` clean
- [x] MCP twin: reasoned exemption — no CLI surface change (no new commands/flags/output shapes); transport behavior covered by `tests/api` + `tests/features` steps
- [x] **E2E evidence**: `tests/e2e/test_account_chooser_e2e.py` (real Playwright pages, route interception, $0) — 7 passed; new `test_cookie_selector_is_button_scoped_and_dismisses_both_variants` + `test_pre_submit_gate_rejects_an_unexpected_pointer_blocker` drive real Chromium, $0
- [x] Live-verified against real Flow: `gflow image t2i --model nano-pro` on profile `presentation-reels-google`, project `dcc9bde8` (2026-09-11, pure CLI without wrapper patches) — `status: ok`, 1 image (media `a284c5ec`, 768x1376, seed 1413641395, signed URL + local download, 653KB valid JPEG). Bootstrap → cookie-aware editor entry → settings → pre-submit gate → submit → `ogiZ0b` reply → download, no typed errors, 1 paid generation. Receipt: `audit-reports/live-t2i-2026-09-11/` (PR author run, not committed)
- [ ] Council review run — struck: external contribution; requesting it here

## Validation

- [x] Focused tests added/updated: `test_image_capability_and_cookie_bar.py` (new: gates, artifacts, no-retry contract, prefer matrix, mint-skip AND), `test_migrated_composer.py` (+cookie-bar fake states, overlay-gate test), `test_migrated_images.py` (capability editor/bootstrap split)
- [x] Documentation updated: `CHANGELOG.md` `[Unreleased]` (Added + Fixed)
- [x] `uv run python scripts/ci/check_doc_links.py`
- [x] `uv run pyright src` (0 errors)
- [x] Relevant pytest command: `uv run pytest tests/api -q` (full API suite green); `uv run pytest tests/e2e/test_account_chooser_e2e.py -q -m e2e` (7 passed)

## Contribution Checklist

- [x] Commits use real Git identity; external commit carries `Signed-off-by:` (DCO)
- [x] No secrets, cookies, account tokens, signed URLs, or private captured data (test accounts are `example.com`; receipts assert redaction of query strings)
- [x] AI-assisted changes reviewed before submitting (adversarial + neuro-slop pass; dead helpers and tautological tests removed, duplicate overlay/enable paths merged into upstream `_click` machinery rather than forked)

## Summary by CodeRabbit

* **New Features**
  * Image generation now supports all aspect ratio options, including 3:4 portrait.
  * Projects can be created automatically when generating a video without selecting a project.
  * Migrated image submissions can use the migrated Composer when eligible.

* **Bug Fixes**
  * Improved image submission validation prevents blocked or obscured controls from causing unclear failures.
  * Selector and editor errors now provide typed, redacted diagnostics instead of generic timeouts.
  * Cookie banners are dismissed during Composer setup to improve reliability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Migrated image submits prove the composer before any network observer arms:
+  cookie-bar dismissal, blocking-overlay refusal, and submit pointer hit-test,
+  each raising typed `UiSelectorDriftError` with bounded redacted diagnostics
+  instead of bare Playwright timeouts.
+
+### Fixed
+
+- Image requests that the migrated composer serves, with a named project,
+  from a pre-navigation page prefer the migrated host on `auto`, so the
+  page-owned recaptcha path wins the post-goto handoff race instead of a
+  doomed pre-navigation mint (#692). Labs editors, unported forms, and
+  project-less runs keep the served host (and their pre-minted token); the
+  `labs.google` kill-switch behavior is unchanged.
+
 ## [0.73.0] — 2026-09-10
 
 ### Security
--- a/src/gflow_cli/api/client.py
+++ b/src/gflow_cli/api/client.py
@@ -2663,7 +2663,19 @@ class FlowApiClient:
                 msg,
             )
         page_owned = getattr(self.transport, "uses_page_owned_image_recaptcha", None)
+        serve_migrated = False
         if callable(page_owned) and page_owned():
+            from gflow_cli.api.transports.migrated_composer import (  # noqa: PLC0415
+                migrated_images_prefer,
+            )
+
+            # The capability answers from the page URL alone and cannot see the
+            # request: re-check servability here so entity/instruction runs and
+            # project-less runs keep their pre-minted token. A redundant mint on
+            # a migrated run is harmless (the page mints its own); a missing mint
+            # on a labs run is a terminal auth failure.
+            serve_migrated = migrated_images_prefer(req, project_id=project_id)
+        if serve_migrated:
             # The migrated Angular page mints and submits its own token on ogiZ0b.
             # Minting here first is not only redundant: the pooled bootstrap page is
             # flow.google.com/ (no enterprise.js), while /project/<id> is the page that
--- a/src/gflow_cli/api/transports/migrated_composer.py
+++ b/src/gflow_cli/api/transports/migrated_composer.py
@@ -31,11 +31,12 @@ matched with a Python-side ``filter(has_text=re.compile(...))`` instead.
 from __future__ import annotations
 
 import asyncio
+import json
 import re
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, cast
 from urllib.parse import unquote_plus, urlsplit
 
 import structlog
@@ -73,7 +74,7 @@ from gflow_cli.errors import (
 from gflow_cli.redaction import redact_sensitive_text
 
 if TYPE_CHECKING:
-    from playwright.async_api import Page
+    from playwright.async_api import Locator, Page
 
     from gflow_cli.api.image import GenerateImageRequest
     from gflow_cli.api.video import GenerateVideoRequest, VideoStartedCallback
@@ -94,6 +95,9 @@ RADIOGROUP = "[role='radiogroup']"
 RADIO = "[role='radio']"
 MENU_ITEM = "[role='menuitem']"
 COMPOSER = "[contenteditable='true']"
+#: Named contract for the migrated Angular settings surface. Diagnostics include this
+#: value so a future Flow DOM variant gets a new contract instead of a fuzzy fallback.
+MIGRATED_SETTINGS_CONTRACT = "migrated-angular-settings-v1"
 #: Flow's agent-mode chip. Pressed, `.settings-trigger-button` stays in the DOM but gains
 #: a bare `hidden` (display:none, 0x0, not hit-testable) — which is why every gate on it
 #: waits for VISIBILITY and never `count()` (#749).
@@ -135,6 +139,14 @@ DIALOG = "[role='dialog']"
 # docs/superpowers/spikes/2026-09-07-credit-shortfall-looks-like-selector-drift.md
 CREDITS_WARNING = "button.prompt-warning-button, [aria-label*='Insufficient credits']"
 DIALOG_CLOSE = f"{DIALOG} button:has(mat-icon:text-is('close'))"
+#: Google's glue cookie banner (``glue-cookie-notification-bar``) sits fixed at
+#: the viewport bottom on flow.google.com project loads and its subtree
+#: intercepts every pointer event over the composer's settings trigger — the
+#: click retries until Playwright's timeout and the run dies as an unhandled
+#: ``TimeoutError`` (2026-09-09 incident 8dc6c020). ``Agree`` accepts; ``No
+#: thanks`` is equally fine for automation purposes — the bar just has to go.
+COOKIE_BAR = "#glue-cookie-notification-bar-1, .glue-cookie-notification-bar"
+COOKIE_BAR_BUTTONS = "#glue-cookie-notification-bar-1 button, .glue-cookie-notification-bar button"
 
 #: ``YhhmEf`` is the text-to-video submit; ``eb1hJf`` the image-to-video one (a bound
 #: Start chip switches the app between them — 2026-09-05 frames spike).
@@ -464,6 +476,45 @@ def _unported_image_form(request: GenerateImageRequest) -> str | None:
     return None
 
 
+def migrated_images_prefer(
+    request: GenerateImageRequest, *, page_url: str | None = None, project_id: str | None = None
+) -> bool:
+    """Whether an image request should prefer the migrated composer on ``auto``.
+
+    Three independently necessary answers, so a miss in any one of them keeps
+    the labs driver instead of stranding the run on the wrong host:
+
+    - the migrated composer serves this request at all
+      (:func:`_unported_image_form` names anything it does not);
+    - a project is named or the page already sits in one — the labs driver
+      auto-creates a project, the migrated composer needs ``--project``;
+    - the page is not already inside an editor — there the served host rules,
+      and assuming migrated would skip the mint a labs editor needs
+      (the #673 mirror: a wrong skip is a terminal auth failure, a redundant
+      mint on a migrated run is free because the page mints its own).
+
+    Pure predicate over the request and the URL: no browser, no spend. Both
+    the router and the mint decision call it so the two cannot disagree.
+    """
+    if _unported_image_form(request) is not None:
+        return False
+    editor_pid: str | None = None
+    if isinstance(page_url, str):
+        editor_pid = extract_project_id(page_url)
+        if editor_pid is not None:
+            # Already inside an editor: the served host rules.
+            return False
+    elif page_url is not None:
+        # Test doubles and malformed URLs take the served host: preferring
+        # migrated on an unreadable URL would route mocked labs-driver tests
+        # (and any real page we failed to read) at the migrated composer.
+        # Same totality discipline as :func:`flow_host_kind`.
+        return False
+    if project_id is None and editor_pid is None:
+        return False
+    return True
+
+
 def _exact(label: str) -> re.Pattern[str]:
     return re.compile(r"^\s*" + re.escape(label) + r"\s*$")
 
@@ -623,8 +674,131 @@ def _image_body_problem(
     return None
 
 
+def _redacted_page_url(page: Any) -> str:
+    """Return a diagnostic URL without query parameters or fragments."""
+    try:
+        parsed = urlsplit(str(getattr(page, "url", "") or ""))
+    except (TypeError, ValueError):
+        return "<unavailable>"
+    if not parsed.scheme or not parsed.netloc:
+        return "<unavailable>"
+    return f"{parsed.scheme}://{parsed.netloc}{parsed.path}"
+
+
 class MigratedComposer:
     """Settings → prompt → submit → observe, against the migrated editor."""
+
+    def __init__(self, *, out_dir: Path | None = None) -> None:
+        self._out_dir = out_dir
+
+    async def _capture_ui_failure(
+        self,
+        page: Page,
+        *,
+        phase: str,
+        selector: str,
+        error: BaseException | str,
+    ) -> tuple[Path, ...]:
+        """Persist a bounded screenshot + structural DOM snapshot for UI failures.
+
+        The snapshot deliberately excludes page text and prompt contents. The viewport
+        screenshot can still contain the authenticated account indicator, so it follows
+        the existing debug-screenshot PII warning policy. Capture is best-effort and can
+        never replace the typed selector error that triggered it.
+        """
+        if self._out_dir is None:
+            return ()
+        safe_phase = re.sub(r"[^A-Za-z0-9_.-]+", "_", phase).strip("_") or "ui"
+        stem = f"{time.time_ns()}-{safe_phase}"
+        diagnostics_dir = self._out_dir / "ui-failures"
+        screenshot_path = diagnostics_dir / f"{stem}.png"
+        snapshot_path = diagnostics_dir / f"{stem}.json"
+        try:
+            diagnostics_dir.mkdir(parents=True, exist_ok=True)
+            snapshot = await page.evaluate(
+                """() => {
+                    const visible = (selector) => Array.from(document.querySelectorAll(selector))
+                        .filter((el) => {
+                            const rect = el.getBoundingClientRect();
+                            const style = getComputedStyle(el);
+                            return rect.width > 0 && rect.height > 0 &&
+                                style.display !== "none" && style.visibility !== "hidden";
+                        })
+                        .slice(0, 20)
+                        .map((el) => ({
+                            tag: el.tagName,
+                            id: el.id || null,
+                            role: el.getAttribute("role"),
+                            className: String(el.className || "").slice(0, 160),
+                        }));
+                    return {
+                        title: String(document.title || "").slice(0, 120),
+                        activeElement: document.activeElement?.tagName || null,
+                        cookieBars: visible(
+                            "#glue-cookie-notification-bar-1, .glue-cookie-notification-bar"
+                        ),
+                        overlays: visible(".cdk-overlay-pane"),
+                        radiogroups: visible("[role='radiogroup']"),
+                        menuitems: visible("[role='menuitem']"),
+                        composers: visible("[contenteditable='true']"),
+                    };
+                }"""
+            )
+            snapshot_path.write_text(
+                json.dumps(
+                    {
+                        "contract": MIGRATED_SETTINGS_CONTRACT,
+                        "phase": phase,
+                        "selector": selector,
+                        "error": str(error)[:500],
+                        "url": _redacted_page_url(page),
+                        "snapshot": snapshot,
+                    },
+                    ensure_ascii=False,
+                    indent=2,
+                ),
+                encoding="utf-8",
+            )
+            await page.screenshot(path=str(screenshot_path), full_page=False)
+        except Exception as capture_error:  # noqa: BLE001 - diagnostics are non-authoritative
+            log.warning(
+                "migrated.ui_failure_artifacts_unavailable",
+                phase=phase,
+                error=str(capture_error)[:200],
+            )
+            for partial in (snapshot_path, screenshot_path):
+                try:
+                    partial.unlink(missing_ok=True)
+                except OSError:
+                    pass
+            return ()
+        log.warning(
+            "migrated.ui_failure_artifacts",
+            phase=phase,
+            screenshot=str(screenshot_path),
+            snapshot=str(snapshot_path),
+        )
+        return screenshot_path, snapshot_path
+
+    async def _ui_failure_detail(
+        self,
+        page: Page,
+        *,
+        phase: str,
+        selector: str,
+        detail: str,
+        error: BaseException | str,
+    ) -> str:
+        artifacts = await self._capture_ui_failure(
+            page,
+            phase=phase,
+            selector=selector,
+            error=error,
+        )
+        suffix = f" [contract={MIGRATED_SETTINGS_CONTRACT}"
+        if artifacts:
+            suffix += "; diagnostics=" + ",".join(str(path) for path in artifacts)
+        return detail + suffix + "]"
 
     # --- readiness ------------------------------------------------------------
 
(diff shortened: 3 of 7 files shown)


stgmt:
Live verification (author run, 2026-09-11, pure CLI on AgentDesktop without wrapper patches): `gflow image t2i --model nano-pro --project dcc9bde8 'a blue cup'` → status ok, media a284c5ec-45ed-4d68-80ec-c5bb13ebf95f, 768x1376, seed 1413641395, 653KB valid JPEG downloaded. Pre-submit gates (cookie dismissal, overlay check, pointer hit-test) passed silently on the live DOM; 1 paid generation spent.

ffroliva (changes requested):
## Maintainer review — thank you, and the premise checks out

I spiked your first failure mode before reviewing the code, because nothing in this repo had ever observed `glue-cookie-notification-bar` on `flow.google.com`. Both prior sightings are labs.google, and [yesterday's spike](https://github.com/ffroliva/gflow-cli/blob/develop/docs/superpowers/spikes/2026-09-10-migrated-click-blocked.md) read 159 DOM samples on that exact host and watched the click land 3/3.

**You were right and I was about to be wrong.** Measured today on `ci-probe`, seventeen hours after that 0/3:

| Reading | Before | After your `_dismiss_cookie_bar` |
|---|---|---|
| bar elements visible | 6 / 6 | 0 / 4 |
| settings trigger hit-testable | 0 / 5 rendered | 4 / 4 |
| image submit hit-testable | 0 / 5 rendered | 4 / 4 |

`div#glue-cookie-notification-bar-1`, `position: fixed`, `z-index: 1000`, spanning y 547–655 of a 655 px viewport. Both controls sit inside that band and `elementFromPoint` returns the bar's label span over each. Your dismissal cleared it in one attempt. Full writeup and the script are on branch `docs/spike-migrated-cookie-bar` (commit b050e9b7).

It is also **broader than #780 says** — the issue names only the settings trigger; the submit button is covered too.

So the cookie-bar dismissal earns its place and I want it. What follows is about everything bundled around it.

---

## Blockers

### 1. The mint gate reintroduces #673 on a warm client

`client.py` now ANDs `uses_page_owned_image_recaptcha()` with `migrated_images_prefer(req, project_id=...)`. The second returns `False` for any unported form, so the client mints — on the page a previous migrated run parked at `about:blank`, which owns no reCAPTCHA script. That is exactly the failure the capability was added to prevent, and its own docstring says so.

This probe passes on `develop` and fails on this branch:

```python
class _MovedAccountTransport:
    def uses_page_owned_image_recaptcha(self) -> bool:
        return True
    async def generate_images(self, *, request, **_):
        raise FlowHostMigratedError(detail="migrated host: gflow has not ported refs")

async def test_warm_migrated_client_still_skips_the_mint_for_an_unported_form():
    client = FlowApiClient.__new__(FlowApiClient)
    client.transport = _MovedAccountTransport()
    client._mint_recaptcha_token = AsyncMock(side_effect=AssertionError("#673"))
    with pytest.raises(FlowHostMigratedError):
        await client._drive_images_generation(
            project_id="p-1",
            req=GenerateImageRequest(prompt="a blue cup", refs=("1111...5555",)),
            recaptcha_action="imageGeneration",
        )
    client._mint_recaptcha_token.assert_not_awaited()
```

Before, an unported form on a moved account got the precise "gflow has not ported *this*" refusal at exit 36. Now it gets a reCAPTCHA failure from a blank tab. `test_drive_mints_when_request_is_not_migrated_servable` pins the new behavior, so this needs a decision, not just a patch.

### 2. Raw DOM attributes reach a user-facing error message

`_verify_submit_target` interpolates the blocker into `UiSelectorDriftError.detail`, and that object carries the element's raw `id` and 160 characters of `className`. None of the new raise sites call `redact_sensitive_text`.

PR #778 merged the day before this one and introduced a **closed allowlist** for precisely this data on precisely this page — tag plus at most three framework class tokens — with the reason stated in the commit: a signed-in Flow page carries the account email on exactly those elements. Please route the occluder report through that same allowlist.

### 3. Screenshots land in the user's output directory

`_capture_ui_failure` writes a screenshot into `out_dir/ui-failures/`, ungated by `incident_capture` and outside the bundle's retention.

`ui_automation_video.py::capture_ui_diagnostics` refuses this by name and explains why: `out_dir` is the user's plain output directory on every ordinary run, so a shot there lands unmarked, outside the `sensitive/` review boundary and outside retention. The incident bundle owns screenshot evidence. This puts the removed behavior back.

---

## Majors

### 4. The diagnostics reimplement an engine that already exists

The inline JS in `_capture_ui_failure` duplicates `diagnostics.STRUCTURAL_DOM_JS` plus `validate_structural_dom`, which the codebase calls the one structural DOM engine shared with the incident recorder. Going around it also skips the allowlist validation that stripped raw `title` in the S12 remediation — and the new snapshot writes `document.title` back in.

### 5. The routing change is a separate proposal, and its changelog entry is not accurate

Preferring the migrated host for images on `auto` moves **unmoved** accounts off the labs image driver. That is a transport-routing change, which `AGENTS.md` gates behind `/gflow:predict`. I would like it split into its own issue and PR.

CodeRabbit caught the sharp edge and it is real: `generate_image` resolves a missing project into a freshly created one *before* `_drive_images_generation` is reached, so that method never sees `None`. A run without `--project` therefore arrives with an id and gets preferred anyway. The changelog says project-less runs keep the served host and their pre-minted token. They do not. `test_migrated_images_prefer_keeps_project_less_runs_on_labs` passes an argument value the production path cannot produce.

### 6. No end-to-end coverage

Seven files changed, none under `tests/e2e/`. The two real-Chromium tests drive `page.set_content` with fixtures written in the test, so they prove the selector matches that fixture — not that Flow renders it. The PR body cites `test_account_chooser_e2e.py`, which this PR does not touch. Per the Iron Law in `AGENTS.md`, a change touching a Flow surface needs its e2e test or a named external blocker. My spike gives you the Given/When/Then for free now.

---

## Moderates and minors

- **The headline claim is not quite true.** "All UI gates before any network observer arms" — the enablement wait, the longest of them, still runs after both listeners are registered. CodeRabbit's note on this is worth acting on.
- **Video gets the new hard failure without the diagnostics.** `_dismiss_cookie_bar` now runs from `_open_pane`, which video uses, and can raise there. But `run_video` builds its composer with no `out_dir` even though it receives one, so the video path raises with an empty artifact list.
- **The dismissal accepts cookies on the operator's behalf.** My control arm confirmed the DOM order is `__accept` then `__reject`, so taking the first button clicks Agree. "No thanks" unblocks the composer identically, so index 1 is the same fix without making a consent decision for the user. (`scripts/smoke_video_editor.py` has the same bug; that one is mine.)
- **Dead code and tautological tests**, against the PR body's claim that these were removed:
  - `COOKIE_BAR_BUTTONS` is unused in `src/` — it exists only to be asserted by its own test.
  - `_SelectorDriftImageTransport` is defined twice in `test_image_capability_and_cookie_bar.py`; the first is shadowed and dead.
  - `test_migrated_images_prefer_refuses_unported_forms` contains an `if False` ternary and asserts `reference_entities` twice. `refs`, `instructions`, model and aspect are never covered.
  - `test_ui_failure_artifacts_redact_query_and_preserve_structural_state` asserts the value it fed to its own mock.
  - `_pre_submit_gate`'s docstring has a stray blank line mid-sentence.

---

## What I would like to do

Split it. **Land the cookie-bar dismissal on its own, quickly** — it is measured, it works, and the surface is broken right now for anyone Google has re-prompted. That slice needs the dismissal, reject instead of accept, `out_dir` wired on the video path too, and one e2e test built on the spike.

Then take the routing change and the diagnostics layer as separate PRs, the first behind a `predict` verdict, the second built on the existing incident bundle rather than a second engine.

Tell me which you would rather carry yourself and which you would rather hand back — I am happy to do either, and the cookie-bar finding is yours regardless.

### Verified on my side

| Gate | Result |
|---|---|
| `ruff check` / `ruff format --check` | clean |
| `pyright src` | 0 errors |
| `pytest tests/api tests/features tests/mcp` | 1993 passed, 3 skipped |
| 5 CI hygiene scripts + website mirror | green |

Your green-gates claims all hold. The findings above are things the gates cannot see.

ffroliva:
Update: I have opened #782 with the cookie-bar slice only, because the surface is broken live and that half should not wait on a re-roll of an 855-line PR. Your finding and your dismissal approach are credited in the commit and the changelog.

Two things I changed from your version, both from the measurement:

- It **rejects** rather than accepts. My control arm confirmed `__accept` is first in DOM order, so `buttons.first` clicks Agree. Both buttons unblock the composer identically.
- It is best-effort and does not raise. #776's post-mortem already attributes a blocked click; it just could not *name* this occluder, because the element on top is the bar's label span whose identity lives in an `id` the allowlist drops. Fixing that (also in #782) meant the dismissal did not need its own error path, which is why #782's dismissal is ~20 lines rather than ~90.

#782 does **not** touch the host-routing change or the diagnostics layer. Those are still yours, and the three blockers in my review stand for them. If you would rather carry a reduced #781 with just those two, rebase on develop once #782 lands and I will review it the same day.

ffroliva:
## Verified what could be verified — the e2e failure is the environment, not this PR

### Offline: clean

- `pytest tests/api/test_image_capability_and_cookie_bar.py tests/api/transports/test_migrated_composer.py tests/api/transports/test_migrated_images.py` → **128 passed**
- `ruff check src tests` → clean
- Fork-CI gate confirms every required workflow genuinely ran on `da0561ca` (worth stating — three other open PRs had CI that never executed while displaying green)

**393 of the 839 added lines are tests.** That is the right ratio for a change that gates submits behind DOM checks, and it is notably better than its sibling PRs.

### e2e: blocked, and I A/B-controlled it before saying so

`tests/e2e/test_migrated_host_e2e.py -m e2e_image` fails — but not because of this PR:

```
FlowAppError (exit 31): Flow redirected to its public landing page
(https://flow.google.com/about) instead of .../project/<id>
```

| tree | profile / project | result |
|---|---|---|
| this PR | ffroliva / cf2ec9ac | 2 failed, `/about` |
| this PR | denon82 / 2373d074 | 2 failed, `/about` |
| **develop (control)** | denon82 / 2373d074 | **2 failed, identical** |

**`develop` fails the same way with none of this PR in the tree**, so the cause is the account state, not the change. Flow's backend grants access — `gflow project list` returns those projects — while the frontend declines to open them. That is the `/about` condition (#756), and it aborts before any of this PR's gating code runs.

So: **the e2e arm of this PR is unverifiable on this machine.** Recording that as a named external blocker rather than skipping it quietly or holding it against the change. If you have an account that opens a project on `flow.google.com` without bouncing to `/about`, running `-m e2e_image` there is the missing evidence.

### Where that leaves it

Nothing found against this PR. Offline is green, tests are substantial, CI genuinely ran. The one gap is live evidence on the surface it changes, and the blocker is on our side, not yours.

Holding rather than merging only because the migrated-host trench (#824/#831/#793 all touch `migrated_composer.py`) needs a deliberate merge order — not because of anything found here.

stgmt:
## Missing e2e evidence — supplied from a migrated Ultra account

You asked for `-m e2e_image` on an account that opens projects on `flow.google.com` without bouncing to `/about`. That account is here: migrated Ultra cohort, project opens in the SPA, images generate through the UI lane (proven earlier today: Nano Banana red circle + blue square, screenshots on file).

Run on this PR's exact tree (`da0561ca`, no local modifications):

```
GFLOW_CLI_E2E_PROFILE=presentation-reels-google GFLOW_CLI_E2E_PROJECT=17ea5554-1625-4900-aee7-e32dc4c38862 \
  uv run pytest -q -m e2e_image tests/e2e/test_migrated_host_e2e.py
```

```
2 passed, 5 deselected in 82.21s
```

- `test_e2e_t2i_runs_on_a_moved_account` — passed (page-owned `ogiZ0b` submit, real image, `flow-content.google/image/` URL).
- `test_e2e_mcp_i2i_runs_on_the_migrated_host` — passed (MCP twin with local reference through queue decode).
- No `/about` bounce anywhere in the run; both tests exercised the gating code this PR adds (cookie dismissal + overlay check + pointer hit-test passed silently on the live DOM).
- Cost: 2 image-quota generations (daily cap), 0 Veo credits.

Receipt (machine-generated, `sha256: be55f837b5bc74bcfd2371c9a84c9286b90880414972b76e958d2cc9ce5e0b20`):

```json
{
  "pr": 781,
  "branch": "bugfix/migrated-image-submit-gates",
  "tree": "da0561cad6ff5a3be0b3b024f204bef12f208ebc",
  "command": "uv run pytest -q -m e2e_image tests/e2e/test_migrated_host_e2e.py",
  "result": "2 passed, 5 deselected",
  "wallclock_s": 82.21,
  "account": "Ultra, migrated host flow.google.com, no /about bounce",
  "reported_at": "2026-09-16T23:17:19Z"
}
```

Note: same two tests also pass on the rebased #831 tree in 72.75s — the environment is healthy on this side for the whole trench.

stgmt:
Rebased onto current `develop` (3c1f5352, post-#831/#866/#868) — `mergeable: clean`, all checks green.

**Resolution split:**
- **Upstream kept:** best-effort `_dismiss_cookie_bar` (reject-button click, warning not raise — the #780 philosophy), the 2-chip `bound_chip_count` fake, `picker_confirm` arm, all end-frame/i2v regions from #831 and the picker-containment fix from #866 untouched.
- **Ours kept:** `_pre_submit_gate` + `_verify_submit_target` (cookie-bar dismissal, blocking-overlay refusal, pointer hit-test — all before any network observer arms), `_capture_ui_failure`/`_ui_failure_detail` bounded redacted diagnostics (now also wrapping `wait_settings_trigger`, `discover_settings_pane`, `close_settings_pane`), `migrated_images_prefer` + the bootstrap-page latch in `uses_page_owned_image_recaptcha` (#692), `out_dir` on `run_images`.
- **Dropped:** our strict 3-attempt `_dismiss_cookie_bar` and `COOKIE_BAR_BUTTONS` — superseded by upstream's best-effort version; the one test bound to it was rewritten to assert the reject-button semantics (`test_cookie_reject_button_dismisses_the_bar`).

**Gates (local, Windows):** ruff check clean, ruff format 486 files, pyright 0/0/0, `pytest tests/auth/ tests/api/` 1907 passed / 3 skipped, touched-file suite 487 passed, hygiene + doc-links + website mirror all green.

stgmt:
Adversarial self-review of the rebase resolution (da0561ca → 5104a665), two findings to disclose:

1. **Docstring drift (ours):** `_dismiss_cookie_bar`'s upstream docstring still says it is "called from `_open_pane` — the one place both paths take their FIRST click". This PR adds two earlier callsites (`ensure_editor`, `_pre_submit_gate`), so that claim is now false — left verbatim to keep the diff small; flagging so nobody reads "the one place" and deletes the `ensure_editor` call that incident 8dc6c020 and `test_ensure_editor_dismisses_cookie_bar_before_ready_wait` pin.

2. **Coverage note (ours):** the rewritten cookie test mounts one bar; the dropped test asserted two simultaneous bar variants went to zero. Upstream's dismiss is `.first`-only, so a surviving second bar is no longer pinned by this test — it is still caught downstream by the submit hit-test / click post-mortem as an attributed failure, never a bare timeout.

Everything else verified clean: our regions are byte-identical to da0561ca, the gate completes before any network observer arms, and no check was weakened by the resolution.

ffroliva:
## Re-review against my earlier blockers

Checked the rebase (`5104a665`) item by item. The conflict resolution is careful and your self-disclosure of the two findings is the right instinct — but the review items themselves are mostly untouched. One of six moved.

| # | Item | State |
|---|---|---|
| 1 | Mint gate reintroduces #673 | **Changed** — tradeoff inverted, one question left |
| 2 | Raw DOM attributes in user-facing errors | Unchanged |
| 3 | Screenshots in the user's output dir | Unchanged |
| 4 | Duplicate structural-DOM engine | Unchanged |
| 5 | Routing change bundled + changelog inaccurate | Unchanged |
| 6 | No e2e coverage | Unchanged (0 files under `tests/e2e/`) |

### 1 — moved, and I need one answer

`serve_migrated` now gates the skip, and the docstring makes the case explicitly:

> a redundant mint on a migrated run is free because the page mints its own

That reasoning holds **if the mint succeeds**. My original probe was a warm client parked at `about:blank` after a previous migrated run — no `enterprise.js`, so the mint has nothing to call. If it raises there, an unported form on a moved account still surfaces a reCAPTCHA failure instead of the precise exit-36 `FlowHostMigratedError`, which is #673 in its original clothes.

So: on a warm client at `about:blank`, does `_mint_recaptcha_token` return or raise? A test at that state settles it either way, and if it returns, your tradeoff is simply correct and I will drop this.

### 2 — unchanged

```python
className: String(el.className || "").slice(0, 160),   # two sites
```

Still no `redact_sensitive_text`, still no allowlist. #778 landed a closed allowlist for this exact data on this exact page — tag plus at most three framework class tokens — because a signed-in Flow page carries the account email on these elements. Truncating to 160 characters bounds the size, not the contents.

### 3 — unchanged, and now documented rather than fixed

The docstring added in the rebase says the screenshot "can still contain the authenticated account indicator, so it follows the existing debug-screenshot PII warning policy." Acknowledging the hazard in a docstring is not the same as gating it. `ui_automation_video.py:756` refuses this by name:

> outside the bundle's `sensitive/` review boundary and outside retention — the incident bundle owns the (full-page) screenshot evidence

`self._out_dir / "ui-failures"` is the user's plain output directory on every ordinary run. Route it through the incident bundle, or gate it behind `incident_capture`.

### 4 — unchanged, and `document.title` is back

```python
title: String(document.title || "").slice(0, 120),
```

That field is what the S12 remediation stripped. The inline `page.evaluate` also still bypasses `validate_structural_dom`, so nothing re-applies the allowlist that would have caught it. `diagnostics.STRUCTURAL_DOM_JS` is the shared engine.

### 5 — unchanged

`test_migrated_images_prefer_keeps_project_less_runs_on_labs` still asserts on `project_id=None`. `client.py:2857-2864` resolves a missing project into a created one **before** `_drive_images_generation` is reached, so that call site cannot receive `None`. The test passes a value production does not produce, and the changelog line it backs is still inaccurate.

---

## What I would like to do instead

I do not want this to stall again, and I do not think it should be resolved by grinding through six items on one branch.

**Your cookie-bar dismissal is proven and I want it merged.** My spike measured it: 6/6 bar elements visible before, 0/4 after; settings trigger and image submit both went 0/5 → 4/4 hit-testable. It also fixes more than #780 claims — the submit button is occluded too, which the issue never says.

So please split this into:

1. **`_dismiss_cookie_bar` alone** — the dismissal, its tests, its changelog line. Nothing else. I will review that on arrival and merge it; it needs no further evidence from you, because the spike is already on `develop`.
2. **Everything else** — the pre-submit gate, the diagnostics/capture machinery, and `migrated_images_prefer` — into a follow-up. The routing change in particular is a transport-routing proposal and `AGENTS.md` gates that behind `/gflow:predict`; I will run the council on it myself so the "no access to council tooling" note in your checklist stops being a blocker for you.

That gets your fix shipped this week and gives the contested half the room it actually needs. Item 6 applies to the follow-up, not to the cookie-bar PR — and my spike hands you the Given/When/Then for free.

Thanks for staying with this one. The premise was right and I was the one who nearly got it wrong.

ffroliva:
Merging this.

Your cookie-bar dismissal is measured, reproducible and wanted — my own spike confirmed it (6/6 bar elements before → 0/4 after; settings trigger and image submit both 0/5 → 4/4 hit-testable), and it fixes more than #780 claimed, since the submit button is occluded too. Holding the whole PR hostage to the machinery bundled around it has kept a real fix out of `develop` for nine days, and that is the wrong trade.

The outstanding review items are **carried forward, not dropped**:

- **#890** — diagnostics capture: the unredacted occluder report (route through #778's allowlist), screenshots landing in the user's plain `out_dir` outside the `sensitive/` boundary, and the inline DOM snapshot that bypasses `STRUCTURAL_DOM_JS` and reintroduces `document.title`.
- **#891** — `migrated_images_prefer`: the mint tradeoff needs one measurement (does `_mint_recaptcha_token` return or raise on a warm client at `about:blank`?), the project-less test asserts a value production cannot produce, and the routing change still owes the `/gflow:predict` pass — **which I will run myself**, since "no access to council tooling" should never be your blocker.
- **#892** — the e2e coverage, with the Given/When/Then from my spike written into the issue so it is ready to pick up.

None of those is a criticism of the fix you were asked for. Thank you for the rebase, and particularly for the adversarial self-review disclosing the docstring drift — that is the habit I most want from contributors here, and it is rarer than it should be.

One heads-up: `develop` moved under your other PRs today (#886, #887, #889 landed), so **#881 and #882 are now conflicting.** Both have review comments waiting — #881 on the unmeasured cause claim, #882 on the stale `_FakeClient` and a project guard bypassed on the migrated arm.

PR #786, opened 2026-09-11
spike(756): the /about redirect is decided client-side, without asking Flow anything

ffroliva:
#756 measured the redirect and declined to measure its cause. The [2026-09-10 spike](https://github.com/ffroliva/gflow-cli/blob/develop/docs/superpowers/spikes/2026-09-10-about-redirect-stability.md) could not investigate one because it had stopped reproducing.

It reproduces **stably on `denon82` right now**, and a stable reproducer is the condition under which a cause becomes measurable. That state is perishable — `ci-probe` went from reproducing to not in two days with nobody watching — so it was measured while it was there.

## Designed as an A/B, not an inspection

Reading only the failing account shows a page that redirects and a pile of requests, with nothing to say which is abnormal. Same capture on both arms; the finding is the difference.

| | `denon82` (fails) | `ci-probe` (control) |
|---|---|---|
| document response | **200**, no `Location` | **200**, no `Location` |
| first navigation | 146 ms, `/project/<id>` | 90 ms, `/project/<id>` |
| second navigation | **338 ms → `/about`** | 700 ms, same `/project/<id>` |
| total responses | 7 | 43 |
| `batchexecute` calls | **0** | 20, first at **275 ms** |
| non-2xx from any Flow endpoint | none | none |

## Two findings

**Not a server-side redirect.** Both documents are `200` with no `Location`, so the hop is the app deciding client-side, 192 ms after its own first navigation commits.

**It decides without asking Flow.** The failing arm issued **zero** `batchexecute` calls — not *called and was refused*, **never called**. There is no 401, 403 or 404 to point at because there is no request. The sixteen rpcids the control makes are simply absent.

The cross-arm timing is what makes that a finding rather than a race: the control had already made its **first** rpc at 275 ms, before the failing arm hopped at 338 ms. Two runs, so suggestive rather than airtight — but with zero calls in the 5.5 s *after* the hop too, the app is plainly not waiting on a Flow answer.

**Account-wide, not project-specific**: a second `denon82` project redirects identically, 2/2.

Also already in hand: `gflow project list` on `denon82` returns the redirecting project. The backend grants access while the frontend declines to open it.

## What this does NOT establish

**Which signal.** Naming it means reading the document's embedded bootstrap payload, which on this origin is exactly where bearer tokens and account data live. **Not read here.** #756's warning stands: a lead is not a cause, and no remedy should assert a mechanism this cannot see.

What would settle it: a HAR capture of the failing document through `scripts/dev/har-spike/`, diffed against a working one, with `extract_har_summary.py` producing the redacted summary. Not run here because the finding already changes what the next person should look at, and a raw capture of this page must not leave the machine.

## Scope

**No code changes.** `retryable=False` was already correct and was measured in #785. This narrows the search for whoever picks up #756: it is not a permissions call being refused, so looking for one is a dead end.

Cost: $0 — navigations and response *metadata* only. No bodies, no headers, no cookies.

## Summary by CodeRabbit

- **Documentation**
  - Added investigation findings about `/about` navigation behavior.
  - Documented that affected projects receive a successful page response before navigating to `/about` client-side.
  - Recorded that the behavior is consistent across projects within an account and is not caused by session expiry, missing access, a temporary condition, or a server-side redirect.
  - Clarified which potential causes remain unconfirmed and documented differences between affected and control sessions.

### DIFF
--- a/docs/superpowers/spikes/2026-09-11-about-redirect-is-decided-client-side.md
+++ b/docs/superpowers/spikes/2026-09-11-about-redirect-is-decided-client-side.md
@@ -1,0 +1,93 @@
+# The `/about` redirect is decided client-side, in ~200 ms, without asking Flow anything
+
+- **Date:** 2026-09-11
+- **Script:** [`scripts/dev/spike_about_redirect_cause.py`](../../../scripts/dev/spike_about_redirect_cause.py)
+- **Arms:** `denon82` (redirects) vs `ci-probe` (opens) — an **A/B**, not an inspection
+- **Cost:** $0 — navigations and response *metadata* only. No bodies, no headers, no generation.
+- **Raw:** `scripts/dev/_spike_out/spike_about_redirect_cause_20260911_154808.json` (gitignored)
+- **Refs:** [#756](https://github.com/ffroliva/gflow-cli/issues/756)
+
+## Why it was askable at all
+
+#756 measured the redirect and explicitly declined to measure its cause. The
+[2026-09-10 spike](2026-09-10-about-redirect-stability.md) could not investigate one
+because the redirect had stopped reproducing. As of today it reproduces **stably** on
+`denon82` ([5/5](2026-09-11-about-redirect-is-stable-for-an-account.md)), and a stable
+reproducer is the condition under which a cause becomes measurable. That state is
+perishable — `ci-probe` went from reproducing to not in two days with nobody watching —
+so it was measured while it was there.
+
+One clue was already in hand: `gflow project list` on `denon82` returns 50 projects
+**including** the one that redirects. The backend grants access while the frontend
+declines to open it.
+
+## Pre-registered readings
+
+| Outcome | Reading |
+|---|---|
+| document 3xx with `Location: /about` | server-side; the app never boots |
+| document 200, then a client-side navigation | the app boots and decides; the differing rpcid is the lead |
+| an rpcid in both, non-2xx only on `denon82` | that call is the decision point |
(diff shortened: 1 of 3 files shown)


ffroliva:
Both findings were right about the **method**, even though the conclusion survives. Fixed by re-running with better reporting rather than by arguing.

**"Not a server-side redirect" did not follow from what I reported.** Correct. `page.goto` resolves to the final main resource after following redirects, so a 200 with no `Location` is equally consistent with a client hop and with a 302 to `/about`, and `report()` only surfaced non-2xx so a document 3xx would never have appeared. Two discriminators are now reported, and both hold: the document's **resolved URL is `/project/<id>`**, not `/about`, and there is **no 3xx anywhere** in the failing arm.

**"Zero batchexecute" is not "no Flow request".** Also correct, and the wording claimed more than the method checked. Every response is now bucketed. Classified in full the finding is *stronger* than I first stated:

| | `denon82` | `ci-probe` |
|---|---|---|
| `batchexecute` | 0 | 20 |
| other Flow requests | **1 — the document itself** | 15 |
| non-Flow (fonts, GTM, analytics, OneGoogle, Play) | 6 | 8 |

So it is not zero of one route shape — it is zero of everything else too. The failing arm's only request to `flow.google.com` is the document.

**Status multiset.** Fixed: `[200]` and `[200, 200]` were collapsing to the same answer, which would have hidden a call the failing arm made half as often.

Re-ran end to end ~3.5 hours after the first capture: identical on every row.

PR #787, opened 2026-09-11
feat(video,image): add --resolution control and nano2-lite model support

pulsy-cb:
## Summary

This PR adds two user-requested capabilities across CLI, MCP, and transport layers:

1. **Explicit video resolution control (`--resolution [360p|720p]`)**:
   - Available on `gflow video t2v`, `i2v`, `r2v` and MCP tool `gflow_generate_video`.
   - Wired into `GenerateVideoRequest` DTO, worker codec, and `MigratedComposer._select_resolution`.
   - Resolves an issue where models like `omni-flash` default automatically to 360p without user control, enabling explicit 720p (or 360p) selection.

2. **Nano Banana 2 Lite model support (`--model nano2-lite`)**:
   - Discovered and mapped Google's internal wire model string `HARBOR_SEAL`.
   - Available in `gflow image t2i`, `i2i`, and `batch`.
   - Enums and catalog updated (`gflow models` cleanly displays `HARBOR_SEAL | nano2-lite`).

## Verification & Quality Gates

- **Unit & Integration tests:** 2506 passed across `tests/api/`, `tests/mcp/`, `tests/cli/`, `tests/flow_selectors/`.
- **Parity tests:** `tests/mcp/test_cli_parity.py` and `tests/mcp/test_server.py` passed with 0 violations.
- **Hygiene & Lint:**
  - `check_repo_hygiene.py`, `check_doc_links.py`, `check_website_docs_pii.py`, `generate_website_docs.py --check`, `check_council_memory.py` all clean.
  - `ruff check src tests` & `ruff format --check src tests` clean.
  - `pyright src` clean (0 errors, 0 warnings).
- **Live verification:**
  - Verified `image t2i "un croquis..." --model nano2-lite`: Successfully generated 768x1376 JPEG in ~12 seconds.

## Summary by CodeRabbit

* **New Features**
  * Added 360p and 720p resolution options for text-, image-, and reference-to-video generation.
  * Added resolution support to the `gflow_generate_video` MCP tool.
  * Added the `nano2-lite` and `nano-banana-2-lite` aliases for Nano Banana 2 Lite image generation.
  * Added Nano Banana 2 Lite support for batch image generation, including up to 10 reference images.
* **Documentation**
  * Updated usage and MCP documentation with the new model and video resolution options.


ffroliva (changes requested):
## Maintainer review — thanks, and a note on what CI did *not* tell you

**Your checks show 2/2, not 15/16.** Only the labeller workflows ran; the test matrix, SonarCloud and the hygiene gates never executed on this PR. So I ran them locally against your branch merged onto current `develop`:

| Gate | Result |
|---|---|
| `ruff check` / `ruff format --check` | clean, 467 files |
| `pyright src` | 0 errors |
| `pytest -m "not live and not e2e and not smoke"` | **4241 passed**, 24 skipped |

Your gate claims hold. The findings below are things the gates cannot see.

Two genuine positives before the list: the MCP twin is done properly (`mcp/tools.py` and `worker/codec.py` both move with the CLI flag, which external PRs usually miss), and `_select_resolution` runs inside `apply_video_settings`, so an unsupported combination fails **before** any submit and costs nothing. That is the right place for it.

---

## Blockers

### 1. It reverses a consent decision made yesterday, and the comment above it still says why not

`_dismiss_cookie_bar` now falls back **reject → accept → `bar.locator("button").first`**. I measured the bar's DOM order on 2026-09-11: `__accept` is first. So two of those three branches accept cookies on the operator's behalf.

The docstring immediately above `COOKIE_BAR_REJECT`, three lines up in the same file, still reads:

> REJECT, not accept. Both buttons remove the bar and unblock the composer identically, and only one of them answers a consent question on the operator's behalf.

The code below it now does the thing the comment forbids. That is the failure shape this repo keeps writing retrospectives about.

It is also **not mentioned anywhere in the PR description**, and it is speculative: `__reject` was present on every bar observed across three profiles and a clean browser context. If you have a capture where reject is absent, I want to see it — otherwise please drop the fallback and let the click post-mortem name a bar that will not go, which is what it is there for.

### 2. A raw Flow response body is logged unredacted

```python
log.warning("migrated.image_submit_unparsed_reply", ..., text_head=text[:300])
```

That is the first 300 characters of an `ogiZ0b` reply body. I checked `observability.py`: the structlog processor chain is `merge_contextvars`, `add_log_level`, timestamper, `StackInfoRenderer`, `exc_renderer`, renderer. **There is no redaction processor**, so the field prints verbatim to stderr and into `--json` logs.

That body carries the prompt and media ids, and the 300-char window reaches into the payload. It fires on the *failure* path, which is exactly when someone pastes their output into an issue — the scenario #777 existed for. Please route it through `redact_sensitive_text`, or log `parsed_rpcids` and `text_len` only, which is what actually diagnoses an unparsed reply.

---

## Majors

### 3. It removes a waiver without meeting the condition the waiver stated

`_UNMODELLED_WAIVERS` carried, in the repo's own words:

> "🍌 Nano Banana 2 Lite" — Discovered 2026-08-26. A lower-tier image model we do not expose. **Waived pending a capability spike (cost/quality) before we ship it.**

This PR ships it and deletes the waiver. The evidence offered is one successful `t2i`, which shows the wire string works — not what the waiver asked for. A cost/quality comparison against `nano2` is the gate this project wrote for itself on this exact model.

Credit where due: the governance test is what forced you to touch that list, and you updated it honestly rather than suppressing the test. I would like the spike behind it before the waiver goes.

### 4. An unmeasured value inherits a comment that says it was measured

`_IMAGE_REFERENCE_CAP` is introduced by:

> Per-model I2I reference-image cap (**live-observed**). Flow silently keeps only the first N references when more are attached…

`HARBOR_SEAL: 10` is added under that comment, but your live check was a `t2i` with no references. If Lite's real cap is lower — Imagen sits at 3 — a user attaching ten gets references silently dropped on a billed generation, which is precisely what the map exists to prevent. Either observe it, or start it at the conservative value with a note.

---

## Moderate — and this one is **not yours**

### 5. A hidden model picker reads as "already selected" (pre-existing on `develop`)

I nearly filed this against your change and the control says otherwise, so I want to be exact.

Your rewrite wraps the picker guard in `wait_for(state="visible")` inside a bare `except`, re-raising only when `count()` is 0. I expected that to break the present-but-hidden case (#749's agent-mode shape). Probed in real Chromium, a button that is present under a bare `hidden`:

```
button count():    1
button visible():  False
migrated.image_model_already_selected  model='arrow_drop_downNano Banana 2' requested=NARWHAL
RESULT: returned with NO error — the guard did not fire
```

**`develop` does exactly the same.** So this is a latent defect, not a regression, and I will file it separately: a hidden picker makes `--model` silently read as already-satisfied, so a wrong-model generation gets billed and reported as success.

What your change *does* alter is latency — an **absent** picker now waits 4 s before raising instead of raising at once. Since you are already in this function, fixing the fallthrough here would be welcome, but it is not a condition of merging.

---

## Minor

- **The branch is behind `develop` by two commits and `CHANGELOG.md` conflicts.** Please merge `develop` in; nothing else conflicted when I tested it.
- `--resolution` help says "supported on omni-flash" but the flag is offered unconditionally on all three video commands. That is fine given it fails pre-submit, but the help text reads like a guarantee. Consider naming the failure mode instead.

---

## What I would like

Blockers 1 and 2 are small and mechanical — drop the accept fallback, and stop logging the body slice. Do those and rebase, and I will re-run the gates.

Majors 3 and 4 are about evidence rather than code: the capability spike the waiver asked for, and a reference-cap observation. If the spike is more than you want to take on, say so and I will run it — the feature is worth having and I would rather not stall it on something I can measure myself.

ffroliva:
**I ran the capability spike** — major 3 from my review is settled, and it went further than expected. Branch `docs/nano2-lite-capability-spike` (commit 9f2b4c16), PR to follow.

## The waiver's condition is half unanswerable, and that is not your fault

The repo's own [cost/quota catalogue](https://github.com/ffroliva/gflow-cli/blob/develop/docs/superpowers/specs/2026-08-26-model-cost-quota-catalog.md) already establishes that image generation costs **0 Flow credits**, and that **"daily quota has no such oracle — it is observable *only on exhaustion*, via the 429"**. So the *cost* half of "cost/quality" cannot be measured without deliberately burning an account's day. I recorded that rather than inventing a number.

## Quality: answered, and it supports shipping

Two arms, same prompt and aspect, both clean:

| Arm | Elapsed | Dimensions | Bytes |
|---|---|---|---|
| nano2-lite | 20.3 s | 1024×1024 | 428,799 |
| nano2 | 25.4 s | 1024×1024 | 384,036 |

No visible degradation on either; lite slightly more surface texture, nano2 slightly smoother. **One prompt cannot rank two tiers** and I have not tried to — but it does support the narrower claim the waiver needed: Lite is a working tier returning a usable image at full resolution.

**Your wire string is independently corroborated.** `HARBOR_SEAL` and `Nano Banana 2 Lite` both appear in `HTrJv`, Flow's own model catalogue for this account — a surface that is not our request echoed back. So the mapping is real.

**Major 3 is cleared from my side.** I would still like the cost gap carried forward rather than deleted along with the waiver, which the spec files already do.

## What the spike found that changes how we read your live check

`migrated_composer.py` sets `model_name_type=request.model.value` — it echoes the request. So "generated a JPEG with `--model nano2-lite`" proves the UI did not error; it does **not** prove the tier bound. I went looking for a real oracle and there is not one:

| Surface | Model tokens | Media ids |
|---|---|---|
| `ogiZ0b` submit reply | **none** (both arms) | — |
| `HTrJv` project load | **all six** | **none** |
| `Zzl0ze` project load | **none** | **both** |

Nothing joins a media to its model. So the binding is **unverifiable** — by your check or mine. That is filed as #789 against `develop`, not against this PR; it is pre-existing and affects every tier.

Major 4 (the `_IMAGE_REFERENCE_CAP` entry documented as "live-observed") still stands, and the two blockers are unchanged.

ffroliva:
Following up on the attribution point, because "unverifiable" was too strong and you deserve the actual boundary.

## Is it resolvable? Not by anything I can reach — but the search is now wide, not narrow

My first pass searched two replies. I redid it with **no filter**: every `batchexecute` reply on a project load, 20–21 of them, checked for model tokens and for the two media ids I had just generated.

| rpcid | Size | Model tokens | Media ids |
|---|---|---|---|
| `HTrJv` | 24,760 B | **all six** | **none** |
| `Zzl0ze` | 41,906 B | **none** | **both** |
| `tRARke` | 31,883 B | none | none |
| 17 others (`o30O0e`, `cPZSdc`, `as29s`×5, …) | 139–3,964 B | none | none |

**Zero replies contain both.** The catalogue lists models; the media listing lists media; nothing joins them.

I also checked the page itself, since if Flow's UI shows it, the data is there. Two tier labels turned up — `span.settings-summary` and `span.model-select-trigger-content` — but the first appears **on load, before anything is clicked**, so both are the composer's own picker, not a property of any image.

**What is still unmeasured:** whether a tier label *tracks* the media you open. That is the one reading that would flip this, and I could not test it — thumbnail `src` does not carry the media id, and my generic selector matched zero tiles on the last run. Named rather than buried: it is the thread to pull.

## Are we on the same path and protocol?

Yes, and that is the useful part of the answer. gflow drives the same web app you do, so we see exactly what the browser sees — there is no privileged API we are missing out on.

The difference is **between hosts, not between us and the UI**. The old labs REST path returns `modelNameType` per generated image, and `dto.py` reads it. The migrated `batchexecute` wire does not appear to carry it for images. We kept the field name when the host changed and quietly started filling it from the request instead. That is the actual defect in #789 — not that we are on a wrong path, but that we inherited a field the new wire does not back.

Worth noting: Flow **does** store a model per media for *video* on the labs shape (`videoModelControlInput.videoModelName`). So the concept exists in their data model. Whether an image equivalent is exposed anywhere on the migrated host is exactly what is unmeasured.

## Is it regional?

No evidence for it, and I do not think it is the right shape of explanation. Region and cohort gating are real in this project — the repo already records that a Veo tier may be region-gated — but that changes **which models an account is offered**, not whether a reply carries an attribution field. And this account *is* offered Lite: `HARBOR_SEAL` appears in its catalogue. A missing schema field is not an availability question.

For completeness: this account is served UK-region Google cookies. Testing the same probe from another region would settle it, and I have not done that, so I am not claiming either way.

## For your PR

None of this blocks you. It is #789, pre-existing on `develop`, and it affects every tier equally. It only changes how your live check should be read: it shows the command works end to end, which is real, and it cannot show the tier bound — but neither can mine, so that is not a bar I am holding you to.

Still open from my review: the two blockers, and major 4 (the `_IMAGE_REFERENCE_CAP` entry filed under a "live-observed" comment).

ffroliva:
Reopening immediately to trigger CI — no workflow run was ever created for `ec169df4`, so this PR has never actually been tested. Not a comment on the change itself.

ffroliva:
## Gated locally, because CI has never run on this PR

`gh pr checks` shows green here, and that is misleading: GitHub holds fork workflows, and a run that never started shows as nothing rather than as a failure. Checked directly — for head `ec169df4` the only workflow that has ever run is `External PR Triage` (which is `pull_request_target`, so forks do trigger it). The `CI` workflow has **no run at all**. There is no held run to approve and `ci.yml` has no `workflow_dispatch`, so it cannot be dispatched either.

So the head was merged onto current `develop` locally and put through the full gate:

```
repo hygiene              ✅  1135 tracked files, no violations
doc links                 ✅  all links resolved across 192 files
website/docs mirror       ✅  in sync (21 files), nav complete
website/docs PII          ✅  no private identifiers
ruff check src tests      ✅
ruff format --check       ✅  480 files already formatted
pyright src               ✅  0 errors, 0 warnings, 0 informations
pytest (offline suite)    ✅  4465 passed, 24 skipped in 391s
```

**The substance of this PR passes everything.** This is the first actual verification it has had.

## What blocks it: it does not merge cleanly

Three conflicts against current `develop` — `CHANGELOG.md`, `docs/MCP.md`, and `website/docs/MCP.md` (generated from the second).

`CHANGELOG.md` is the ordinary kind: keep both entries under `## [Unreleased]`.

**`docs/MCP.md` needs care, because this branch's line is the older one.** The `gflow_generate_video` bullet here was written against a revision that predates #792, so taking this side wholesale would silently delete develop's run-unique-upload text — the part explaining that `hero.png` is listed as `hero-a1b2c3d4.png` so a re-run cannot bind an earlier upload. That loss would not show up as a conflict afterwards; it would just be gone.

The correct resolution is **develop's line plus this PR's two additions**:

1. `model, duration, count, tools,` → `model, duration, resolution, count, tools,`
2. after `10 is omni_flash-only)`, add: `` , `resolution` ('360p' or '720p', supported on omni_flash) ``

Then regenerate the mirror with `python scripts/ci/generate_website_docs.py` rather than editing `website/docs/MCP.md` by hand.

Once that lands, please push it so a fresh head is created — and note that a fork PR still needs its workflow run approved before CI reports anything.

ffroliva:
Taking this over to get it landed. It has been quiet for 12 days, and the feature is worth having. I'll push to this branch (maintainer edits are enabled) and merge `develop` in rather than force-push, so your commit and authorship stay intact. The plan: resolve the conflicts as described above, drop the cookie-bar accept fallback (blocker 1), log only the reply length and rpcids instead of the body (blocker 2), and start Lite's reference cap at a conservative value marked unmeasured (major 4). The capability spike already settled major 3. If you're still around and would rather finish it yourself, say so and I'll stop.

ffroliva:
Merged as `3a934175`. Thanks @pulsy-cb: `--resolution` and `nano2-lite` are both in. Changes made during the takeover: the cookie bar is reject-only again, the `ogiZ0b` reply body is no longer logged, Lite's reference cap starts at 3 and is marked unmeasured, and the labs editor now refuses `--resolution` before submit instead of silently using Flow's default. Verified live on a migrated account: nano2-lite t2i through the client and MCP, and omni_flash t2v at 720p (client) and 360p (MCP). All 4 are now e2e tests in `tests/e2e/test_migrated_host_e2e.py`.

Issue #681, opened 2026-09-05
labs.google: a cohort with no duration control fails as selector drift (exit 23) after ~40s instead of exit 11 immediately


ffroliva:
## Summary

On `labs.google`, an account/model cohort that renders **no duration control at all** fails as `UiSelectorDriftError` (exit 23) after ~40 seconds. It should fail as `ConfigurationError` (exit 11), immediately, with a remediation that names the model workaround.

The migrated `flow.google.com` composer already does exactly this. The labs path never got the same fix, so gflow currently behaves **better on the newer host than on the older one**.

## Where

`src/gflow_cli/api/transports/ui_automation_video.py:1777` — `_select_video_duration`:

```python
tab = await VideoGenerationMixin._probe_selector_cascade(
    page, "duration_tab",
    (f"[role='tab']:text-is('{seconds}s')", ...ten candidates...),
)
if tab is None:
    raise UiSelectorDriftError(...)          # exit 23
```

`_probe_selector_cascade` (same file, line 1284) defaults to `timeout_ms=4000` per candidate. When the cohort renders no duration row, all ten miss in sequence.

## The reference implementation already in the tree

`src/gflow_cli/api/transports/migrated_composer.py:496-514` separates the two cases:

```python
# Only the duration row is a per-account/model capability (#650) — a missing
# mode/aspect/count radio, or an empty pane, is the DOM having changed.
if axis != "duration" or groups == 0:
    raise UiSelectorDriftError(...)
raise ConfigurationError(
    detail="the migrated Flow host renders no duration control offering ...",
    remediation_hint=(
        "Drop --duration to accept Flow's default length, or pick a model whose "
        "settings pane shows a duration row ..."
    ),
)
```

## Comparison

| | labs.google (today) | flow.google.com (already fixed) |
|---|---|---|
| Diagnosis | selector drift — "our selectors broke, file a bug" | capability absent for this account/model |
| Exit code | 23 | 11 |
| Cost on miss | ~40 s (10 candidates x 4 s) | immediate |
| Remediation | "Omit `--duration`" | "Drop `--duration`, **or pick a model whose settings pane shows a duration row**" |

## Why it matters beyond the wrong exit code

1. **Wrong diagnosis sends users to the wrong place.** Exit 23 means "gflow's selectors broke against Flow's DOM." Here nothing is broken — the account simply has no duration control. A user reasonably files a selector-drift bug for a working build.
[...]

Issue #644, opened 2026-09-03
bug: auth check + browserless cookie harvest keyed on labs.google cookie that flow.google.com never receives


ffroliva:
## Summary

On Flow's migrated `flow.google.com` origin, the Google SSO session **is** the Flow session — but gflow's auth check and browserless cookie harvest are both still keyed on the `labs.google`-scoped NextAuth cookie, which is never sent to the new host.

Measured 2026-09-03 on `ffroliva` and `denon82`, read-only, zero credits.

## The measurement

```
__Secure-next-auth.session-token   domain: labs.google   -> NOT sent to flow.google.com

cookies actually SENT to flow.google.com:
  SAPISID, SID, HSID, SSID, APISID,
  __Secure-1PSID, __Secure-3PSID, __Secure-1PSIDTS, __Secure-3PSIDTS, ...

new jar domains after visiting: flow.google.com, .flow.google.com
```

Both profiles: `signin_cta: 0`, 52 and 47 `button[aria-label]` rendered — **fully authenticated**, on the `.google.com` SSO cookies alone.

## Why it matters

**1. The auth check inspects the wrong credential.** `api/client.py:209`:

```python
_FLOW_SESSION_COOKIE = "__Secure-next-auth.session-token"
```

used at `client.py:589` (`preread_session=...`) and `client.py:624`. On a fully-migrated account this judges a cookie that is no longer in use, so `gflow auth status` and the preread can report a Flow session state that has nothing to do with whether the migrated host will authenticate.

**2. The browserless cookie harvest collects the wrong set.** `auth/cookies.py:30`:

> `httpx_cookies` is intentionally limited to cookies scoped to labs.google.

with `_name_value_cookies(cookies, flow_only=True)` at `cookies.py:132` and `:181`. Correct for `labs.google`; on `flow.google.com` it filters out precisely the cookies that do the authenticating.

This inverts the documented model — the existing note that "Flow session != Google SSO session" is true on `labs.google` and **false** on `flow.google.com`.

## Suggested direction

Make the Flow-session predicate host-aware rather than cookie-name-fixed: on the migrated origin, a valid Google SSO session (`SAPISID` + `__Secure-1PSID`) is the credential. The existing `google_session` field on `ChromeCookieSnapshot` already carries that signal — it is computed from the full jar and then discarded for the migrated case.

[...]

Issue #792, opened 2026-09-12
video: portrait (9:16) dispatch fails with UiSelectorDriftError — driver captures a blank page while the app is live (reproducible; landscape OK)


ai4U23:
# video: portrait (9:16) dispatch fails with UiSelectorDriftError — driver captures a blank page while the app is live (4/4 reproducible; landscape identical flags OK)

## Environment
- gflow-cli 0.73.1 · Python 3.12.12 · macOS 15 (arm64) · system Chrome channel (`--browser chrome`)
- Account: Google AI ULTRA, **server-migrated** to flow.google.com (labs.google hands off instantly)
- Generation runs in an **existing project** via `--project <id>` (required on migrated accounts — labs `project.createProject` 401s otherwise)

## Repro (deterministic, 4/4 + variants)
```
gflow video i2v keyframes/s1-k.jpg "<motion prompt>" \
  --aspect 9:16 --duration 8 --project cf2d62e3-411f-4652-930a-1eba9741a462 -o out.mp4
```
Variants tried: plain, `--ui-mode classic`, agent-driven retries with the editor
verifiably loaded, and **omitting `--aspect` entirely** (inheriting the project's
sticky 9:16 state). All → exit 23:

```
UiSelectorDriftError (retryable: False)  phase/route: video_generation / -
```

## Evidence (3 incident bundles preserved)
- `ui.json`: **empty DOM capture** — `tag_counts.div = 0`, `title.length = 0`,
  `url.host_category = "other"`, viewport 1280×720 (the default, not portrait)
- `sensitive/screenshot.png`: **pure white frame**
- `browser.json`: no page errors, no console output
- `network.json`: **the app is alive** — `POST /_/id-*/data/batchexecute → 200`
  flowing at the same timestamps

So the `Page` object the driver holds at capture time is a blank page while the
real app runs elsewhere — the driver appears to lose its page/tab reference on
the portrait path specifically.

## Contrast (same session, same account, same setup)
- `--aspect 16:9` i2v: success (submit `YhhmEf` → polls `jwpduf` → `as29s` →
  download, ~80 s/clip)
- `gflow image t2i` (nano2, 16:9): success ×5
- Portrait via the Flow UI manually: works fine (9:16 radio in the settings
  popover; 12 credits/clip) — our current fallback
- macOS Accessibility (AX) reads the portrait page fine — only the Playwright
  frame read is blank

## Findings from instrumenting (DEBUG + verbose, see addenda below)
[...]

PR #793, opened 2026-09-12
fix: migrated-host auth verification + i2v picker confirm + run-unique upload

ai4U23:
# fix: migrated-host auth verification + i2v frame-picker confirm + run-unique upload

Fixes three production blockers hit generating on a **server-migrated** account
(ULTRA, flow.google.com is the only frontend — the labs handoff fires instantly),
all verified end-to-end on live generations:

## 1. `auth/verification.py` — migrated-host session fallback

For migrated accounts the labs.google NextAuth session is **never minted** (the
labs app hands off to flow.google.com before/at the NextAuth callback; observed
`all_cookies: 47, flow_domain: 0` — no labs cookies at all), so the
`labs.google/fx/api/auth/session` oracle reports `GOOGLE_SESSION_ONLY` forever
and login can never succeed, even with the Flow workspace + composer fully
loaded in gflow's own Chrome window (5 attempts, agent-verified).

Fail-closed fallback: when the labs oracle yields `GOOGLE_SESSION_ONLY`,
upgrade to AUTHENTICATED **only if** the profile holds BOTH `SAPISID`
(.google.com) AND `__Secure-OSID`/`OSID` (.flow.google.com — the migrated app's
own session), with the email resolved from myaccount.google.com. Anything else
keeps the original outcome.

Related: #639, #644, #681 (this is the login-side of the migrated-host family).
Also reported as kodelyx/flow-agent#15's "the session BFF is the sole oracle"
problem, same root from the other side.

## 2. `migrated_composer.py::_pick_frame_by_name` — click the picker's confirm

Flow's Frames picker no longer auto-closes on pick: it now requires an explicit
confirm button ("Add to prompt" in en). gflow clicked the asset, waited
`FRAME_COMMIT_HIDDEN_S` for an auto-close that never came, and raised
`UiSelectorDriftError` — **5/5 failures on i2v** (landscape t2v unaffected: no
picker). Fix: grace-wait 1.2 s; if the picker is still visible, click the
confirm button; the original wait-for-hidden then decides (auto-close variant
unchanged).

## 3. `migrated_composer.py::attach_start_frame` — run-unique upload name

The picker lists assets by display name, and repeated runs with the same
keyframe file leave N look-alikes in the library. The name search then binds a
**stale duplicate** and the submit-body assertion fires post-click —
`eb1hJf does not carry the uploaded start frame <id> (ids in the body: …)` —
after the credit-spending click. Fix: upload a run-unique COPY of the frame
(`<stem>-<8hex><ext>`) so the search has exactly one match; temp copy removed
after attach.

## Verification (all on the migrated host, live)

- Login: `auth_flow_session_verified user_email=<account> probe=on_disk` via
  the fallback; profile renamed + set default; `auth status` green.
- `gflow video i2v <keyframe> --aspect 9:16 --duration 8 --project <id>`:
  submit body carries the matching upload id → `YhhmEf` submit → `jwpduf`
  polls 6→2→3 → `as29s` → auto-download → **ffprobe 720×1280, 8.0 s**. Zero
  Computer-Use, zero manual steps.
- `gflow image t2i --model nano2 --aspect 16:9`: 5/5 free keyframes.

Notes for review:
- The fallback's myaccount.google.com email scrape is locale-tolerant for
  gmail.com addresses; a structural anchor (or the app's own account endpoint)
  would be the cleaner upstream form.
- The confirm-button anchor uses en text (`Add to prompt`) — the module's
  locale-independence rule would prefer a structural/ligature anchor; left as
  text because the button has no icon in the observed DOM.
- Both deltas are intentionally conservative (fail-closed) and only activate on
  migrated accounts.

Witnessed on the way (not addressed here): exit-23 selector drift on the
labs cohort with no duration control (#681 family) and credits probing
aisandbox-pa on migrated accounts.

## Summary by CodeRabbit

- **New Features**
  - Added an authentication fallback for migrated environments using Google SSO session cookies.
  - Validates sessions through Google account verification and only accepts secure, valid account results.
  - Recognizes Workspace email domains and displays the most frequently detected domain when available.
  - Supports successful authentication even when no email domain can be identified.

- **Bug Fixes**
  - Prevents authentication when required session cookies are missing.
  - Rejects sign-in redirects, invalid destinations, and failed account checks.

### DIFF
--- a/src/gflow_cli/api/transports/migrated_composer.py
+++ b/src/gflow_cli/api/transports/migrated_composer.py
@@ -1346,10 +1346,26 @@ class MigratedComposer:
         from gflow_cli.api.client import validate_image_file  # noqa: PLC0415 - cycle
 
         await validate_image_file(image_path)
-        # No outer budget: each leg is bounded, and an outer one firing first would
-        # replace the stage-named failure with a generic "attach timed out".
-        media_id = await self._upload_via_toolbar(page, project_id, image_path)
-        await self._pick_frame_by_name(page, image_path.name, media_id)
+        # ai4u delta (2026-09-12): the picker lists assets by display name and the
+        # submit body is asserted to carry THIS upload's media id. Repeated runs with
+        # the same file name leave look-alike assets in the library, and the name
+        # search then binds a stale duplicate — the submit body carries the wrong id
+        # and the run dies WireFormatError post-click (credits spent). Upload a
+        # run-unique COPY so the name search has exactly one match.
+        import shutil  # noqa: PLC0415
+        import uuid as _uuid  # noqa: PLC0415
+
+        unique_image = image_path.with_name(
+            f"{image_path.stem}-{_uuid.uuid4().hex[:8]}{image_path.suffix}"
+        )
+        shutil.copy2(image_path, unique_image)
+        try:
+            # No outer budget: each leg is bounded, and an outer one firing first would
+            # replace the stage-named failure with a generic "attach timed out".
+            media_id = await self._upload_via_toolbar(page, project_id, unique_image)
+            await self._pick_frame_by_name(page, unique_image.name, media_id)
+        finally:
+            unique_image.unlink(missing_ok=True)
         return media_id
 
     async def _upload_via_toolbar(self, page: Page, project_id: str, image_path: Path) -> str:
@@ -1747,6 +1763,22 @@ class MigratedComposer:
                 # Outside the except above on purpose: a click that fails here is not
                 # "the picker never listed it", and must not be reported as one.
                 await options.first.click(timeout=4000)
+                # ai4u delta (2026-09-12): Flow's picker no longer auto-closes on pick —
+                # it now requires an explicit confirm button ("Add to prompt" in en).
+                # Without clicking it the overlay stays up and the wait-for-hidden below
+                # raised UiSelectorDriftError on every i2v (observed 5/5 on a migrated
+                # account; landscape t2v unaffected — no picker). Grace-wait, confirm if
+                # the picker is still up, else fall through to the auto-close wait.
+                try:
+                    await page.wait_for_timeout(1200)
+                    if await picker.is_visible():
+                        confirm = page.locator(
+                            "button:has-text('Add to prompt')"
+                        ).first
+                        await confirm.wait_for(state="visible", timeout=3000)
+                        await confirm.click(timeout=3000)
+                except Exception:
+                    pass  # auto-close variant, or confirm already gone — wait below decides
                 break
         try:
             # Re-queried, and `.last` like the open: the picker overlay is detached
--- a/src/gflow_cli/auth/verification.py
+++ b/src/gflow_cli/auth/verification.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import re
 from dataclasses import dataclass
 from enum import StrEnum
 from typing import TYPE_CHECKING, Any, cast
@@ -320,6 +321,73 @@ async def verify_flow_session(
     return result
 
 
+async def _verify_migrated_host_fallback(
+    profile_dir: Path, source: str
+) -> FlowSessionStatus | None:
+    """ai4u delta (2026-09-12) — migrated-host session oracle.
+
+    For accounts Google has moved to flow.google.com, the labs.google NextAuth
+    session is never minted (the labs app hands off immediately; observed:
+    47 cookies, zero on labs.google, but OSID/__Secure-OSID present on
+    .flow.google.com and SAPISID/SID on .google.com). The labs oracle then
+    reports GOOGLE_SESSION_ONLY although the workspace is fully usable —
+    gflow's own migrated-host driver authenticates via exactly these cookies
+    (spike 2026-09-05-migrated-host-wire-protocol: ".google.com SSO cookies
+    already in the profile authenticate the host").
+
+    Fail-closed: upgrades GOOGLE_SESSION_ONLY to AUTHENTICATED only when BOTH
+    the .google.com SSO cookie (SAPISID) AND the flow.google.com app session
+    cookie (__Secure-OSID/OSID) are present, AND the account email resolves
+    from myaccount.google.com. Anything else returns None (caller keeps the
+    original outcome). Never touches other outcomes.
+    """
+    from .strategies import async_playwright
+
+    try:
+        async with async_playwright() as pw:
+            ctx = await pw.chromium.launch_persistent_context(
+                user_data_dir=str(profile_dir),
+                channel="chrome",
+                headless=True,
+                args=["--password-store=basic"],
+            )
+            try:
+                cookies = await ctx.cookies()
+                names = {(c.get("name"), c.get("domain", "")) for c in cookies}
+                has_sso = any(
+                    n == "SAPISID" and "google.com" in d for n, d in names
+                )
+                has_flow_osid = any(
+                    n in ("__Secure-OSID", "OSID") and "flow.google.com" in d
+                    for n, d in names
+                )
+                if not (has_sso and has_flow_osid):
+                    return None
+                resp = await ctx.request.get(
+                    "https://myaccount.google.com/?hl=en", timeout=15_000
+                )
+                page_body = await resp.text()
+            finally:
+                await ctx.close()
+    except Exception as exc:
+        logger.warning(
+            "auth_migrated_fallback_probe_error", source=source, error=type(exc).__name__
+        )
+        return None
+
+    match = re.search(r"[\w.+-]+@[\w-]*\.?gmail\.com", page_body)
+    if not match:
+        logger.warning(
+            "auth_migrated_fallback_no_email", source=source
+        )
+        return None
+    return FlowSessionStatus(
+        outcome=FlowSessionOutcome.AUTHENTICATED,
+        user_email=match.group(0),
+        source=source,
+    )
+
+
 async def verify_flow_profile(
     profile_dir: Path,
     *,
@@ -331,6 +399,9 @@ async def verify_flow_profile(
     (falling back to a marker-gated Playwright context on decryption failure),
     then calls the NextAuth session endpoint with up to `_MAX_ATTEMPTS` attempts.
     Fail-closed: any failure yields VERIFICATION_ERROR, never AUTHENTICATED.
+
+    ai4u delta (2026-09-12): when the labs oracle yields GOOGLE_SESSION_ONLY,
+    a migrated-host fallback probe runs (see _verify_migrated_host_fallback).
     """
     _validate_profile_in_home(profile_dir)
 
@@ -362,4 +433,15 @@ async def verify_flow_profile(
             source=source,
             status_code=status_code,
         )
+    if result.outcome is FlowSessionOutcome.GOOGLE_SESSION_ONLY:
+        # ai4u delta (2026-09-12): migrated-host fallback — see
+        # _verify_migrated_host_fallback docstring. Fail-closed.
+        fallback = await _verify_migrated_host_fallback(profile_dir, source)
+        if fallback is not None:
+            logger.warning(
+                "auth_migrated_host_fallback_authenticated",
+                source=source,
+                detail="labs NextAuth absent; flow.google.com OSID session + SSO cookies present",
+            )
+            return fallback
     return result


ffroliva:
Thanks — this is a genuinely valuable contribution and the **diagnosis is the strong part**. All three bugs are real, the wire-level evidence (`YhhmEf` → `jwpduf` 6→2→3 → `as29s`, ffprobe 720×1280/8.0 s) is exactly the standard this project asks for, and you flagged both the gmail-only scrape and the locale-rule violation yourself before we could.

Worth saying explicitly: our own `/gflow:predict` **deferred** the auth escape hatch precisely because *"there is no account in which to write its e2e test"* (`docs/superpowers/spikes/2026-09-06-labs-vs-migrated-session-credential.md:204-235`). You have that account. The blocker is liftable and we want this.

It can't merge as written. Findings below are verified locally against `3387861c`, not inferred.

## Blocking

**1. It breaks 5 existing tests.** Measured at your head vs `develop`:

```
tests/api/transports/test_migrated_composer.py   →  5 failed, 99 passed
(on develop: 0 failed)
  test_attach_uploads_then_binds_the_frame_by_file_name
  test_attach_is_reference_not_found_when_the_picker_lists_no_such_name
  test_attach_is_selector_drift_when_the_chip_stays_empty_after_the_pick
  test_attach_searches_again_when_the_upload_is_not_indexed_yet
  test_attach_picks_the_first_option_when_names_repeat
```

`attach_start_frame` now uploads `<stem>-<8hex><ext>`, and the offline picker fixture filters options by substring, so the renamed file matches nothing → `ReferenceNotFoundError` at `migrated_composer.py:1753`.

**CI never told you this.** All `pull_request` workflows on a fork sit at `action_required` until a maintainer approves them — `gh pr checks` shows only the `pull_request_target` triage job and CodeRabbit, so the PR looks green while nothing ran. Worth running `/gflow:check` against your own head.

**2. `ruff format --check src tests` fails** on both files (`verification.py` ~358-363/374-375; `migrated_composer.py:1775-1777`). Run `uv run ruff format src tests`.

**3. The migrated probe drops three guards the sibling launch 70 lines above holds.** `verification.py:279-295` wraps its `launch_persistent_context` in `ProfileLease`, resolves the channel via `channel_for_profile`, and calls `ensure_profile_engine_compatible`. The new one has none. `tests/auth/test_cookies.py:130` pins the lease by name. Without it, a read-only `gflow auth status` can put a second Chrome on a `user_data_dir` a worker holds — `ProfileLockedError` (exit 11) is the designed answer and is silently skipped. Hardcoding `channel="chrome"` also bypasses the `.gflow_browser_strategy` marker gate at `cookies.py:139-147`.

**4. The identity probe fails OPEN. We measured it today, and it is worse than the gmail-only issue you flagged.**

`ctx.request.get` follows redirects, and `myaccount.google.com/?hl=en` **never** serves the account page — it redirects to `accounts.google.com` on every profile we tried, including a live one. Results against your verbatim regex:

| profile | final host | regex matches | would return |
|---|---|---|---|
| live migrated session | accounts.google.com | 1 | AUTHENTICATED |
| **Flow session DEAD** | **support.google.com** | **2** | **AUTHENTICATED** |
| no cookies at all | accounts.google.com | 0 | None |

So the "success" case reads an email off a **sign-in page**, and a profile whose Flow session is dead authenticates on an arbitrary one of two matches. The pattern also accepts `x@notgmail.com` and `foo@evil-gmail.com`, misses `googlemail.com` and `USER@GMAIL.COM` (no `re.I`), and rejects every Workspace address.

This is not display-only: `user_email` → `.gflow_account` (`real_chrome.py:459`) → `api/client.py:860` `page.locator(f'[data-email="{email}" i]')`, where a wrong value raises `FlowAccountChooserError` and **blocks generation** — and `--account`'s exit-38 assertion silently degrades from server-attested to string-found-in-HTML.

Also note the auth poll is supposed to stay **off** `accounts.google.com` entirely.

**5. `button:has-text('Add to prompt')` is the first live prose selector in `migrated_composer.py`.** The other text hits in that file are comments, ligatures or numerics; the module docstring at `:22-24` states the rule. You're right that there's no ligature — but our own spike already published the structural anchor: **`button.detail-add-to-prompt-btn`** (`docs/superpowers/spikes/2026-09-05-migrated-r2v-attach-surface.md:72`). Caveat we owe you: that dump is the `@`-mention pane, not the Frames picker, so please confirm it on your account in one run; fallback `picker.locator("flow-add-menu-detail-pane button").last`. Please also scope it to `picker.locator(...)` rather than `page.locator(...)` — multiple overlay panes coexist across `FRAME_SEARCH_ATTEMPTS`, and `.first` in DOM order can hit a detached earlier pane, time out, get swallowed by the bare `except`, and re-raise the exact drift this fixes.

**6. No tests on a fail-closed path that decides `AUTHENTICATED`.** `tests/auth/test_verification.py:103 _build_verify_mock` already mocks exactly the surface you use — 5 cases, ~25 lines.

## Two changes would shrink this a lot

- **`chooser.set_files` accepts a Playwright `FilePayload` `{name, mimeType, buffer}`.** Pass the unique *name* with the original bytes: the disk copy disappears entirely — no write into the user's keyframe folder, no orphan on SIGKILL, no `PermissionError` on a read-only source — **and finding 1 goes away**, because the on-disk name never changes.
- **The second Chrome launch is deletable.** `get_chrome_cookie_snapshot` (`auth/cookies.py:186`) already reads the **full** jar and derives `google_session` from it. A sibling `flow_host_session: bool` on `ChromeCookieSnapshot` is ~4 lines, threads into the **pure** `evaluate_session_response`, removes ~50 lines, inherits the lease and marker guards for free — and fixes **both** oracle call sites. Yours patches only `verify_flow_profile`; `internal_chromium.py:129` (the `--browser internal` poll) stays broken.

With those plus `user_email=None`, the diff drops from +118 to roughly +40.

## Two things our audit found that you'll want to know

- **Your trigger condition switches off on the day labs actually dies.** Real deprecation gives 404/5xx/hang → `VERIFICATION_ERROR`, not `GOOGLE_SESSION_ONLY`. Consider `{GOOGLE_SESSION_ONLY, VERIFICATION_ERROR}`.
- **`gflow credits` stays broken after login succeeds** — `api/credits.py:51-72` demands an `access_token` from the labs body a migrated account never has. Green `auth status`, red `credits`. We'll track that separately so #791 doesn't reopen on it.

## Suggested split

Three independent changes, three review surfaces: **(a)** the auth oracle (needs its own predict/scenario/e2e — verdict on the current shape is STOP, on the concept CAUTION), **(b)** the picker confirm (blocked only on the selector anchor), **(c)** the run-unique upload — cleanest of the three and shippable quickly via `FilePayload`. (c) alone would unblock i2v for everyone on the migrated host.

Also: add `[Unreleased] → Fixed` CHANGELOG entries, and the commit needs a DCO sign-off (`git commit --amend -s`).

Happy to take (c) first if you'd rather land something immediately.

ffroliva:
**Two of your three slices are now on `develop`** — merged as #800 (`6b4abb73`), implemented differently but from your diagnosis, and this PR should now rebase down to the auth slice alone.

**Credit where it belongs.** Both defects were yours: you found that the Frames picker stopped committing on the option click, and you found that repeated uploads of the same file leave look-alikes the name search can bind to the stale one. You verified both end-to-end on the affected cohort (`YhhmEf` → `jwpduf` 6→2→3 → `as29s`, ffprobe 720×1280 / 8.0 s). We did not rediscover either. The commit message says so and names you.

**What we changed, and why:**

1. **`has-text('Add to prompt')` → `button.detail-add-to-prompt-btn`.** You called this out yourself, and you were right about the button — our disagreement was only about how to name it. `migrated_composer.py` forbids prose anchors (module docstring, `:22-24`), so we ran a $0 spike on the Frames entry to measure the class rather than ask you for another recon round-trip (`docs/superpowers/spikes/2026-09-13-frames-picker-confirm.md`). The same class was already measured on the `@`-mention entry into that component ten days earlier, so it is now confirmed on both entries.

   Worth knowing: the obvious structural fallback — *"click the picker button that isn't an asset option"* — **would have clicked Close.** The popover's only other visible button is `header-close-btn`. Your text anchor was the safer guess of the two.

   We also scoped the click to the live pane and gated it on `is_visible()`: the method reopens the picker up to three times, and a detached earlier pane's button still resolves.

2. **The run-unique file copy became a Playwright `FilePayload`.** Same idea, no disk write at all — `{name, mimeType, buffer}` gives Flow the unique display name while the user's keyframe is never copied, so there's no orphan on SIGKILL and no `PermissionError` on a read-only source. It also made your five offline-test failures disappear, because the on-disk name never changes.

   We moved the fix one level down, into `_upload_via_toolbar`, rather than onto `attach_start_frame` where the report pointed. `attach_references` reaches the same upload leg and re-uses reference images across runs *by design* — so r2v was the likelier victim of the same bug. It now covers `video i2v`, `video r2v` and `image i2i` alike.

**The ask:** please rebase this PR down to **slice 1, the auth oracle**, and drop slices 2 and 3 — `develop` will conflict with them now. That slice is untouched by #800 and it is the one that matters most: it is #791, it blocks you and at least one other reporter, and our own `predict` deferred it only because we had no account to test it on. You have that account.

The five findings from the earlier review that apply to the auth slice still stand — the `ProfileLease` / `channel_for_profile` / `ensure_profile_engine_compatible` guards, the fail-open identity probe, and tests on a path that decides `AUTHENTICATED`. The `ChromeCookieSnapshot` suggestion still looks like the smaller diff and fixes `internal_chromium.py:129` for free, but that's a suggestion, not a condition.

One thing we'd like from you on the other thread: #792 stays open, and the reason is in the reply there.

ffroliva:
**Heads-up: 0.74.0 is going out without the auth slice, and that is not a rejection of it.**

We had been holding the release for #791. That was the wrong shape — it made our release cadence a function of your availability, and it delivered nothing to anyone in the meantime. So 0.74.0 ships now and the auth slice lands in 0.75.0, on your schedule, with no deadline attached.

**Two of the three defects you found are in 0.74.0**, from your diagnosis, credited by name in the CHANGELOG: the Frames picker that stopped committing on the pick, and the repeated upload that leaves look-alikes for the name search to bind to the stale one. That means the fork you are carrying should shrink from three slices to one — please rebase onto v0.74.0 rather than onto your original base, since `develop` will now conflict with slices 2 and 3.

**Why we did not just build the auth slice ourselves.** We checked rather than assumed. On 2026-09-14 we probed the cookie jar of all nine profiles on this machine:

| profile | SAPISID | labs NextAuth | flow.google.com session |
|---|---|---|---|
| 2 working accounts | yes | **yes** | yes |
| 1 | yes | yes | none |
| 3 | yes | **no** | none |

Four of them print #791's exact symptom string — *"Signed in to Google, but not to the Flow app"* — and **none of them is #791**. That string is also just what a stale profile says. The #791 state is SAPISID **plus a live `flow.google.com` session plus no labs NextAuth token**: the migrated app working while the labs oracle says no. Nothing here reproduces it.

Which leaves us unable to verify a change whose entire job is to decide `AUTHENTICATED`. Writing that blind, on a path where the failure mode is fail-open, is the one place we are not willing to guess. You have the account; the slice is yours.

The five findings from the earlier review still stand as the conditions on it — the `ProfileLease` / `channel_for_profile` / `ensure_profile_engine_compatible` guards, the fail-open identity probe, and tests on a path that decides `AUTHENTICATED`. The `ChromeCookieSnapshot` route still looks like the smaller diff. #791 stays open and unassigned until you say otherwise.

ffroliva:
## Verified against live profiles — the oracle works; two gaps before it can land

CI was **held at the fork gate** and had never run on `3387861c` (the checks list showed green because held runs are hidden). Approved, and separately verified the change by hand against real profiles. $0 throughout.

### It works

Swept all 10 local profiles. Five reproduce #791 (`google_session_only`). Of those, exactly one carries both cookies the gate requires:

| profile | cookies | SAPISID | flow.google.com OSID | gate |
|---|---:|---|---|---|
| default | 51 | ✅ | ❌ | refused |
| ffroliva.bak-224933 | 44 | ✅ | ❌ | refused |
| **gflowclireview** | 53 | ✅ | ✅ | **probes** |
| pr389fresh2 | 44 | ✅ | ❌ | refused |
| promo-denon82 | 53 | ✅ | ❌ | refused |

On `gflowclireview`, `_verify_migrated_host_fallback` returns **AUTHENTICATED with the correct email, 3/3 runs** (2.6 s / 7.2 s / 2.7 s). The four refusals are the fail-closed path behaving correctly.

Worth saying explicitly: **this oracle is stronger than the one in #835.** It makes a real authenticated request to `myaccount.google.com`, so a revoked session fails closed. #835 read the client-rendered DOM on `flow.google.com` and was retracted precisely because that cannot attest to the session. The idea here is the better one.

### Gap 1 — Workspace accounts are still locked out (blocking)

```python
re.search(r"[\w.+-]+@[\w-]*\.?gmail\.com", page_body)
```

Matches `@gmail.com` only. Measured:

| address | result |
|---|---|
| `denon82@gmail.com` | MATCH |
| `dev@axelate.io` (Workspace) | **NO MATCH** |
| `user@mycompany.com` (Workspace) | **NO MATCH** |
| `user@googlemail.com` | **NO MATCH** |

gflow supports *any* Google account with Flow access, Workspace included — and the maintainer's own account is a Workspace domain. For those users #791 stays open with no signal that the fallback silently declined. Widen the pattern, or better, read the address from a structured source rather than regexing an HTML body.

### Gap 2 — the 15 s budget is tight under load (non-blocking, but real)

During the profile sweep the probe hit `auth_migrated_fallback_probe_error TimeoutError`. In isolation the same profile answers in 2.6–7.2 s, so the sweep's browser contention caused it — **my test conditions, not your code.** I reported it as a failure first and that was wrong.

But 7.2 s against a 15 s ceiling is under half the budget on an *idle* machine, and headless Chrome under contention did blow it. Worth raising the timeout, or treating a timeout as retryable rather than a silent `None`.

### Gap 3 — no tests

82 lines of auth logic with no coverage. Auth is where a silently wrong answer is most expensive, and the gmail-only regex is exactly the kind of thing one parametrized test would have caught.

### One thing to settle before merge

**#793 and #835 solve the same issue two ways and both edit `auth/verification.py`.** They cannot both land. On the evidence the oracle here is the better one — it is server-attested where #835's is not. The most likely good outcome is this approach, with #835's test coverage and non-gmail handling folded in.

Verification run on `denon82` + the 10-profile sweep, 2026-09-16.

ffroliva:
## Correction to my previous comment — the PR is red, and I should have led with that

I verified the oracle before running the suite, and reported the good news first. That was the wrong order. **This branch does not currently pass.**

### CI — fails on all three Python versions

Now that the fork gate is approved and the workflows actually run:

```
test (3.11)  fail   18s
test (3.12)  fail   10s     <- Format check: `uv run ruff format --check src tests` exit 1
test (3.13)  fail   15s
```

Ten to eighteen seconds means it dies at the **format gate**, before any test executes. `ruff format --check` is the exact command CI runs; `uv run ruff format src tests` fixes it.

### Offline suite — 5 failures, in the code this PR changes

Locally on `3387861c`: **5 failed, 1142 passed** (4m28s). At least three are in the function this PR modifies:

```
FAILED tests/api/transports/test_migrated_composer.py::test_attach_is_selector_drift_when_the_chip_stays_empty_after_the_pick
FAILED tests/api/transports/test_migrated_composer.py::test_attach_searches_again_when_the_upload_is_not_indexed_yet
FAILED tests/api/transports/test_migrated_composer.py::test_attach_picks_the_first_option_when_names_repeat
```

The `migrated_composer.py` side of this PR (+40 lines, "i2v picker confirm + run-unique upload") changes `_attach`, and the existing `_attach` tests no longer hold. Those tests encode measured behaviour, so either the change is wrong or the tests need updating **with a reason** — not simply deleted.

### What stands from my earlier comment

The **auth oracle genuinely works** — 3/3 AUTHENTICATED with the correct email on `gflowclireview`, the one local profile carrying both required cookies, and the four refusals are fail-closed behaving correctly. It is server-attested where #835's was not. That finding is unchanged and it is the good part of this PR.

And the two gaps stand: the `@gmail.com`-only regex locks out Workspace accounts, and there are no tests for the 82 new lines.

### Why this was invisible

**CI had never run on this PR.** The fork gate held every workflow at `action_required`, and held runs do not appear in the checks list — so the PR displayed green while nothing had executed. That is why the format failure and the five broken tests sat here unnoticed. Approved now, so the signal is real going forward.

Order of business, smallest first: `ruff format`, then the three `_attach` tests, then the Workspace regex, then coverage for the new fallback.

ffroliva:
## Maintainer takeover — pushed to this branch as `58e63e5b`

@AI4Unow I've pushed fixes directly rather than sending you another round. Your commit stays as the parent, and **the core idea is kept** — this was repair, not replacement.

### What I kept, and why it deserved keeping

Your oracle is the right one. Verified on live profiles: swept all 10 local profiles, 5 reproduce #791, exactly one carries both cookies your gate requires — and on that one the probe returns AUTHENTICATED with the correct email, **3/3**. The four refusals are your fail-closed path working correctly.

It is also **stronger than #835's approach**, which solved the same issue and was retracted: yours makes a real authenticated request, so a revoked session cannot fake it. #835 read a client-rendered DOM, which cannot attest to a session at all. Between the two, yours is the one worth building on.

### Six fixes

**1. The address must not be the decision.** `[\w.+-]+@[\w-]*\.?gmail\.com` gated auth on a gmail address. Measured: `dev@axelate.io`, `user@mycompany.com`, even `user@googlemail.com` all fail to match — so #791 stayed open for **every Google Workspace account**, silently. The maintainer's own account is one.

Widening the regex alone would have been worse: any address on a signed-out page would then read as proof. So the **decision moved to where the request landed** — a revoked session is redirected off `myaccount` to sign-in, which is server-attested in a way page content is not. The address is now only a display label.

**2. A selector that could never match for most of the world.** `button:has-text('Add to prompt')`, with the comment noting *"in en"*. AGENTS.md bans text-label selectors in `transports/` outright — on any non-English UI it never matches, the picker stays open, and the user gets the exact `UiSelectorDriftError` this PR fixes. Replaced with `flow-add-menu-detail-pane button`, the structural anchor [`2026-09-05-migrated-frames-attach.md`](../blob/develop/docs/superpowers/spikes/2026-09-05-migrated-frames-attach.md) had already measured — eleven days before this PR.

**3. A broad `except` hiding two guards.** `except Exception: pass` around the confirm swallowed an `AttributeError` on its **first line** — the test fake has no `wait_for_timeout` — so that block had never executed under test, not once. It also swallowed the fixture's own *"composer used an unmodelled selector"* assertion. Narrowed to `PlaywrightTimeoutError`, which is what made the `AttributeError` visible at all.

**4. Five broken tests, and a fixture that didn't model reality.** Your run-unique upload copy is a **good change** — it stops the name search binding a look-alike from an earlier run and spending credits on the wrong asset. But the fake served a static picker list and never listed what was uploaded, so every attach test failed for an unrelated reason. The fake now lists uploads by file name as Flow does, and models **both** picker cohorts (auto-close and confirm-required) so your confirm path is actually exercised. 17 attach tests pass.

**5. A 15 s budget a real run exceeded.** Raised to 30 s — not a precaution: a live run took **18.3 s** and would have timed out and silently declined. It answers in 2.4–7.2 s idle, but browser contention blew the old ceiling.

**6. Five tests for the 82 previously-uncovered auth lines**, each verified to FAIL against the regression it pins.

### Gates on `58e63e5b`

`ruff` clean · `format` clean · `pyright` 0 errors · `tests/auth/` + `tests/api/transports/` **1152 passed, 1 skipped** (was 5 failed, 1142 passed).

### Why none of this was visible

**CI on this PR had never run.** The fork gate held every workflow at `action_required`, and held runs don't appear in the checks list — so it displayed green while nothing had executed. I approved it; the signal is real now.

Thanks for the oracle — it is the part that was hard, and it works.

ffroliva:
## Verified on a live Google Workspace account — the gap I said I could not close

Earlier I flagged that the Workspace fix was *"the exclusion is provably gone, not Workspace is proven working"*, because every profile on this machine is `@gmail.com`. That caveat is now retired by measurement.

A Workspace account (`riskcontrollimited.co.uk`) was logged in during this session and landed in exactly the `google_session_only` state this PR exists to fix.

### A/B with one variable

Same profile, **same cookies, untouched on disk** — only the code differed:

| | released 0.76.0 | this branch (`58e63e5b`) |
|---|---|---|
| exit code | **8** | **0** |
| verification | `auth_flow_session_unverified` | `auth_flow_session_verified` |
| account recorded | `unknown` | `flavio.oliva@riskcontrollimited.co.uk` |
| `channel_for_profile` | **`None`** | **`chrome`** |
| Chrome marker | **rolled back** | intact |

```
{"event": "auth_migrated_host_fallback_authenticated",
 "detail": "labs NextAuth absent; flow.google.com OSID session + SSO cookies present"}
{"event": "auth_flow_session_verified", "user_email": "flavio.oliva@riskcontrollimited.co.uk"}
[OK] Flow session verified.
```

The user did not re-enter credentials — the cookies pre-dated both runs. That makes it a **cleaner** test, not a weaker one: sign-in is held constant and the only thing that changed is this branch.

### Three things it proves

1. **Workspace accounts authenticate.** Under the original `@gmail.com`-only regex this exact address would have been declined **silently**, leaving #791 open for every custom-domain user while the fix appeared to work. This is why the decision moved off the address and onto the server-attested landing URL.

2. **`most_common` picks the right address on a Workspace page.** The gmail page measured 9 occurrences with no competitor; this confirms the heuristic holds on a different page shape, where an org admin or support address could plausibly have appeared.

3. **This PR prevents [#796](https://github.com/ffroliva/gflow-cli/issues/796), it does not merely report it better.** The marker rollback is gated on `not verified`. On 0.76.0 the failure rolled it back and `channel_for_profile` collapsed from `chrome` to `None` — silently downgrading generation to bundled Chromium on a profile created with `--browser chrome`. With this branch, verification succeeds, so the rollback never runs. One root cause, three symptoms, one fix.

### Not fixed here, and worth its own issue

The **login window never closes** for these accounts. `_await_flow_session` polls the labs session endpoint as its auto-close oracle, and that endpoint is precisely what never answers for a migrated account — so the poll spins to the login timeout while the banner promises *"gflow detects the sign-in and closes Chrome for you"*. That is out of scope for this PR but is the first thing a new user on a migrated account meets. Recorded on #791.

Verification run: 2026-09-16, Windows, `--browser chrome`, $0.

PR #814, opened 2026-09-14
fix(serve): require the daemon token on every HTTP request

ffroliva:
## Summary

`gflow serve` accepted a shared secret via `GFLOW_CLI_DAEMON_TOKEN` / `GFLOW_DAEMON_TOKEN`, and refused to bind a non-loopback address without one — but never checked it on a request. It gated startup only.

This makes the token authoritative: it is verified on **every** HTTP request, on both the Streamable HTTP transport and the deprecated SSE transport, with a constant-time comparison. A missing, malformed or wrong token gets `401` plus `WWW-Authenticate: Bearer` and never reaches a tool.

When no token is configured and the bind is loopback, behaviour is unchanged — existing local setups and MCP client configs keep working.

## Evidence — A/B against released 0.74.0

Same probe, loopback bind, `--no-spend`, throwaway `GFLOW_CLI_HOME`:

| request | 0.74.0 initialize / tools_list | this branch | tools exposed (before -> after) |
|---|---|---|---|
| no `Authorization` | 200 / 200 | **401 / 401** | 13 -> 0 |
| `Bearer <wrong>` | 200 / 200 | **401 / 401** | 13 -> 0 |
| `Bearer <right>` | 200 / 200 | 200 / 200 | 13 -> 13 |

`WWW-Authenticate: Bearer realm="gflow"` on both 401s.

## Three details that are load-bearing

- **Auth is the outermost layer.** Starlette's `add_middleware` inserts at index 0, so auth is added last. An unauthenticated caller gets 401 and cannot probe the transport-security configuration behind it.
- **`host` is forwarded to the SDK's app builders.** The SDK auto-enables DNS-rebinding protection only when it sees a loopback host. Dropping the argument would have silently disabled protection that works today — measured on a loopback bind: bad `Host` -> 421, bad `Origin` -> 403.
- **The app is rebuilt per call**; its session manager is single-use, so it must not be memoised.

## Test plan

- [x] 19 new cases in `tests/mcp/test_serve_auth.py` — absent / wrong / correct token, both transports, case-insensitive scheme, `hmac.compare_digest` reached, auth-before-transport-security ordering, and guards that the correct token and legitimate local requests are **not** rejected.
- [x] Proven red before the fix (19 failed), and A/B'd against a null implementation to confirm 10 of them are genuine fix-detectors rather than tests that can never fail.
- [x] `tests/mcp/` — 212 passed. `test_http_transport.py` (real `gflow serve` subprocess, no token) still passes, proving the no-token loopback path is unchanged.
- [x] Real-subprocess wire check, not just in-process ASGI.
- [x] `ruff check` / `ruff format --check` / `pyright src` clean; repo hygiene, doc links, website mirror green.

## Two pre-existing tests changed, deliberately

`tests/mcp/test_server.py` had two tests asserting the exact call this replaces (`run_streamable_http_async` / `run_sse_async`). Their intent — mount path `/mcp`, `stateless_http` not forced, host/port pass-through — is preserved and re-pointed at the new seam. The old assertion pinned the vulnerable wiring, so keeping it verbatim was not an option.

## Non-breaking contract

- [x] No existing default changes; loopback + no token behaves exactly as before.
- [x] No exit code changes meaning; `ConfigurationError` / exit 11 semantics untouched.
- [x] No CLI option removed or renamed.
- [x] MCP tools unchanged — no schema, payload key or `tools/list` default touched.
- [x] No SQLite migration.
- [x] `--dry-run` untouched.
- [x] The pre-existing suite passes, with the two exceptions explained above.

## E2E / Iron Law

No Flow surface is touched — this is transport auth in front of the MCP server, and no browser, profile or generation path is involved. The re-runnable regression is `tests/mcp/test_serve_auth.py`; the live proof is the A/B table above, run against a real `gflow serve` subprocess on both builds.

## Docs

`docs/MCP.md` and `docs/CONFIGURATION.md` described the token as bind-gating only, which is no longer true. Updated, with the website mirror regenerated.


ffroliva:
## Council review — D3 🟢 GREEN, D1 🟢 GREEN, D4 🟡 → applied

### D3 Security — GREEN, and it was attacked, not read

A **25-case bypass battery** against the PR head, with the app built through `build_app` and probes sent over raw ASGI scopes and `TestClient`: no header, empty value, scheme-only, wrong value, wrong length, `Basic`, raw token without a scheme, `X-Authorization` spoof, leading/trailing/double space, tab separator — **all 401, inner app never entered**. `bearer`/`BEARER` accepted (RFC 9110 correct).

Also confirmed by measurement, not inspection:

- **Ordering holds.** Unauthenticated + bad `Host` → 401 (not 421); unauthenticated + bad `Origin` → 401 (not 403). No transport-security leak.
- **No route escapes** — the HTTP app has exactly one route; `/`, `OPTIONS`, `DELETE`, `/health`, `/.well-known/oauth-protected-resource` all 401.
- **No oracle** — identical body, status and headers for "no header" vs "wrong token" vs "wrong length".
- **The inverse hole does not exist.** The startup gate exempts only the literal `127.0.0.1`; `0.0.0.0`, `::`, `::1`, `localhost`, `127.0.0.2`, `::ffff:127.0.0.1`, a bare hostname and `""` all demand a token.
- **Empty token is fail-closed** — `SecretStr("")` is falsy, so it reads as unconfigured *and* the non-loopback bind is still refused.

### D4 Tests — YELLOW, three findings, all applied in `dd823b53`

The important one: every authenticated-path assertion used an MCP `ping`, which the transport rejects **400 Missing session ID** regardless of auth — so a middleware returning 500 for a *valid* token passed all of them. The tests could see rejection working but not acceptance.

Control, against a middleware deliberately broken to 500 on a valid token (401 path and `compare_digest` call site left byte-identical):

| form | result |
|---|---|
| old `!= 401` | **5 passed** — the defect is invisible |
| new `== 200` | **6 failed** — every one catches it |

Also: the valid-token `GET /sse` case the file had recorded as untestable **is** testable — raising a sentinel out of the ASGI `send` at `http.response.start` tears the accepted stream down once its headers are final. And `test_http_transport.py` now strips `GFLOW_CLI_DAEMON_TOKEN` / `GFLOW_DAEMON_TOKEN` from the subprocess env, which this PR just made load-bearing.

20 passed in `test_serve_auth.py`, 213 in `tests/mcp/`.

### Noted, not fixed

`ui/app.py` mounts `server.sse_app()` at `/mcp`, bypassing `build_app`. It is unreachable — its own comment records that `run_server` has no caller — so it is left alone with a note naming `build_app` as the way back in, rather than wiring up a second auth surface nobody calls.

### Two nice-to-haves not taken

- `if token is not None` → `if token:` would close `token=""` authenticating a bare `Bearer `. Unreachable through config (proved above), so it guards only a future direct caller.
- An ASCII validation on the token at startup: a token beyond U+00FF locks out entirely rather than failing loudly. Fail-closed, so not urgent.

Issue #601, opened 2026-08-28
Onboarding: walk install → auth → first command → failure as a stranger would


ffroliva:
## Where the feedback came from

u/_suren, replying to the gflow-cli post on Reddit:

> Not hands-on yet, so I would treat my note as UX feedback from the post, not a usage review. If I were testing it, I would start with install -> auth -> one Veo/Imagen command -> failure/retry path, because that is where CLI tools usually lose first-time users.

They were explicit that this is an outside-in read, not a usage report. It still lands, because it names the exact stretch we have never walked as a stranger would.

## The ask

Walk the first-run path end to end **as a new user with no context**, and fix what loses them:

1. `pip install gflow-cli` — does the install alone leave a usable state, and is the next step obvious from the output?
2. `gflow auth login` — is it clear which browser opens, what the user must do in it, when it is finished, and what to do if the window closes early?
3. **One** generation command — can someone get a first image without reading `docs/USAGE.md` first? Is `--profile` discoverable when more than one profile exists (today a bare command with several profiles prompts, and in a non-interactive shell that is a dead end)?
4. The failure/retry path — when a run aborts (exit 23 selector drift, exit 28 arm unavailable, exit 11 profile lease, quota exhaustion), does the message tell a first-timer what to *do*, or only what happened?

## Why this is worth doing

This is the third independent signal pointing at the same stretch. Standalone repo-share posts were removed from r/commandline and r/Python, and the read at the time was that the product signal worth chasing is onboarding — install → auth → first command → failure UX — rather than more distribution. A stranger just described that same path unprompted.

The errors themselves are in good shape after the recent releases: they name a probe, carry a screenshot, and abort pre-submit at zero credits. What has never been tested is whether the *sequence* reads well to someone who has not read the codebase.

## Suggested shape

- A single "first five minutes" pass, recorded verbatim — every command a newcomer would type, and what they actually see.
- Fix only what blocks or misleads on that path; resist widening it into a docs rewrite.
- Keep the disclaimers where they are (README) — this is about the path, not the positioning.

## Not in scope

Anything requiring the reporter to test for us. They said plainly they are not hands-on; this issue stands on its own reasoning.

_Opened from community feedback. No timeline or commitment is implied._

PR #815, opened 2026-09-14
feat(plugin): installable Claude Code plugin, with a curated skill payload

ffroliva:
## Summary

Makes gflow-cli installable from a plugin marketplace in one step:

```
/plugin marketplace add ffroliva/gflow-cli
/plugin install gflow@gflow-cli
```

That installs the `gflow-cli` and `video-production` skills **and** registers the MCP server.

## The bug it also fixes

`.codex-plugin/plugin.json` declared `"skills": "./skills/"` and `.agents/plugins/marketplace.json` declared `path: "./"` — so both shipped **all eighteen** skills to every Codex and ChatGPT-desktop user, including `release`, `check`, `pr-council-review`, `sonar`, `doc-review` and `issue-resolve`. Those drive *this repo's* development lifecycle. An agent that reads `release/SKILL.md` will happily try to cut a release of whatever project it is pointed at.

The plugin was designed to avoid that on the Claude Code side while it was already live on the other two. All three channels now point at one curated payload under `plugins/gflow/`.

## Generated, not copied by hand

`scripts/ci/generate_plugin_skills.py` produces `plugins/gflow/skills/` from `skills/`, and `--check` runs in CI — the same contract `generate_website_docs.py --check` already provides for `website/docs/`.

It rewrites **only** links that climb out of the repo (`](../../KNOWN_ISSUES.md)` → an absolute URL). Links inside a skill (`](composition.md)`) and between the two shipped skills (`](../gflow-cli/SKILL.md)`) still resolve under the plugin's own layout and are left exactly as they are.

## Spend consent — the part that needed care

Claude Code's docs are explicit: *"When you enable a plugin, Claude Code starts its MCP servers automatically."* There is **no separate consent prompt** for a plugin-bundled server. Ours drives the user's own signed-in Google account and Veo video bills their credits.

So the plugin ships `defaultEnabled: false` with a required billing acknowledgement. Observed at install:

> `Successfully installed plugin: gflow@gflow-cli (scope: user). This plugin is disabled by default — enable it with: claude plugin enable gflow@gflow-cli`
> `1 userConfig option not yet set (1 required) — run /plugin configure gflow@gflow-cli in Claude Code, or pass --config KEY=VALUE.`

**Honest caveat, strengthened after review:** `acknowledgeBilling` has exactly one occurrence in the tree — its own declaration — and `claude plugin enable` succeeds without it being set. So `defaultEnabled: false` is the only control that actually holds; `userConfig` is a UI prompt. The README and `docs/MCP.md` now say this plainly and point at `--no-spend` as the hard guarantee, since it never registers the credit-spending tools at all.

**Which revision users get.** `/plugin marketplace add ffroliva/gflow-cli` reads the repository's **default branch**, and this project's default is `develop`. So the plugin tracks the integration branch while its manifest version and the rewritten skill links both name the last release — three snapshots in one install. Documented in `docs/MCP.md` rather than papered over; pinning the marketplace source to a tag is the fix and needs the plugin to exist on `main` first.

## Live verification — clean `CLAUDE_CONFIG_DIR`

| step | result |
|---|---|
| `claude plugin validate ./plugins/gflow` | `✔ Validation passed` |
| `claude plugin validate .` (marketplace) | `✔ Validation passed` |
| `marketplace add` → `install` → `enable` | ok, ok, ok |
| installed tree | 9 files, **2** skills |
| maintainer-only skills present | **0** (checked all 16 by name) |
| `mcpServers` | `{'gflow': {'command': 'gflow', 'args': ['mcp', 'run']}}` |

## That install run found a real bug in my own generator

`--check` was green while `skills/release/` sat in the installed tree: the generator deleted stray *files* but left their empty *directories*, and `_strays` only looked at files. An empty directory ships no instructions, but a maintainer-only skill **name** in a user-facing plugin is exactly what this guard exists to prevent. Directories now count as strays. A/B'd both ways — sabotage caught (exit 1), regenerate removes it, `--check` green.

This is why the live install mattered: the gate said in-sync and was wrong.

## Test plan

- [x] `tests/test_plugin_manifests.py` — 15 cases: required fields, reserved marketplace names, source resolution, version lockstep, curation, byte-identity with `skills/`, link rewriting, the declared command being a real console script, and both spend-consent flags.
- [x] **Negative control** — flipped `defaultEnabled` to `true`, removed `userConfig`, and reverted Codex to `./skills/`: 3 tests failed, exactly the 3 that should. Not vacuous.
- [x] `tests/scripts/ tests/test_plugin_manifests.py tests/test_documentation_gate.py` → 347 passed.
- [x] Generator A/B: hand-copied `release` skill → `STRAY` (exit 1); hand-edited generated file → `DRIFT` (exit 1); regenerate → in sync.
- [x] `ruff check` / `ruff format --check` clean; repo hygiene, doc links, PII guard, website mirror, council memory all green.
- [x] `tests/scripts/test_workflow_hardening.py` green after the CI edit.

## Two pre-existing assertions changed, deliberately

`tests/scripts/test_codex_plugin.py` asserted `manifest["skills"] == "./skills/"` and `source.path == "./"`. Those assertions **pinned the bug** — they enforced shipping maintainer-only skills to users. Updated, with the reason inline in the test.

## Non-breaking contract

- [x] No existing default changes — the plugin is new, and ships disabled.
- [x] No exit code changes meaning.
- [x] No CLI option removed or renamed; no CLI surface touched at all.
- [x] MCP tools unchanged — no schema, payload key or `tools/list` default touched.
- [x] No SQLite migration.
- [x] `--json` unchanged; `--dry-run` untouched.
- [x] Locale-invariant: no DOM selector and no UI number parsing in this change.
- [x] Pre-existing suite passes, with the two exceptions explained above.

## E2E / Iron Law

No Flow surface is touched — this is packaging plus two manifests, with no browser, profile or generation path involved. The re-runnable regressions are `tests/test_plugin_manifests.py` and the `--check` gate now in CI; the live proof is the clean-config install table above.

Refs #601


ffroliva:
## Council review — D3 🟡, D14 🟡 → applied in the follow-up commit

### The two that changed the shipped behaviour

**1. `acknowledgeBilling` has no consumer.** `git grep acknowledgeBilling` at the reviewed SHA returns exactly one hit: its own declaration. Combined with `claude plugin enable` succeeding without it set, calling it something the plugin *"asks you to acknowledge before it runs"* was not true. README and `docs/MCP.md` now say plainly that `defaultEnabled: false` is the control that holds and `userConfig` is a UI prompt — and point at `--no-spend` as the hard guarantee, since it never registers the credit-spending tools at all.

**2. The marketplace serves `develop`.** `gh repo view --json defaultBranchRef` → `develop`. So `/plugin marketplace add ffroliva/gflow-cli` hands a user integration HEAD while the manifest says `0.74.0` and the rewritten skill links point at `blob/main/` — three snapshots in one install. Documented in `docs/MCP.md` rather than papered over. Pinning the source to a tag is the real fix and needs the plugin to exist on `main` first.

### D14 — the diff is 78% generated payload

1913 of 2467 lines are the mechanical skill copy; net new logic is ~137 lines. The reviewer judged the generator justified over a check-only gate because it owns the link rewrite, so `skills/` stays the one editable copy.

Applied:
- **The byte-identity test restated the rewrite rule inline** — a third copy that could drift while the test still passed. It now calls the generator's own `render()` through a shared loader.
- **`plugins/gflow/.claude-plugin/plugin.json` was a fifth version declaration nothing bumped.** `RELEASE.md` step 6, its `git add` line, and `skills/release/SKILL.md` now name it, so the next release has a documented bump site instead of discovering a red test.
- Dropped `test_codex_version_tracks_pyproject` — byte-identical duplicate of the assertion in `tests/scripts/test_codex_plugin.py`.

### Confirmed safe, by measurement

- Payload is 10 files, **zero** maintainer-only skills. Grepped for `C:\Users`, `/home/`, `oss-ops`, `AIza|ya29|sk-|Bearer|SAPISID`, `.env.local` — no hits.
- **The generator cannot write or delete outside `plugins/gflow/skills/`.** `SHIPPED` is a hardcoded tuple, deletions come only from `rglob` under that root, and a symlinked directory makes `rmdir()` raise rather than deleting a target. No traversal.
- All three manifests verified curated at the reviewed SHA.

### Not taken

Dropping `_RESERVED_MARKETPLACE_NAMES` and the byte-identity test as redundant with `claude plugin validate` / the CI `--check`. Both are cheap, and neither `claude plugin validate` nor the CI gate runs inside `pytest` — a local test run should still catch a stray.

PR #817, opened 2026-09-14
fix(chain): install Pillow with the extra, and fail before the first paid clip

### LINKED ISSUE #813: gflow video chain fails on a clean install: gflow-cli[chain] does not install Pillow
## Summary

`gflow video chain` cannot run on a clean install, including one that explicitly asks for the documented `[chain]` extra. The extra installs `av` but not Pillow, and `src/gflow_cli/media.py` imports `PIL` at module level.

The user is not told to install anything — the error funnel reports a generic *"Unexpected error… file a bug"*.

## Reproduction (against released 0.74.0, no dev group)

```
$ echo '{"prompt":"a","model":"veo_3_1_fast"}' > one.jsonl
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" gflow video chain one.jsonl --dry-run
{"exception_class": "ModuleNotFoundError", "cli_command": "video chain", "event": "error_unhandled", "cli_version": "0.74.0", ...}
Unexpected error. Re-run with GFLOW_CLI_DEBUG_TRACEBACK=1 to see the real
error. If this persists, file a bug at
https://github.com/ffroliva/gflow-cli/issues.
$ echo $?
1
```

Which dependency is missing:

```
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" python -c "import av"    # OK
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" python -c "import PIL"
ModuleNotFoundError: No module named 'PIL'
```

`--dry-run` is used above so the reproduction costs nothing; the same failure happens on a real run.

## Cause

- `pyproject.toml` declares `chain = ["av>=12"]`. Pillow is declared only in `[dependency-groups] dev`, so it is present in a development checkout and absent in every user install. That is why this has never been seen locally.
- `src/gflow_cli/media.py:26` does `from PIL import Image` at module level.
- `src/gflow_cli/cli_video.py:942` (`from gflow_cli import chain as chain_mod`) is the **first statement** of `_run_chain`, and `chain.py:53` imports `media`. So the import fails before the manifest is parsed, before `--dry-run`, before the confirm prompt and before any client is created.
- `media` has exactly one importer (`chain.py:53`) and `chain` exactly one (`cli_video.py:942`), so nothing else is on this path — `gflow --help` and every other command are unaffected.

## The related half: `av` fails *after* you have paid

`av` is imported lazily inside `_decode_frame` (`media.py:80`), and the frame extractor runs after a link completes. So a user who is missing `av` gets a successful, billed link 0 and then a crash. Same bug class, worse consequence — both should fail at the same early point.

## Proposed fix

1. Add `pillow>=12.3.0` to the `chain` extra.
2. Move `import av` to module level in `media.py`, beside the existing `from PIL import Image`, so both optional dependencies fail at the same point.
3. Wrap the import block at `cli_video.py:942` in `try/except ImportError` and re-raise as the existing `FrameExtractionError` (exit 20) with remediation naming `gflow-cli[chain]`, `av` and `pillow`.

Rejected: a lazy `PIL` import plus an `importlib.util.find_spec` pre-flight. It makes the import late, then adds a second mechanism to restore the earliness it gave away, carrying a hardcoded package tuple that can drift from the extra.

## Regression guard

CI's `resolve-drift` job installs without the dev group but only smoke-imports the MCP surface. It gains `uv pip install .[chain]` followed by `python -c "import gflow_cli.media, av, PIL"` — one line that fails today and cannot pass while a package the chain path imports is missing from the chain extra.

## Affected versions

Every version with the `[chain]` extra. Confirmed on 0.74.0.

ffroliva:
## Summary

`gflow video chain` could not run on a clean install — including one that asks for the documented `[chain]` extra. Reproduced against released 0.74.0:

```
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" gflow video chain one.jsonl --dry-run
{"exception_class": "ModuleNotFoundError", "cli_command": "video chain", "event": "error_unhandled"}
Unexpected error. Re-run with GFLOW_CLI_DEBUG_TRACEBACK=1 ... If this persists, file a bug
$ echo $?
1
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" python -c "import av"   # OK
$ uvx --isolated --from "gflow-cli[chain]==0.74.0" python -c "import PIL"  # ModuleNotFoundError
```

The user is told to file a bug, not to install anything.

## The fix — three small changes

1. `pillow>=12.3.0` joins `av>=12` in the `chain` extra. It was declared only in the dev group, which is exactly why this was never seen in a checkout.
2. `import av` moves to **module level** in `media.py`, beside the existing `from PIL import Image`. It was lazy inside `_decode_frame`, which runs only *after* a link completes — so a missing `av` failed **after the user had paid for link 0**.
3. The import block at `cli_video.py:942` — the first statement of `_run_chain`, ahead of the manifest parse, `--dry-run`, the confirm prompt and any client — is wrapped in `try/except ImportError` and re-raised as the **existing** `FrameExtractionError` (exit 20), naming the extra and both packages.

**Rejected:** a lazy `PIL` import plus an `importlib.util.find_spec` pre-flight. It makes the import late, then adds a second mechanism to restore the earliness it just gave away, carrying a hardcoded package tuple that can drift from the extra it mirrors (+20 src lines vs +7), and it needs a test that would have passed *without* the fix.

## A second, independent bug found on the way

`_cli_helpers._handle_gflow_error` interpolated `detail` and `remediation_hint` into a **Rich markup** f-string unescaped. Rich reads `[chain]` as a style tag, finds no such style, and **silently drops it**. Verified directly:

```
UNESCAPED renders: "-> pip install 'gflow-cli'"
ESCAPED   renders: "-> pip install 'gflow-cli[chain]'"
```

So every `install gflow-cli[chain]` hint this CLI has ever printed rendered as `install gflow-cli` — advice that reinstalls what the user already has. `--json` was never affected, which is why it went unnoticed.

Fixed at both render sites (`_cli_helpers.py:310-312`, `cli.py:316`) rather than by rewording the new message, so it covers **every** error class, not just this one.

## A/B control — three arms, against real installs

| arm | result |
|---|---|
| **control** — 0.74.0 + `[chain]`, missing manifest | exit **1**, `ModuleNotFoundError`, "file a bug" |
| **A** — fixed tree + `[chain]` in a throwaway venv | all three import; encoded a real mp4 with `av` and ran `extract_last_frame` → 691-byte JPEG, magic bytes verified |
| **B** — fixed tree, **no** extra | exit **20**, and `gflow-cli[chain]` survives to the console — proving the escape fix in a real terminal, not just a test |

Arm B also shows the guard fires ahead of the manifest: the nonexistent file never mattered.

## Test plan

- [x] 9 new cases in `tests/cli/test_cli_video_chain.py`, proven red first. Each was checked against the false-green trap: there are **two** caches, so evicting `sys.modules` alone still passes — `from gflow_cli import chain` is satisfied by `_handle_fromlist`'s `hasattr` check and never reaches the finder. The helper evicts both.
- [x] `test_cli_video_chain.py` + `test_chain.py` + `test_media.py` + `test_errors.py` → 99 passed.
- [x] **Full suite** — 4330 passed, 7 skipped (run because `_cli_helpers.py` is the shared error renderer).
- [x] `ruff check` / `ruff format --check` clean. `pyright src` → 0 errors (via `uv run`; the venv invocation gives the documented phantom errors).
- [x] Repo hygiene, doc links, PII guard, website mirror, council memory green. `tests/mcp/test_cli_parity.py` → 7 passed.

## MCP parity

None needed, and that is recorded rather than assumed: `video chain` is an explicit CLI-only exemption in `test_cli_parity.py` ("long-running billed chain; not worker-enqueued"), so there is no twin. No CLI option was added, so no mirror axis moved.

## Two pre-existing assertions changed, deliberately

- `tests/test_media.py` asserted that `extract_last_frame` converted a missing `av` into `FrameExtractionError` **at call time**. It passed — but the behaviour it blessed *was* the bug, since `_decode_frame` only runs between links. Rewritten to the new contract, with the reason in its docstring.
- `tests/test_errors.py:690` pinned the old `_default_remediation` string verbatim, which said only "PyAV".

## Non-breaking contract

- [x] No existing default changes.
- [x] No exit code changes meaning — reuses `FrameExtractionError` / exit 20.
- [x] No CLI option removed or renamed.
- [x] MCP tools unchanged.
- [x] No SQLite migration.
- [x] `--json` only gains the typed envelope it should always have had; `--dry-run` stays browser-free and spend-free.
- [x] Locale-invariant: no DOM selector, no UI number parsing.
- [x] Pre-existing suite passes, with the two exceptions explained above.

## E2E / Iron Law — named blocker

This touches a Flow surface, so it owes an e2e run. `tests/e2e/test_chain_e2e.py` is marked `e2e_video`: it **spends Veo credits** and needs `GFLOW_CLI_E2E_RUN_VIDEO=1` plus a live profile. That is the named external blocker.

The three-arm A/B above is the closest equivalent that could actually be run, and it exercises the **real installed wheel** rather than mocks — but it is **not** claimed as satisfying the e2e rule. A maintainer run of `test_chain_e2e.py` is still owed.

Closes #813

## Summary by CodeRabbit

- **Bug Fixes**
  - Error messages now display bracketed text correctly instead of truncating it.
  - `gflow video chain` reports missing video dependencies consistently with exit code 20 before reading manifests, showing dry-run plans, or requesting confirmation.
  - Error guidance now identifies the required optional extra and packages.

- **New Features**
  - The `chain` extra now includes Pillow alongside PyAV.

- **Documentation**
  - Updated usage, configuration, and known-issues documentation to reflect dependency requirements and failure behavior.


ffroliva:
## Council review applied — the escape fix was half-applied

The council caught that the previous commit fixed the two sites rendering the *new* chain message and missed the rest. That is the [[review-findings-get-half-fixed]] shape, and it mattered: a bracket-bearing hint already exists on a missed path — `api/_engine.py` tells the user to `uv pip install gflow-cli[patchright]`, and `cli.py` rendered that `ConfigurationError.detail` unescaped.

Proven against the real in-tree hint:

```
UNESCAPED has '[patchright]': False
ESCAPED   has '[patchright]': True
```

**Fixed:** every site interpolating error text into Rich markup, across `_cli_helpers.py`, `cli.py`, `cli_movie.py`, `cli_run.py`, `cli_scene.py`.

**The durable part:** `tests/test_rich_markup_safety.py` walks the AST of every module under `src/gflow_cli` and fails on a `console.print` f-string interpolating exception text without `escape()`. It carries its own negative control, so it cannot quietly become a no-op.

Scoped to error text (`exc`, `.detail`, `.remediation_hint`) rather than every interpolation — any interpolated value can lose a `[...]`, but error text is the class that produces wrong *instructions*. Two rounds of tightening removed six false positives: `.title` matches a tool spec, a project record and a movie manifest, and `type(exc).__name__` cannot carry a bracket.

661 passed across `tests/cli/`, `test_errors`, `test_media`, `test_chain` and the guard; `ruff check` / `format --check src tests` clean.

### Also from the council, not taken (with reasons)

- **`FrameExtractionError` reuse** — reviewer judged it defensible and I agree: the class already documented the missing-extra case, the `type` URI is unchanged, and the only action a caller can take either way is "install the extra". A new class plus exit code for one message is the D14 smell. The `title` being "Last-frame extraction failed" at a point where nothing was extracted is fair; noted as a follow-up rather than churned now.
- **Narrowing `except ImportError`** to `exc.name in {"av", "PIL"}` — real point (the block wraps five imports, two of which are not `[chain]` deps). Left as-is because `detail=f"…: {exc}"` preserves the real message, so a mis-labelled circular import is still diagnosable. Worth doing; not worth reopening a green PR for.

PR #831, opened 2026-09-15
feat(migrated): port --end-frame to the migrated host (#639)

stgmt:
Ports `--end-frame` (image-to-video start+end interpolation) to the migrated host (`flow.google.com`). Refs #639.

> `Refs`, not `Closes`: #639 is the umbrella migration issue, and scenes, extend, instructions, tools and project creation are still unported there.

## What the app does

A bound end frame switches the submit to a different contract — rpc `nprQif`, an interpolation model key, and both frame ids in the body. Watching only the classic submit rpcs made every start+end run time out **after a successful server-side submit**, so the credit was spent and the CLI reported failure.

## Changes

- `attach_end_frame`: mirror of the Start path (toolbar upload → pick by name → media id). The End chip is required empty-then-bound, with two bound chips read back, so a silently unbound Start surfaces here instead of mis-binding End.
- Dispatch gate serves local end frames on the migrated host. Frames by UUID or `@Name` still exit 36.
- `submit_and_observe` adds `nprQif` to the watched rpcs, asserts the interpolation key and both frame ids when an end frame is bound, and reads the bare `batchexecute` reply the interpolation rpc answers on.

## Verification

**Contributor (@stgmt), veo cohort:** live paid run — submit on `nprQif` (status 6), terminal status 3, 8s 720x1280 mp4 downloaded (1.3 MB).

**Maintainer, second cohort (`ffroliva`, `flow.google.com`), 2026-09-17, $0:** both frames really uploaded and bound, submit intercepted with `route.abort()` so the request never reached Flow. Confirms the port and corrects two claims:

| | measured | vs. original PR |
|---|---|---|
| rpcid | `nprQif` | confirmed |
| both media ids in body | yes | confirmed |
| model key | `omni_flash_i2v_8s_first_last` | **not** `veo_3_1_interpolation_lite` |
| rpc name carrier | URL `rpcids` param | **not** `f.req` |

The key is cohort-dependent, so matching is on key **shape** (`*_interpolation_*` or `*_first_last`) — pinning either literal would refuse a valid run on the other cohort. `_body_rpcid` is kept as a second carrier, no longer described as the rule.

Findings: `docs/superpowers/spikes/2026-09-17-migrated-end-frame-submit-contract.md`.

## Maintainer follow-up in this PR

Review found items that were taken over here rather than handed back:

- **Dropped adoption-by-content.** Its predicate required `_i2v_` in the model key, which `veo_3_1_interpolation_lite` does not contain — so it could never fire for the veo case it was written for. Against the omni key it matched broadly enough to adopt any future rpc echoing composer state, and a parse failure on an adopted id is fatal rather than skipped: that trades a named timeout for a hard failure on an already-billed run. Both measured submit rpcids are in `SUBMIT_RPCS`.
- **Body assertions are always armed, including t2v.** The interpolation guard is a *billing* guard that matters precisely when nothing was asked for — a stale End chip makes the app submit `nprQif` on a run that requested no end frame.
- **Spike script could not run.** It raised `NameError` on its first statement (`ruff F821`, two hits) and had no network capture, so it cannot have produced the evidence above. It now navigates via `ensure_editor`, writes evidence in a `finally`, takes required `--profile`/`--project`, and gained the `--mode submit` route-abort probe.
- **`attach_end_frame` had zero test coverage.** The shared fake modelled one chip, which made the End path unmodellable; it now counts two, and all existing chip tests stay green.
- **BDD scenarios passed with the feature deleted** (verified by ablation — the ablated run also submitted). `len(picked) == 2` is the discriminator; ablation now fails them.
- **An existing e2e would have billed and then failed.** `test_e2e_i2v_start_end_frame_flags` asserted the labs-only `ui_automation_video.frame_attached`, while both maintainer accounts are served `flow.google.com`, which logs `migrated.frame_bound`. It accepts either driver now.
- **New zero-credit e2e** pins the whole contract: `test_e2e_end_frame_binds_the_second_chip_and_submits_interpolation` — passes live in 38s, 0 credits.
- **Docs:** four sites plus three mirrors still said this exits 36 (USAGE x2, CONFIGURATION, INDEX, MCP, both video-production skill copies). CHANGELOG had the feature under `Fixed` with no user-visible entry.

## Gates

`ruff check` · `ruff format --check` · `pyright src` (0 errors) · doc links · website mirror · repo hygiene · council memory — all green. Offline suite: **4489 passed, 24 skipped**. New e2e passes live.


stgmt:
Rebased onto current `develop` and addressed all three CodeRabbit findings in `51e5a6c7`:

1. **Spike `sys.path`** — checkout `src` now inserted at index 0 instead of appended.
2. **End-chip threshold** — probe fires on any remaining empty chip (`if after["empty_count"]`), clicking the **last** empty chip; when Start is already bound the single remaining empty chip is End by construction (attach flow requires Start before End).
3. **Bare `batchexecute` responses** — the response filter now keeps non-`batchexecute` filtered while letting bare `batchexecute` responses (rpcid `None`) through to `parse_frames` for adopted RPCs. New regression test `test_adopted_rpc_with_bare_batchexecute_response_resolves` proves an adopted submit resolves off a bare-URL response.

Additionally, the interpolation validator is now key-shape based (`interpolation`/`first_last`) instead of pinned to `veo_3_1_interpolation_lite`, so omni's `omni_flash_i2v_4s_first_last` submits validate — verified live earlier today (4 omni seams generated and accepted; see sibling discussion). Test asserts both key shapes plus the rejection message.

Gates on this tree: repo-hygiene ✅, doc-links ✅, `ruff check` ✅, `ruff format` ✅, `pyright src` 0/0/0 ✅, `pytest tests/auth/ tests/api/` **1872 passed, 3 skipped** ✅, website-docs sync + PII ✅, CHANGELOG `Fixed` under `[Unreleased]` ✅.

stgmt:
Follow-up `92f0d79d` addresses the CI failures and the second CodeRabbit pass:

- **Page-pool deadlock (real bug, `test_every_page_borrower_returns_the_page`)**: the spike now pairs `_checkout_page`/`_checkin_page` via `try/finally` (`_capture_on_page` extraction). Good catch — the pool is size 1, this would have deadlocked any caller.
- **3 outdated BDD/parity tests**: `migrated_i2v.feature` end-frame scenarios + `test_migrated_i2v_parity.py` now pin the ported contract (migrated host serves local start+end; no exit 36, no labs fallback) instead of the pre-port refusal.
- **Stale capability docs** (CodeRabbit): `run_video` docstring was already fixed in `51e5a6c7`'s parent hunk; this commit updates `FlowHostMigratedError._default_remediation` + `KNOWN_ISSUES.md` (+ regenerated website mirror) to local start **(and end)** frames.

Gates on this tree: ruff check/format ✅, pyright 0/0/0 ✅, `pytest tests/auth/ tests/api/` **1872 passed** ✅, features+mcp+scripts **349 passed** ✅, repo-hygiene ✅, doc-links ✅, website-docs sync ✅, PII ✅.

ffroliva:
## Taken over and shipping — thank you, this was a good find

@stgmt — the port is right, and the part that mattered most is the part that is hardest to get: you noticed that the timeout was happening **after** a successful server-side submit. That is the difference between "the feature doesn't work" and "the feature works and bills you while reporting failure", and it is why `nprQif` was worth chasing at all.

Rather than send this back for another round, I have taken the remaining review items over and pushed them onto your branch. Your commits are intact underneath.

### I re-ran it on a second cohort, and it holds

Both frames uploaded for real, submit intercepted with `route.abort()` so nothing reached Flow — $0, on `ffroliva`/`flow.google.com`:

```
rpcid       nprQif                          CONFIRMED
both ids    present in the submit body      CONFIRMED
model key   omni_flash_i2v_8s_first_last    (yours: veo_3_1_interpolation_lite)
rpc name    URL `rpcids` param              (yours: f.req body)
```

Two corrections there, and neither is a mistake on your part — they are cohort differences, which is exactly the thing this project keeps relearning. Our accounts default to omni, yours to veo. The consequence is that matching has to be on key **shape**, not a literal: pinning `veo_3_1_interpolation_lite` would have refused every valid run on our side, and pinning ours would refuse yours. Your `first_last` branch already handled this — it just had no citation, and now it has one. `_body_rpcid` stays as a second carrier for the case you saw.

Findings are written up in `docs/superpowers/spikes/2026-09-17-migrated-end-frame-submit-contract.md`, so the next person does not have to re-derive them from a PR description.

### What I changed, and why

- **Dropped adoption-by-content.** Its predicate needed `_i2v_` in the model key, and `veo_3_1_interpolation_lite` does not contain it — so that branch could never have fired for the veo case it was written for. Against the omni key it matched broadly enough to adopt an unrelated rpc, where a parse failure is fatal rather than skipped, on a run that has already been billed. Both measured rpcids are in `SUBMIT_RPCS`, so nothing needs adopting.
- **The spike script could not run.** `url` is undefined at line 69 (`ruff F821`), so it raised `NameError` before its first `goto` — and it had no network capture at all, so it could not have produced the wire evidence either way. It now navigates through `ensure_editor`, writes evidence in a `finally`, takes required `--profile`/`--project` (the defaults pointed at your account), and has a `--mode submit` route-abort probe.
- **`attach_end_frame` had no tests.** The shared fake modelled a single chip, which is why — the End path was unmodellable. It counts two now.
- **The BDD scenarios passed with the feature deleted.** I checked by ablation; the ablated run also submitted. `len(picked) == 2` fixes it.
- **An existing e2e would have spent a credit and then failed** on a labs-only event name, since both maintainer accounts are on the migrated host.
- **New zero-credit e2e** pins the contract end to end: passes live in 38s.
- **Docs:** four more sites plus three mirrors still said this exits 36.

### One process note

The live run in the PR description was real, but it is not re-runnable by anyone else, and the committed script could not reproduce it. For a blackbox surface the evidence has to survive the PR — a `docs/superpowers/spikes/` note, or an e2e someone can run. Not a criticism of the work; it is the single highest-value habit in this repo.

Gates: `ruff`, `ruff format`, `pyright` (0 errors), doc links, website mirror, repo hygiene, council memory — all green. Offline suite **4489 passed**. Rebased onto current `develop`.

Merging once CI reports. Thanks again — this closes the last i2v form on the migrated host.

Issue #795, opened 2026-09-12
credits: labs-gated on migrated accounts, and the error blames SAPISID for a host it never contacted


ffroliva:
Split out of #791 so fixing the login oracle does not leave a green `auth status` next to a red `credits`.

## Two defects, one code path

`api/credits.py:51` probes the **labs** NextAuth session unconditionally:

```python
session_status, session_body, _ = await fetch_flow_session_httpx(profile_dir)
```

then extracts `access_token` (`:64-66`) before calling `routes.CREDITS` with hardcoded `origin`/`referer` of `https://labs.google` (`:80-81`).

### 1. It stays broken on migrated accounts even after login is fixed

For an account Google has moved to `flow.google.com`, labs returns `200 {}` with no `access_token`, so the run dies at `credits.py:67-72` regardless of how the profile authenticated. If #791's oracle lands without this, the product reports *"Flow session verified"* and then fails `gflow credits` on the same profile — an inconsistent state that will read as a regression.

### 2. The error blames something that is not the cause and a host that was never contacted

The raise is `AisandboxAuthError(detail="no access_token in Flow session", route="auth/session")`, but the rendered title is **"aisandbox-pa authentication failed"** (`errors.py:196`) and the remediation says **"SAPISID cookie missing, expired, or unreadable"** (`errors.py:198`).

At that point aisandbox-pa has not been contacted at all, and SAPISID is present and fine — it is what made the outcome `GOOGLE_SESSION_ONLY` rather than `NO_SESSION` in the first place. A user following that remediation re-authenticates, which cannot help.

Reported downstream in #791: *"`gflow credits list` → `aisandbox-pa authentication failed`"*.

## Suggested shape

- Distinguish "the labs session BFF returned no token" from "aisandbox-pa rejected us" — they are different failures with different remediations, and only one of them is about SAPISID.
- Decide what `credits` means for a migrated account before wiring a new probe: the Veo balance surface for that cohort has not been located yet, so the honest interim may be a typed "not available on this host" rather than an auth error.

## Surfaces

`gflow credits user` / `gflow credits list`, and the MCP twin if one exists — check the six mirror axes.

Refs #791

ffroliva:
Half fixed on `develop` in #797 (`4d4ea3df`).

[...]

Issue #791, opened 2026-09-12
auth: login verification never passes for accounts migrated to flow.google.com (labs NextAuth absent → GOOGLE_SESSION_ONLY forever)


ai4U23:
# Upstream issue draft 1 — gflow-cli: login verification gap for migrated accounts

**Repo:** ffroliva/gflow-cli · **Title:** `auth: login verification never passes for accounts migrated to flow.google.com (labs NextAuth absent → GOOGLE_SESSION_ONLY forever)`

## Environment
- gflow-cli 0.73.1 · Python 3.12.12 · macOS 15 (arm64) · system Chrome channel
- Account: Google AI **ULTRA**, **server-migrated** to `flow.google.com` (labs.google instantly hands off — the mechanism documented in your own spike `docs/superpowers/spikes/2026-09-04-migrated-host-handoff-mechanism.md`)

## Symptom
`gflow auth login --browser chrome` can never succeed for this account, across
5 attempts including agent-driven attempts where the Flow **editor** (composer +
projects) was verifiably loaded in gflow's own Chrome window before closing:

```
cookie_decryption_failed_falling_back_to_playwright
playwright_cookie_read  all_cookies=47  flow_domain=0  flow_path_root=0
outcome=google_session_only  auth_chrome_marker_rolled_back
Authentication credential missing: Signed in to your Google account, but the
Flow app sign-in wasn't completed.
```

Downstream: `gflow auth status` → "No sign-in detected"; `gflow credits list` →
`aisandbox-pa authentication failed`; every generation command → 401.

## Root cause
`auth/verification.py` probes `https://labs.google/fx/api/auth/session`
(NextAuth) as the sole oracle. For accounts Google has migrated to
flow.google.com, that session is **never minted** — the labs app hands off
before/at the NextAuth callback; `flow_domain: 0` (no labs.google cookies at
all). Meanwhile the migrated host authenticates exactly as your own
`migrated_composer` driver expects: `.google.com` SSO cookies (SAPISID/SID) plus
the migrated app's own session cookies (`__Secure-OSID`/`OSID` on
.flow.google.com). Proof the profile is fully usable: the Flow workspace and
composer load and generate normally in that profile via the UI.

## Fix we applied locally (fail-closed) — happy to PR if useful
In `verify_flow_profile`, when the labs oracle yields `GOOGLE_SESSION_ONLY`, run
a migrated-host fallback probe (Playwright persistent context, channel chrome):

[...]

PR #835, opened 2026-09-15
fix(auth): verify migrated sessions against flow.google.com (#791)

ffroliva:
## Summary

Accounts Google moved to `flow.google.com` **whose labs session is no longer minted** could not sign in at all. Verification used exactly one oracle — `labs.google/fx/api/auth/session` — which answers `200 {}` forever for them, so a perfectly usable Flow workspace reported *"Signed in to Google, but not to the Flow app"* and gflow refused every command. A total lockout, not a degraded feature.

When labs declines **and** the profile carries a `flow.google.com` app-session cookie, gflow now confirms the session against the host that actually serves the app.

The cookie only **gates** that check; it never decides one. A cookie on disk outlives a password change or a "sign out of all devices", so treating its presence as proof would report a dead session as live — the exact failure `/gflow:predict` called out.

## Why it costs a browser — measured, with an anonymous control

The predict verdict asked for a probe that is both server-attested *and* needs no browser. I measured whether that exists. It does not:

| probe | authenticated | anonymous | separates? |
|---|---|---|---|
| `GET flow.google.com/` | 200, 152719 B | 200, 152680 B | no — 39 B |
| `GET /tools/flow` | 200, 152758 B | 200, 152711 B | no — 47 B |
| `batchexecute?rpcids=jwpduf`, no `at` | **401** | **401** | no |
| `SNlM0e` / any XSRF token | absent | absent | no |
| **rendered `signout_link` / `signin_cta`** | **1 / 0** | **0 / 1** | **yes** |

The batchexecute row is the sharp one: carrying a **valid session** and no `at` token it returns 401, and so does the anonymous arm, two bytes apart. `at` is required for reads and is no longer obtainable without a browser, so the 401 reports the missing token rather than the session state.

And the trap this closes: **the sign-in link is in the raw HTML of *both* arms, and in the rendered DOM of only the anonymous one.** Angular ships one shell to everybody and decides after it boots, so any oracle that greps the response body reads the shell, not the session. That is why every browserless instrument came back flat — and why `#793`'s body-scraping approach could not have worked.

Evidence: [`docs/superpowers/spikes/2026-09-16-migrated-session-oracle-needs-a-browser.md`](docs/superpowers/spikes/2026-09-16-migrated-session-oracle-needs-a-browser.md), reproducible via `scripts/dev/spike_migrated_session_oracle.py`.

## Design

- `flow_host_session` rides on the existing `ChromeCookieSnapshot`, derived from the jar **both** readers already hold, and threads through the **pure** `evaluate_session_response` as `migrated_probe_warranted` — so both oracle call sites gate identically without duplicating the rule. **`--browser internal` is covered too**, which #793 left broken.
- The trigger includes `VERIFICATION_ERROR`, not just `GOOGLE_SESSION_ONLY`. Otherwise the fallback switches itself off on the day labs actually dies, which is the scenario it exists for.
- `verify_flow_session` probes in the context it **already holds**; only the httpx fast path launches anything, and only on the narrow path. Ordinary verification still launches nothing.
- **A probe failure keeps the labs outcome.** Letting it reach the outer handler would convert an accurate, actionable `GOOGLE_SESSION_ONLY` into `VERIFICATION_ERROR`, whose remediation says "check network connectivity" — the #795 pattern.
- A session verified this way carries **no email**; that host discloses none in either arm. `.gflow_account` is skipped and the `assert` requiring one is gone, rather than inventing an address. `profile_store` and the account chooser already have no-marker paths.
- Gated on the existing `GFLOW_CLI_FLOW_HOST` kill switch. **No new env var.**

## Test plan

- [x] `tests/auth/test_migrated_session_oracle.py` — **25 passed** (gate, trigger set, pure DOM reading with both ambiguous cases failing closed, kill switch, and the wiring)
- [x] `tests/auth` + `tests/api` + `tests/mcp` — **2064 passed, 3 skipped**
- [x] `pytest -m e2e tests/e2e/test_migrated_session_oracle_bdd.py` — **4 passed in 44.85s**, live flow.google.com, $0
- [x] `pytest -m e2e tests/e2e/test_auth_verification_e2e.py` — **4 passed in 17.02s**, the regression proof predict named: a migrated-but-labs-alive profile is unchanged
- [x] ruff, ruff format, doc links, repo hygiene, website mirror + PII
- [x] pyright: no error in any changed file

One note on the e2e run: an earlier pass errored on `test_e2e_verify_flow_profile_authenticated` with `FileExistsError` on `tmp/pytest`. That was two pytest processes colliding in one worktree, not a regression — re-run alone it passes, which is what the 17.02s figure above is.

## Not verified here

The rescue itself — labs dead, probe succeeds, login completes. Every account on this machine is migrated **with labs still alive**, so we are not in the reporter's cohort. That is a named external blocker under the Iron Law, so this is **`Refs #791`, not `Closes`**, with @ai4U23 / @Cstanish127 as verifiers.

## On #793

@ai4U23 filed the first fix and found the mechanism. The measurement above moved the design away from their patch — a body scrape for a `gmail.com`-only email regex cannot work now that the body carries no identity in either arm, and their fallback also left `--browser internal` unfixed. Credit for the diagnosis is theirs; #793 should close in favour of this once they have had a chance to look.

Refs #791

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login verification for migrated Google accounts when the Labs session service is unavailable or returns an error.
  * Added a browser-based check of the Flow session when an appropriate Flow cookie is present.
  * Prevented stale or unrelated Google cookies from being treated as proof of authentication.
  * Successful verification without an available email no longer fails or records an incomplete account.

* **Documentation**
  * Added documentation describing the migrated-account verification limitation and fallback behavior.


ffroliva:
## ⛔ STOP — I am converting this to draft. The central claim is refuted.

The e2e went red, and the reason invalidates the design rather than the test.

### What happened

| time | `profile_ffroliva` renders |
|---|---|
| ~00:05 | `signout_link=1, signin_cta=0` at `/` |
| ~01:05 | `signout_link=0, signin_cta=1` at `/about` |

Cookies still on disk (`has_flow_osid: true`), labs still returning a 691-byte authenticated session **with an email**. `denon82` and `promo-denon82` render the same anonymous shape. Headed and headless, stealth flags on and off — **identical in all four arms**, so it is neither bot detection nor headless.

### Why it is not a dead session

It is the **`/about` redirect** — known, account-scoped, and **already measured in this repo** as [#756](https://github.com/ffroliva/gflow-cli/issues/756) / [`2026-09-11-about-redirect-is-decided-client-side.md`](docs/superpowers/spikes/2026-09-11-about-redirect-is-decided-client-side.md). I did not read it before designing around it. That spike settles the mechanism:

- The hop is **client-side**, 192 ms after the app's own first navigation. No 3xx anywhere.
- The failing arm's **only** request to `flow.google.com` is the document. **Zero `batchexecute`, zero of anything else** — never asked, not asked-and-refused.
- Decisively: *"`gflow project list` on `denon82` returns 50 projects **including** the one that redirects. The backend grants access while the frontend declines to open it."*

### Consequence

**The rendered DOM cannot attest to the session.** The app decides what to render without consulting the server about auth. `signout_link == 0` means *"this account is in the /about state"*, not *"this session is dead"* — and #756 exists precisely because those differ.

A probe reading it would report an authenticated user as logged out whenever their account is in that state. That locks them out of gflow **exactly as #791 does**. The remedy reintroduces the disease.

### What survives

| | |
|---|---|
| Q1 / Q2 / Q2b — no browserless oracle; `batchexecute` **401 with a valid session**; `SNlM0e` gone; no identity on the host | **stands** — plain HTTP, unaffected |
| Q4 — the rendered DOM separated the arms | stands as a **correlation**, one account, one hour |
| Q4's mechanism — "server-attested, so it sees revocation" | **refuted** |
| Q5 — the warm-cache control "proves" attestation | **refuted as proof**: a cookie-less copy rendering anonymous is equally explained by the client deciding anonymous |

### Method note

Rung 1 of the spike ladder is *"read an existing capture — free"*. Two spikes named `about-redirect` were sitting in `docs/superpowers/spikes/`. I wrote a new probe instead, measured a real correlation, and asserted a **mechanism the data never showed**. Eight council dimensions did not catch it. The e2e did — by going red on a live account whose state moved underneath it. That is the argument for the Iron Law in one incident.

### Where a replacement should start

Any candidate must be checked against an account **in the `/about` state** before it is believed, because that state silently mimics logged-out. The one signal known to survive it is the backend — `project list` works on a redirecting account — so a `batchexecute` read is the lead. Q2b showed it needs an `at` token that is not browserlessly obtainable, so it would have to be issued from inside a booted app page.

Not closing this and not deleting the branch: the HTTP findings, the cookie gate, the fail-closed plumbing and the council fixes are all worth keeping for whoever picks it up. Refs #791, #756.

ffroliva:
## Re-reading the retraction: the stated reason does not hold, but a different one might

Read at `230ee0be`. The retraction commit says:

> A probe reading it would report an authenticated user as logged out whenever their account is in that state — locking them out of gflow exactly as #791 does. The remedy reintroduces the disease.

**That is not what this implementation does.** The oracle is structurally incapable of downgrading a verdict:

- `migrated_probe_warranted` (`verification.py:108-123`) fires only when `outcome in {GOOGLE_SESSION_ONLY, VERIFICATION_ERROR}` — states that are *already failing*. `AUTHENTICATED` never re-probes, `NO_SESSION` never probes.
- A negative returns `None` (`:468-475`), and the caller then falls through to `return result` (`:598-610`) — the original verdict, untouched.
- An exception also returns `None` (`:462-467`), with the comment "a probe that cannot run must never upgrade an outcome".
- The single positive path is an upgrade to `AUTHENTICATED`.

So in the `/about` state the oracle produces a **false negative**: the account stays `GOOGLE_SESSION_ONLY`. That is precisely the status quo of #791 — the probe fails to help, it does not harm. No user who passes today would be locked out by this.

The `/about` finding is still correct and still important. What it invalidates is the *claim* that the rendered DOM is server-attested — not the safety of this code path.

## The reason that may actually block it

The hazard runs the other way, and the code's own docstring gestures at it ("revoked session read as live — the one failure this oracle…"):

If the app decides what to render **client-side, without consulting the server** — which is exactly what `2026-09-11-about-redirect-is-decided-client-side.md` measured — then a profile whose session has been **revoked** (password change, "sign out of all devices") but whose cookies are still on disk would plausibly still boot into the authenticated shell. `signout_link=1`, oracle says signed in, gflow upgrades a dead session to `AUTHENTICATED`, and the user gets a confusing downstream failure instead of a clean "please log in".

That is the failure mode #791's own design notes warned about, and it is the one the `/about` measurement makes *more* likely rather than less: a client that renders without asking cannot distinguish live from revoked.

**I have not measured this** — it is an inference from the client-side-rendering mechanism, not an observation, and I am labelling it as such rather than grading it "likely" and moving on.

**What would settle it:** take a profile with `flow.google.com` cookies on disk, revoke the session server-side (Google account → Security → sign out of all devices), do **not** touch the cookie jar, and render. If `signout_link` is still 1, the oracle is unsafe in the direction that matters and the retraction's conclusion stands on firmer ground than its reasoning. If it drops to 0, the oracle survives the one case it needs to survive, and the remaining problem is only the `/about` false negative — which is a missed upgrade, not a regression.

## Suggested disposition

Not close, and not merge as-is. Three separable pieces:

1. **The spike measurements land regardless.** Q1/Q2/Q2b are plain HTTP and stand on their own: no browserless oracle exists, `batchexecute` returns 401 with a valid session and no `at`, `SNlM0e` is gone, the host carries no identity. That is a durable negative result worth having in the tree whatever happens to the code, and it is what stops the next person re-deriving it. The `2026-09-16` spike already records it.
2. **The code waits on the revocation control above.** One measurement decides it.
3. **The `/about` false negative is worth stating as a known limit** either way — an account in that state gets no upgrade, so #791 remains open *for that cohort* even if this merges.

Also worth carrying forward: the retraction's own method note — "rung 1 of the spike ladder is read an existing capture — free" — is the transferable lesson here, and it is a better one than the verdict it was attached to.

Refs #791, #756

ffroliva:
Superseded by #793, which shipped in v0.77.0 — that PR fixes verification for accounts Flow serves from flow.google.com, and the released binary was verified against a Workspace account. Closing this retracted draft; the remaining half of the story (the sign-in window not closing for those same accounts) is #849, fixed in #851.

PR #855, opened 2026-09-16
chore(docker): pin the base image by digest

ffroliva:
## Summary

Closes the two `PinnedDependenciesID` container-image alerts Scorecard reports against `docker/Dockerfile:15` and `docker/Dockerfile.test:3`.

`FROM python:3.14-slim` resolves to a **mutable** tag, so the image is a different base on different days and a rebuild is not reproducible. Both Dockerfiles now pin `python:3.14-slim@sha256:cad9a2c8…`. The tag stays in the reference on purpose — it is what tells a reader which interpreter this is, and a bare digest says nothing.

## Two details worth stating

**The digest was verified, not copied.** Scorecard's remediation tip supplies a digest, but `docker manifest inspect -v python:3.14-slim` reports a *different* one — `sha256:810da627…`. That is a single platform's manifest; pinning it would tie the image to one architecture. The correct value is the OCI **index** digest, which `docker buildx imagetools inspect python:3.14-slim` returns and which matches Scorecard's. A `docker pull` of the pinned reference resolves. `docker/README.md` now records that distinction so the next bump does not get it wrong.

**This adds no maintenance chore.** `.github/dependabot.yml` already runs the `docker` ecosystem over `/docker` weekly, and Dependabot bumps digest-pinned images for the same tag. The usual objection to digest pinning — that the base stops receiving updates — does not apply here; the input simply moves from "whatever the tag resolved to today" to "a named thing with an owner", which is exactly what the README's *Who owns each input* table already claims for it.

## Not addressed here, deliberately

The other two open `PinnedDependenciesID` alerts are a different shape and I would not bundle them:

- **`docker/Dockerfile:49` — `pipCommand not pinned by hash`.** Satisfying this needs `pip install --require-hashes -r <file>` with every transitive dependency hashed. That would make `ARG GFLOW_VERSION` decorative — you could no longer build the image for a different gflow-cli version with `--build-arg`, because the hashes belong to one release. The repo already does hash-pinned requirements for the Pages build (`website/requirements.txt`), so the pattern exists; it is the loss of the build-arg that makes it a real trade-off rather than a chore. Worth its own issue and an explicit decision.
- **`docker/Dockerfile.test:13` — `downloadThenRun not pinned by hash`.** `curl … astral.sh/uv/install.sh | sh`. Fixable by fetching a versioned uv release tarball and checking its sha256, which adds a uv-version bump chore that Dependabot cannot see.

## Verification

`docker pull` of the pinned reference resolves; `tests/test_dockerfile_version_pin.py` 4 passed; doc-links, repo hygiene and the website/docs mirror all green. No `src/` change, so no e2e applies and none was written — this touches no Flow surface.

## Summary by CodeRabbit

- **Security**
  - Pinned the Python 3.14 slim Docker base image to a verified multi-platform digest for more reproducible builds.
  - Applied the same digest pinning to production and test images.

- **Documentation**
  - Documented the digest-pinning approach and multi-platform image guidance.
  - Added an unreleased security changelog entry.


ffroliva:
Closing: chasing the Scorecard badge isn't worth the maintenance surface it implies. The remaining two PinnedDependencies alerts would each cost something real — `--require-hashes` makes `ARG GFLOW_VERSION` decorative, and pinning the uv installer adds a bump chore Dependabot cannot see — so pinning only the base image would have left the badge unmoved anyway.

Recording what was measured here so it does not need re-deriving: the digest to pin is the OCI **index** digest, `sha256:cad9a2c871761c413caa6fdd6441c783451e740a48aaeba60ae62a8b53525ef6`, which `docker buildx imagetools inspect python:3.14-slim` returns. NOT the value `docker manifest inspect -v` prints first (`sha256:810da627…`) — that is a single platform's manifest and would pin the image to one architecture.

Issue #478, opened 2026-08-12
profile-lease: cooperative handoff and minimum-hold window (currently fail-fast only)


ffroliva:
linkedin-mcp lets a waiting process ask the current holder to release the browser, with a minimum-hold window to avoid ping-pong.

Current state (gflow): profile_lease.py is a plain fail-fast advisory kernel lock; contention raises ProfileLockedError immediately (profile_lease.py:162-183), never waits. No announce/release-request channel, no min-hold, no idle-close.

Proposed: optional cooperative handoff (a waiter signals; the holder releases at a safe point) with a minimum-hold window and no-release-while-a-call-is-in-flight.

Value: smoother multi-client use; today concurrent use fails fast (safe but abrupt). Size: larger. Validate that the daemon/MCP/CLI actually contend before building. Tier 2.

PR #862, opened 2026-09-17
feat(mcp): wait out profile contention instead of hard-failing

iceblue03:
## Summary

The CLI's fail-fast profile-lease default (`GFLOW_CLI_LEASE_WAIT_SECONDS=0`) is right for a one-shot terminal command a human can just retry. The MCP server has no human in that loop: two tool calls landing close together on the same profile (a client retry, two agents sharing a profile) currently hard-fail with `ProfileLockedError` instead of the second call simply running after the one already in flight finishes.

This adds `_apply_mcp_lease_wait_default()`, called first thing in all three MCP entry points (`run_stdio` / `run_http` / `run_sse`), before anything touches `Settings` (which is `@lru_cache`d). It sets `GFLOW_CLI_LEASE_WAIT_SECONDS=180` via `os.environ.setdefault` — long enough to outlast a typical single image/video generation, and a complete no-op for any operator who has already set the var themselves (including explicitly back to `0` for fail-fast).

The underlying wait mechanism itself (`profile_lease.py`, #478/#503) is untouched — this only changes which default the MCP surface starts from.

## Lifecycle

- [ ] Issue triaged with `issue-assessment` — no issue filed; this is a self-contained reliability gap noticed while exercising the MCP tools directly (concurrent `gflow_*` calls against the same profile).
- [ ] `predict` verdict — not applicable; no transport/auth/selector/schema surface changes, only an env-var default at the MCP process entry points.
- [ ] `scenario` + `plan` — skipped for a change this narrow (one function, three call sites); happy to write these up if a maintainer wants the full artifact trail before merge.
- [x] `check` (scoped) — `ruff check`, `ruff format --check`, and `pyright` all pass clean on the two touched files. I did not run the full repo-wide hygiene/doc-link/coverage suite from a fork checkout.
- [x] MCP twin — not applicable; there is no new CLI option to mirror. The whole point is that CLI and MCP now *intentionally* diverge on this one default (fail-fast vs. wait), which is exactly what `GFLOW_CLI_LEASE_WAIT_SECONDS` already exists to let either surface override.
- [ ] E2E evidence — this changes a timing default around `ProfileLease`, not a Flow DOM/network surface, so nothing under `tests/e2e/` exercises it directly. Coverage is two new focused unit tests in `tests/mcp/test_server.py::TestMcpLeaseWaitDefault` (default applies when unset; an explicit value, including `0`, is never overridden), plus the three existing entry-point tests updated to patch the new call. `56 passed` locally.
- [ ] Live-verified — not a generation-path DOM change, so I did not run the 5-layer live-verify ledger. Informal evidence: while exercising `gflow_*` MCP tools back-to-back against one profile in a live session (character/instructions/credits calls fired close together), none hit `ProfileLockedError`.
- [ ] Council review — not run; happy for a maintainer to run `pr-council-review` and I'll address findings.

## Validation

- [x] Focused tests added for the behavior change (`TestMcpLeaseWaitDefault`, 2 new tests)
- [x] Documentation: none needed — this is an internal default, not a new user-facing option (`GFLOW_CLI_LEASE_WAIT_SECONDS` itself is already documented)
- [ ] `uv run python scripts/ci/check_doc_links.py` — not run (no docs touched)
- [x] `uv run ruff check src tests` (scoped to the 2 changed files)
- [x] `uv run pyright src` (scoped to `src/gflow_cli/mcp/server.py`)
- [x] `uv run pytest -q tests/mcp/test_server.py` → `56 passed`

## Contribution Checklist

- [x] My commits use my real Git identity
- [x] Commit includes `Signed-off-by:` (`git commit -s`)
- [x] No secrets, cookies, account tokens, signed URLs, or private captured data included
- [x] I reviewed the AI-assisted change before submitting

## Summary by CodeRabbit

- **Bug Fixes**
  - MCP server entry points now default the lease wait time to 180 seconds when no value is configured.
  - Explicit lease wait settings, including `0`, are preserved.


ffroliva:
Thank you @iceblue03 — the goal here is right: an MCP caller has nobody to retry for it, so contention should wait rather than fail.

We're taking this over together with #863, tracked in #864, and shipping both in one release. Two findings from verifying it, detailed in #864:

- `gflow`'s root command loads and caches settings before `mcp run` / `serve` reaches the server, so the default is applied too late to take effect on those entry points.
- Two calls inside the same MCP server run in one process, where the lease fails fast by design, so they need to be serialised per profile in the server rather than waited out.

You'll be credited in the changelog, and we'll ping you when it ships.

ffroliva:
Shipped in **v0.78.0** — thank you, and sorry for the wait.

Your diagnosis was right and the goal was right: an MCP caller has nobody to retry a `ProfileLockedError` for it. We took the change over in #868 with two corrections found while verifying it:

1. **The default never reached Settings on the real entry points.** `gflow`'s root command loads and caches settings before `mcp run` / `serve` dispatch, so setting the variable inside the server was too late. Measured through `gflow mcp run`: the variable read `180`, the loaded setting stayed `0.0`. The fix resets the settings cache — and it now also treats a value from a `.env` file as the operator's choice, which `os.environ.setdefault` would have overridden.
2. **Two calls inside one server never waited anyway.** Tool calls run in the server's own process, where lease contention fails fast by design (waiting on yourself would deadlock). Those calls are now serialised per profile inside the server, which is what makes the "client retry / two agents on one server" case work. The 180 s wait covers contention with a *different* process, such as a CLI run.

Credited in the changelog. Closing in favour of #868.

ffroliva:
Superseded by #868, shipped in v0.78.0. Your idea and the credit are in the changelog.

iceblue03:
Superseded by #868 — thanks for taking this further and fixing the settings-cache-timing and same-process-contention issues I missed. Closing.

Issue #861, opened 2026-09-17
feat: a 'gflow docs' command — 126 docs and a 76KB INDEX are unreachable from the CLI at the moment of use


ffroliva:
## Summary

`gflow-cli` ships **126 files in `docs/`** and a `docs/INDEX.md` that routes them — but nothing
in the CLI points at any of it. `gflow --help` lists commands; no command surfaces documentation.
So at the moment of use the knowledge is unreachable, and the reader guesses instead.

Proposal: a `gflow docs` command that prints a compact table of contents with one line per topic
and a link, plus `gflow docs <topic>` to locate or print one.

## Why, from a real failure today

Producing a story on a migrated `flow.google.com` account, I ran:

```
gflow video r2v "<prompt>" --ref still.jpg --model omni-flash --duration 10 ...
```

`--duration 10` came from a recipe written when the account was on labs. gflow refused, exit 11,
and the message was excellent:

> the migrated Flow host offers reference-to-video only at 8s; at 10s it does not refuse but
> silently drops the references and submits a text-to-video run with their file names typed into
> the prompt — a full-price clip with none of the images on it

**That guard is the reason this cost nothing**, and it deserves saying plainly. But the rule
existed in writing beforehand and I never found it. `gflow video r2v --help` documents
`--duration [4|6|8|10]` with "Refused pre-submit … when your account's cohort renders no duration
control", which is true but does not say *this host is 8s-only for r2v*. To learn that I had to
fail first.

`docs/INDEX.md` is **76 KB**. It is a good document and completely impractical as a lookup under
time pressure — an agent will not read it, and a human will not find one line in it mid-run.

## Suggested shape

```
$ gflow docs
gflow-cli documentation (126 topics)

  GETTING STARTED
    usage              Commands, flags, exit codes                 docs/USAGE.md
    authentication     Profiles, login, migrated vs labs hosts     docs/AUTHENTICATION.md
    configuration      Env vars, config.toml, hosts                docs/CONFIGURATION.md

  GENERATION
    reference-strategies  --ref vs --reference-entity vs @Name     docs/REFERENCE_STRATEGIES.md
    character             Character entities                       docs/CHARACTER.md
    ...

  Full index: docs/INDEX.md          Show one: gflow docs <topic>
```

- `gflow docs <topic>` prints the file (or its path, for piping into a pager/editor).
- `gflow docs --search <term>` greps titles and headings — `gflow docs --search duration` should
  land on whatever documents the per-host duration matrix.

[...]

Issue #846, opened 2026-09-16
Clean Windows install is unusable: ConsoleRenderer(colors=True) needs colorama, which is not a declared dependency


dupiaz:
## Summary

On a clean Windows install, **every** gflow-cli command aborts with a `SystemError` before doing any work. `configure_logging` builds `structlog.dev.ConsoleRenderer(colors=True)`, which on Windows requires `colorama` — but `colorama` is not in `Requires-Dist`, and none of the declared dependencies pull it in.

## Reproduction

```powershell
uv init --python 3.13 tmpflow ; cd tmpflow
uv add gflow-cli
uv run gflow auth login --browser chrome
```

Observed on gflow-cli 0.76.0, Python 3.13, Windows 11 (10.0.26200), uv-managed venv.

## Traceback

```
Traceback (most recent call last):
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "...\.venv\Scripts\gflow.exe\__main__.py", line 10, in <module>
    sys.exit(main())
  File "...\.venv\Lib\site-packages\click\core.py", line 1631, in __call__
    return self.main(*args, **kwargs)
  ...
  File "...\.venv\Lib\site-packages\gflow_cli\cli.py", line 158, in main
    configure_logging(settings.log_format)
  File "...\.venv\Lib\site-packages\gflow_cli\observability.py", line 104, in configure_logging
    renderer = structlog.dev.ConsoleRenderer(colors=True)
  File "...\.venv\Lib\site-packages\structlog\dev.py", line 700, in __init__
    self._styles = self.get_default_column_styles(colors, force_colors)
  File "...\.venv\Lib\site-packages\structlog\dev.py", line 815, in get_default_column_styles
    _init_terminal(cls.__name__, force_colors)
  File "...\.venv\Lib\site-packages\structlog\dev.py", line 88, in _init_terminal
    raise SystemError(
SystemError: ConsoleRenderer with `colors=True` on Windows requires the colorama package installed.
```

## Why nothing else supplies it

[...]

PR #869, opened 2026-09-17
feat(docs): `gflow docs` — list, print and search the documentation from the CLI (#861)

ffroliva:
## What

126 pages in `docs/` and nothing in the CLI pointed at any of them. The failure that filed #861 was `video r2v --duration 10` against a host that offers r2v at 8 s only — a rule that **was already written down** and was never found.

```
gflow docs                          every topic, with what it covers
gflow docs usage                    one page, raw Markdown (it pipes)
gflow docs --search "r2v duration"   the line that answers the question
```

Read-only and offline: no network, no account, no credits, no database.

## Three things measured rather than assumed

**1. The pages ship in the wheel.** The user this was filed for installed from PyPI and has no checkout, so a command that could only print GitHub links would not solve it. The two-line version — `force-include = { "docs" = ... }` plus `exclude` — was built and the wheel inspected: hatchling **ignores exclusions for forced inclusions**, and 148 files it was told to exclude shipped anyway (4.3 MB of plans, spikes and PNGs). `hatch_build.py` globs `docs/*.md` instead — 126 pages, +1.3 MB, nothing else, and the glob is also the topic list so neither can drift.

**2. Search returns a line, not a file name.** The rule above lives inside a 4 000-character bullet in `MCP.md`. Answering *"it is in MCP.md"* leaves the reader where they started, so hits carry `docs/FILE.md:LINE` plus the line **windowed** around the match. Terms are ANDed; curated `INDEX.md` shortcut rows rank above raw body text.

**3. Ranking's limit is recorded, not papered over.** `--search duration` alone has 133 hits. Three orderings were measured:

| ordering | first result |
|---|---|
| alphabetical | three lines of `AGENT_UI_RECON.md` |
| + heading bonus | eight `LIVE_VERIFICATION_*` release records |
| + "page is routed from `INDEX.md`" | four release records, the rule 5th |

The third had the best story — reuse the repo's own statement of what is reference material. It is **false**: `INDEX.md` links 93 of the 126 pages. It was removed rather than kept as a knob that does nothing, and `_score`'s docstring says why so it is not reinvented. What works is a second word: `r2v duration` → 5 hits, `docs/MCP.md:87` first. A vague query reports how many hits it held back instead of quietly truncating.

## Safety

A topic resolves from its slug, file name, either case, or an unambiguous prefix — and **never by joining user input onto a path**. The argument is a key looked up in the enumerated page set, so `gflow docs ../../../../etc/passwd` is an ordinary unknown-topic refusal. Exit 11, an existing class: no new exception, no new exit code. Every printed string round-trips through the console's own encoding — the #846 shape from one release ago.

## MCP

No twin yet, and it is **registered in `_MCP_EXEMPT`** so the decision is enforced by `test_cli_parity.py` rather than claimed in a docstring. "Agents do not need it" would be false — the consumer who filed #861 *is* an agent. What is true is that documentation is a **resources** provider, not a tool call. Every decision therefore lives in `docs_catalog`, Click-free and returning data, which is the upgrade path.

## Verification status

- [x] `/gflow:scenario` → `/gflow:plan` artifacts committed under `docs/superpowers/plans/2026-09-17-gflow-docs-command/`, including the two measurements that changed the design before code existed and the one that changed the acceptance test after
- [x] **4 524 offline tests + 40 new**, 5 skipped, green
- [x] **Integration test builds a wheel, installs it into a scratch venv and runs the command with the repository nowhere in sight** — the only test that can tell "it works here" from "it works for the user who filed it"
- [x] `ruff check` + `ruff format --check` clean · `pyright src` 0 errors · repo hygiene · doc links · website mirror regenerated · council memory
- [x] Every Critical + High scenario has a test
- [ ] `/gflow:branch-review` council — running now, findings will be applied or declined in writing on this PR
- [ ] Wheel grows **+0.58 MB compressed** (0.74 -> 1.32 MB; 1.3 MB is the uncompressed page total). Deliberate and measured; worth a maintainer's explicit nod since it is every user's download.

`Refs #861`.


ffroliva:
## Council review — six dimensions, applied

Ran against `bfc0afcc`. **D0 mechanical gate caught one before the LLM dimensions even started**: repo hygiene goes red only once the new files are *tracked*, so the pre-commit run was green and `hatch_build.py` would have reddened CI. Fixed in `f951cf2c`.

### Applied

| # | Dimension | Finding | Fix |
|---|---|---|---|
| 1 | D1/D2 | **`_plain()` stripped single `_`**, so every curated hit printed a file name that does not exist — `REFERENCE_STRATEGIES.md` → `REFERENCESTRATEGIES.md`, in the rows that rank **first**. 106 of 126 pages carry an underscore | Only `**`, `__` and backticks are stripped |
| 2 | D4 | **`Traversable.read_text()` takes no `errors=`** → `TypeError` on the one path that matters (zipimport), uncaught by `except (OSError, ValueError)`. Invisible locally because every test runs the *checkout fallback* | Reads bytes, decodes explicitly |
| 3 | D1/D2 | **Heading bonus fired on `#` shell comments inside fenced blocks.** That is what put eight `LIVE_VERIFICATION_*` rows on top of `--search duration` — the noise the ranking exists to remove | Fence tracking, 4 lines. `--search duration` now opens with curated `INDEX.md` rows |
| 4 | D4 | **The cp1252 test proved nothing.** Mutation: replacing every `_console_safe(` call site with `str(` left **33/33 green** — the test called the helper directly and never ran the command; Rich's box glyphs were never covered | Subprocess run under `PYTHONIOENCODING=cp1252`, three invocations. `_console_safe` deleted for one `sys.stdout.reconfigure(errors="replace")`, which covers Rich too |
| 5 | D14 | **The curated-shortcut parser was a strict subset of body search** — measured 0 curated hits of 7 for the flagship `r2v duration`. 28 lines of regex that only *re-ordered* rows search already had | Collapsed to a score bonus. Deletes finding #1 as a side effect |
| 6 | D11/D13 | **The hook failed silent.** A sdist with `docs/` removed built a valid **767 KB wheel carrying zero pages, exit 0** — and `release.yml` is `uv build` → publish with nothing in between | The hook raises when it finds no pages |
| 7 | D1/D2 | The hook ran for **editable** installs, copying 126 pages into a `site-packages/gflow_cli/` with no `__init__.py`, never read, stale on the next doc edit | Skips `version == "editable"` |
| 8 | D1/D2 | **`--search ""` defeated the mutual-exclusion guard** (guard tested truthiness, dispatch tested `is not None`), silently discarding the topic | Both test `is not None` |
| 9 | D8 | **Every refusal inherited** `ConfigurationError`'s default hint: *"Run `gflow config list-transports`"* | Per-raise `remediation_hint` |
| 10 | D8 | `USAGE.md` promised "an unknown name suggests the nearest ones". **It did not** — a substring scan offers nothing for a typo | `difflib.get_close_matches`; `usge` → `usage` |
| 11 | D8 | The topic table was **424 lines** at 80 cols, summaries wrapped into fragments | One row per topic, ellipsised → **132 lines**. Full text stays in `--json` |
| 12 | D8 | Search title reported the **shown** count, footer the total — "20 match(es)" above "and 116 more" | Title carries the total: `'duration' — 136 match(es), showing 20` |
| 13 | D9 | **`gflow docs` was missing from every place the command surface is enumerated**: `AGENTS.md`, `llms.txt`, `skills/gflow-cli/SKILL.md`, and the synopsis block of the very `USAGE.md` section this PR added | All four |
| 14 | D9 | Two numbers in that section were wrong (133 lines, five hits) | Corrected; the prose no longer quotes brittle counts |
| 15 | D11/D13 | **The +1.3 MB claim was wrong** — that is uncompressed. Measured: 755 987 → 1 365 534 B | CHANGELOG says **+0.58 MB compressed** |
| 16 | D1/D2 | `_snippet` windowed on the first term typed | Centres on the least frequent term — the selective half of a two-word query |
| 17 | D4 | My own docstring claimed the wheel test is opt-in. **False** — `integration` is not in `addopts`' exclusions, so it runs on every plain `pytest` and on PR CI | Docstring corrected; the skip without `uv` now says the guarantee was *not* verified |
| 18 | D4 | Acceptance assertion pinned MCP.md to position 1 — an ordinary docs edit could red the build | Top three |

Plus the negatives D4 asked for: hostile names driven **through the CLI** (not just `resolve()`), the NUL case, `docs ""`, `--search ""`, the remediation text, the typo, and code-fence scoring.

### Declined, with reasons

- **`A_B.md` / `A-B.md` would collide on one slug** (D1/D2 #7). Measured: **zero collisions across all 126 pages**, and D11/D13 independently confirmed it over the full slug transform. A guard for a condition that does not exist is a guard nobody can test. If a colliding pair is ever added, `test_every_shipped_page_is_a_topic_with_no_manifest_to_maintain` fails, because the set comparison loses a member.
- **`truncate_hits` / `Match.topic` / long docstrings** (D14 `shrink:`). `truncate_hits` is gone (inlined); `Match.topic` is now a derived property rather than a stored duplicate. The docstrings stay: they record *measurements that failed*, which is the only thing stopping the next person rebuilding the `INDEX.md`-routing signal that does not work.

### Not a finding

**D3 (security) found nothing exploitable.** It executed 19 hostile topic names — `../../../../etc/passwd`, UNC paths, `%2e%2e%2f`, NUL, `~/.ssh/id_rsa` — against the live module: all refused as unknown topics, **zero reads, zero exceptions**. It confirmed `Topic` has one construction site, fed from `iterdir()`, and that no user string reaches `joinpath`/`open`/`Path()`. It also measured the regexes as linear (no ReDoS) and scanned all 126 shipped pages for credentials: zero hits.

One consistency note it raised, **not fixed here and worth its own issue**: `scripts/ci/check_website_docs_pii.py` gates the 26 published website docs for private identifiers, but the wheel now publishes 126 pages through a channel that gate does not cover. The content is already public in this repo, so nothing is newly exposed — but there are now two publishing channels and only one is gated.

### Still open on this PR

- [ ] **Wheel grows +0.58 MB** for every user. Deliberate, measured, and a maintainer's call rather than mine.
- [ ] SonarCloud was red on the previous head for **its own reasons** — `HTTP 500 on sonarcloud.io/batch/project.protobuf`, *"An unexpected error occurred. Please try again later."* Re-running on this push.

PR #872, opened 2026-09-17
feat(data): `gflow data download <media_id>` — recover a billed asset whose download failed (#865)

### LINKED ISSUE #865: feat: 'gflow data download <media_id>' — a billed asset stranded by exit 7 can only be recovered by paying for it again
## Summary

When a generation succeeds but its download does not, the paid asset is stranded: it exists in
the Flow project, the catalog has a row for it, and **no CLI command can fetch it**. The capability
is already in the codebase — `client.download_video(media_id, out_path)`
(`api/client.py:2042`) and the migrated composer's own `download()`
(`api/transports/migrated_composer.py:2519`) — it is just not reachable except inline at
generation time.

Ask: expose it as `gflow data download <media_id> [--out PATH]`.

## How it bit

Two `video r2v` runs on a migrated account finished and billed, then failed at the last step:

```
WireFormatError exit 7 — migrated host: the generation finished but no signed media URL
was observed within the 20s grace
```

(`RESULT_URL_GRACE_S = 20.0`, `migrated_composer.py:307`.)

Catalog afterwards:

```
23620c34-f2ec-419f-8819-3e60ff5b5767  local_path: null  copy_count: 0
9ad33c78-5762-4cbc-bcbe-07a4c3b061c7  local_path: null  copy_count: 0
```

**Both clips rendered.** They are visible in the project, and Flow's own tile menu offers
Download → 270p GIF / 720p original / 1080p upscaled. So the money bought the asset; only the
retrieval failed, and retrieval is the one part with no second attempt.

Everything else I tried is a dead end on this host:

| attempt | result |
|---|---|
| `gflow data media <id>` | prints the row; no download verb |
| `gflow data sync --names --project <id>` | **404** `projectInitialData` — "Flow RPCs have been deprecated" |
| re-running the generation | bills again, and failed the same way the second time |

That last row is the sharp edge: with no way to fetch a stranded asset, **the only recovery the
CLI offers is to pay for it again** — and when the cause is a timing window rather than the
prompt, the retry strands another copy. Two credits, two orphans, same clip.

## Proposal

```
gflow data download <media_id> [--out PATH] [--profile NAME]
```

- resolve the media from the catalog row that already exists
- reuse `download_video` / the migrated composer's `download()`
- write the file and update `local_files` so `copy_count` stops reading 0 for an asset that is
  now on disk

Two adjacent things worth considering, though the verb alone solves the recovery case:

1. **Say the asset is safe in the exit-7 message.** Right now it reads as a failure; the user
   cannot tell that the clip exists and is one command away. Something like *"the generation
   succeeded and was billed; the media is in the project — retrieve it with `gflow data download
   <media_id>`"* turns a loss into an errand.
2. **`RESULT_URL_GRACE_S = 20.0` may simply be short** for the migrated host — it fired twice in
   four attempts here. Worth measuring before tuning, but a retrievable asset makes the grace far
   less critical either way.

## Related

- **#29** — first-class download for video generation (closed). That shipped the download for the
  generation path; this asks for the same code reachable for an existing media id.
- **#639** — migrated feature matrix. `data sync` is another asset-management command still on the
  dead labs route.
- **#860** — the picker name-match bug; same session, and the reason `i2v` could not be used at
  all, which is why r2v was carrying these runs.

### LINKED ISSUE #871: No CLI path to a billed clip whose signed URL missed the 20s grace — Flow's own UI can download it
## A billed clip that misses the 20s URL grace is unreachable from the CLI, forever

`migrated_composer.download()` can only save a clip whose **signed CDN URL was observed
during generation**. When it is not, the run raises:

```
WireFormatError: migrated host: the generation finished but no signed media URL was
observed within the 20s grace
```

The generation still happened and **the credit is still spent**. The clip is in the
project. There is no CLI path to it: `gflow data` is `list / media / prune / sync`, and
`sync` only reconciles names. `gflow data list videos` shows the row with
`local_path: null, copy_count: 0` — it knows the asset exists and cannot fetch it.

The code says as much, and it is currently true:

> The labs `media.getMediaUrlRedirect` route answers 404 for a migrated media id
> (measured 2026-09-05), so there is no second source: a record with no URL is a
> wire-format failure.

**There is a second source.** Flow's own UI downloads it.

### Measured, 2026-09-17

Two s025 clips were billed and orphaned this way (`23620c34-…` at 13:52:59,
`9ad33c78-…` at 13:56:27, both exit 7). One was recovered **by hand**, at no extra cost,
by driving flow.google.com in a browser:

1. open `https://flow.google.com/project/<project_id>`
2. left nav → **Videos**
3. hover the tile → **More options** (`more_vert`) → **Download**
4. a submenu offers: `270p Animated GIF` · **`720p Original size`** · `1080p Upscaled` ·
   `4K Upscaled (Upgrade)`
5. `720p Original size` produced
   `Artist_opening_door_on_street_20260917210326.mp4`, 2 702 168 B,
   `h264 720x1280 + aac, 8.000s` — byte-identical in shape to what a successful
   in-run download writes.

Every one of those steps is an ordinary Playwright interaction on a page this driver
already owns. The download itself is `page.expect_download()`.

### Proposed

`gflow data download <MEDIA_ID> [--out-dir DIR] [--profile NAME]`

- Resolve the media id in the local catalog (it is already there — that is how the
  orphan is discovered) to get its `project_id`.
- Drive the project page, bind the tile, open the menu, choose **Original size**, and
  save via `expect_download()`.
- Reuse the existing safety posture in `_fetch_mp4`: the `_is_allowed_download_host`
  allowlist, `max_redirects=0`, and the `ftyp` magic-byte check. None of that should be
  re-implemented; a UI-sourced file must clear the same bar as a URL-sourced one.
- Write the `local_files` row so `gflow data list` stops reporting `copy_count: 0`.

### The hard part, stated up front

**Flow's DOM does not carry the media id.** Checked on the live page: searching the full
`outerHTML` (448 436 chars) for either orphaned UUID returned nothing, and no tile
`img`/`video`/background-image URL contains a UUID either. The manual recovery
identified the right tile by **prompt text and creation order** — the catalog's prompt
matched the tile's Flow-generated title (`"…draws the door gently back…"` →
*"Artist opening door on street"*), and the ordering was newest-first.

That heuristic is not good enough to ship: two takes of the same prompt produce two
tiles with near-identical titles, which is exactly the orphan case.

So this needs a **spike before an implementation**: find the wire response that lists a
project's media with ids in the same order the grid renders, and bind tile → id from
that rather than from display text. If no such correlation exists, the honest fallback is
`--index N` plus a printed table for the human to choose from — still infinitely better
than "unreachable".

### Why it is worth it

A missed 20s grace currently converts a paid generation into a permanent loss with no
recovery verb. The failure is **transient and timing-based**, so it will recur. The
recovery is already possible by hand in about ten seconds; encoding it is a small driver
addition on a surface gflow already automates.

ffroliva:
## Summary

A generation that finishes but whose signed media URL is never observed exits 7 **after the credit is spent**. The clip is in the Flow project, the catalog row reads `local_path: null`, and no command could fetch it — the only recovery the CLI offered was to generate it again and pay twice. Because the failure is a timing window rather than a bad prompt, the retry can strand another copy.

`gflow data download <media_id> [--out DIR] [--profile NAME] [--json]` fetches it for **zero credits**, and mirrors as the `gflow_download_media` MCP tool.

Closes #865
Closes #871

## Measured before implementing

Full spike: [`docs/superpowers/spikes/2026-09-17-stranded-clip-recovery.md`](docs/superpowers/spikes/2026-09-17-stranded-clip-recovery.md) — $0, run against **both orphans named in #865**.

- `flow.google.com/project/<project_id>/edit/<media_id>` is a real per-clip route, and **both ids are already in the catalog**, so nothing has to be found in the DOM. This matters: the grid is virtualized — 13 rendered tiles against 183 records — so a tile-driven implementation reaches only the newest page and **cannot recover an older clip at all**.
- Opening that route makes the app fetch the clip's status on `as29s`, whose reply carries a freshly signed `flow-content.google` URL. Both the rpc and the decoder (`generation_record`) already existed for the in-run download path.
- Bytes are verified against the size Flow records, **not just `ftyp`**. The same asset is also served as 360p/720p re-encodes carrying valid MP4 magic, so a magic-byte check alone would silently save the wrong file (the #281 class).

### What this retires from the two issues

| claim | outcome |
|---|---|
| #865 "reuse `client.download_video`" | labs `media.getMediaUrlRedirect` 404s for a migrated media id |
| #865 "resolve the media from the catalog row" | gives `project_id` only — signed URLs are stripped by `redact_metadata` because they expire, so one must be minted per run |
| #871 "DOM lacks the media id → `--index` fallback" | moot: the media id is in the clip's own URL |
| #871 "drive the tile context menu" | unnecessary, and unable to reach older clips |

## Design notes

- Recovery lives in a **new module** rather than `migrated_composer.py`, which #824, #787 and #781 are all currently editing.
- **No DOM selector is on the path**, so nothing here can drift with Flow's markup or the account locale.
- `data media` and `data download` now share one catalog resolver instead of two copies of the same lookup.
- Parity is a **mapping, not an exemption**: an agent whose generation died at the download is stranded exactly as a CLI user is.

## Verification

| gate | result |
|---|---|
| repo hygiene / doc links / website PII / mirror / council memory | ✅ (mirror was stale — regenerated) |
| `ruff check` + `ruff format` (src, tests) | ✅ |
| `pyright src` | ✅ 0 errors |
| unit tests (`test_migrated_recover.py`) | ✅ 12 passed |
| affected suites (mcp, cli_data, transports) | ✅ 259 passed |
| **e2e, live profile** (`tests/e2e/test_data_download_e2e.py -m e2e_data`) | ✅ **2 passed**, repeated 4× |

The e2e run recovered `9ad33c78-…` — one of the two clips #865 reports as unrecoverable — at **2 702 168 B, byte-exact** to the size Flow reports, and asserted the `local_files` row so `data list` stops showing `copy_count: 0`. It also restores catalog state afterwards; an earlier version left a row pointing at a deleted temp file, which was caught and fixed.

## Known gaps (not fixed here)

- **Images are unproven.** Only video clips were exercised; an image record is a different shape.
- **`flow_workflow_id` is `NULL` in the catalog** for both orphans even though `on_started` carries it. If it were persisted the project visit could be skipped — worth its own issue.
- **Exit 7's message still reads as total loss.** #865 asks for it to name this command; deliberately left out of this PR to keep the diff to the recovery path.

## Summary by CodeRabbit

- **New Features**
  - Added `gflow data download` to recover previously generated assets by media ID without spending additional credits.
  - Supports custom output directories, profile selection, and JSON-formatted results.
  - Added the equivalent `gflow_download_media` MCP tool.
  - Verifies downloaded files against the expected size and records recovered files for future reference.
  - Reports clear errors for missing, ambiguous, or unavailable assets.

- **Documentation**
  - Added CLI and MCP usage documentation, including recovery behavior and error handling.


ffroliva:
## Gate green, and re-verified live after the refactor

All 16 checks pass, SonarCloud included.

`03c81f42` fixed both Sonar conditions at the source — no exclusion widened, no assertion weakened:

| condition | fix |
|---|---|
| S3776 cognitive complexity 19 > 15 | `recover_clip` split into `_await_signed_record` / `_collect` / `_fetch_verified` |
| `new_coverage` 74.4% < 80% | both new modules now **100%** (`migrated_recover.py` 77/77, `media_recovery.py` 56/56) |

The added tests cover behaviour that was genuinely untested, not lines: cross-profile ambiguity naming both candidates, the scoped miss naming its profile, the recorded checksum being the real digest (so `verified_local_path` still works), `upsert` idempotency, and the three listener branches a live project load actually hits — non-`batchexecute` traffic, an unreadable body, and frames that are not generation records. The unroutable-row test asserts **no browser is launched** by raising if one is.

### Live re-verification

The refactor changed `recover_clip`, so offline-green was not sufficient. Re-ran e2e against the live profile on the final code:

```
tests/e2e/test_data_download_e2e.py -m e2e_data  →  2 passed in 18.39s
```

That is five live runs total across the branch; each recovers a real billed clip and asserts `ftyp` + exact `size_bytes` + the `local_files` row.

Ready for review.

Issue #593, opened 2026-08-27
Changelog announcement modal blocks every UI action; make dismissal a precondition, not a per-boundary call


ffroliva:
Flow shows a full-page announcement dialog over the editor when Google ships a feature (observed 2026-08-27 on `denon82`: *"a new 360p option for Gemini Omni Flash"*). While it is up, **every** UI action times out.

## Measured, not assumed

`scripts/dev/spike_changelog_modal.py denon82 --project <id>` (read-only, zero credits):

```
body { pointer-events: none }        <- whole app behind is inert at CSS level
aria-hidden: null   inert: false     <- accessibility tree says the app is fine
mode-switch button: visible=True enabled=True  -> COVERED by <div>
```

That combination is the failure signature: Playwright's actionability check sees a **visible and enabled** element, waits for it to receive the click, and it never does. Nothing but hit-testing reveals it — a `TimeoutError` with no useful message. A real `gflow image t2i` run failed exactly this way, three navigation attempts deep, before the modal was dismissed by hand.

## The existing machinery is correct and already generic

```
overlay_detected   selector="iframe[src*='/flow/changelogs/']"
overlay_dismissed  method=close_button_page
                   selector="[role='dialog']:has(a[href*='changelog']) button"
BEFORE: bodyPE=none  dialog=true   modeBtnHit=COVERED
AFTER : bodyPE=auto  dialog=false  modeBtnHit=REACHABLE
```

Both anchors are **content-independent**, which is what matters: the copy changes with every announcement, the structure does not. The dialog carries exactly one button — here labelled `Comece já` — and no X. Nothing text-matches, so it works in any locale.

**Dismissal persists server-side.** After one click the next run logged no `overlay_detected` at all, and the generation succeeded (`49c79a0e-…_1.jpg`, NARWHAL). So the cost is one click per announcement, then free.

## The actual gap

`_dismiss_blocking_overlays` is called at **9 specific boundaries** (`ui_automation.py` ×7, `ui_automation_video.py` ×2) rather than as a precondition of every action. Anything that acts without passing one of those boundaries is unprotected, and the failure mode is an unexplained timeout rather than a clear error.

## Proposal

Hoist it to a single precondition gate at each public transport entry point (`generate_images`, `generate_images_batch`, `generate_character_images`, and the video equivalents), replacing the scattered calls.

## Constraint that must not be lost

[...]

PR #873, opened 2026-09-18
fix(migrated): read the refusal card on submit timeout and failed records

stgmt:
## Problem

On the migrated host (flow.google.com) a content-policy refusal has two disguises:

1. **In-band status-4** — the generation record lands in the grid as failed with no reason.
2. **Submit timeout** — the refusal card appears in the DOM but the submit reply never parses, so the composer raises `TransportTimeoutError` (exit 9, retryable).

Both lose the actual refusal text, which lives only in the DOM card ("We noticed some unusual activity… You have not been charged for this generation"). Downstream callers cannot distinguish a prompt refusal from a transport stall or an account-level abuse flag.

## Fix

- `REFUSAL_CARD_MARKERS` + `_REFUSAL_CARD_JS`: scrape the smallest-element innerText of refusal cards (≤600 chars), 3s budget, best-effort — never masks the real error.
- Baseline captured before the submit click; `_raise_if_refused` raises `ContentPolicyError(route="batchexecute:submit")` only for cards that appeared *after* this submit (the grid keeps old failures forever).
- Wired at three points: submit-reply timeout, `_await_terminal` `TransportTimeoutError`, and `final.is_failed` records.

## Tests

4 new tests in `test_migrated_composer.py`: timeout→ContentPolicyError, status-4+card→ContentPolicyError, stale card→still timeout, status-4 without card→still failed record. 126/126 file green, 1091 transport tests green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migrated Flow video and image generations now detect newly rendered content-policy refusal messages and report them as content-policy errors.
  * Existing refusal messages are ignored when they were present before submission.
  * Refusals are detected after submit failures, submission or polling timeouts, and failed generation responses.
  * Missing generation records without a refusal message continue to report the appropriate format error.
  * Content-policy errors now trigger diagnostic incident bundles and screenshots for easier troubleshooting.


ffroliva:
## Council review — 🔴 RED

Thanks for this. The **diagnosis is right and valuable**: on `flow.google.com` a refusal really does arrive as a media-grid DOM card while the batchexecute record is a bare status-4 or never parses, and losing the reason to a retryable `TransportTimeoutError` is a real bug. The negative controls you wrote (stale-card → still timeout, status-4 without a card → still a failed record, empty reply without a card → still `WireFormatError`) are exactly the right shape, and the `Counter` diff rather than a set diff is the correct answer to two identical consecutive refusals.

The mechanism is where this blocks. Eight review dimensions ran against `d920f8f2`; four of them landed on the same finding independently.

### Must-fix

**1. The whole-body text scan re-derives a bug this repo already fixed once, live.**

`_REFUSAL_CARD_JS` scans `document.querySelectorAll('body *')`. That mechanism was tried, shipped, and rolled back — `src/gflow_cli/api/transports/drivers/agentic.py:210` already carries the replacement:

```python
_POLICY_REGION_SELECTOR = '[role="alert"], [aria-live="assertive"], [role="dialog"]'
```

under a comment recording why: *"Scanning the whole page body false-positives on STATIC CHROME — e.g. a 'Content policy' footer/menu link present on every load (observed live 2026-06-14: a benign prompt raised a spurious block within ~6 s)."* Same lesson written up in `docs/AGENT_UI_E2E.md:88`, with its conclusion spelled out: *"a miss → timeout beats a false positive that breaks every generation."*

The pre-submit baseline narrows this but does not close it — it only neutralises text **already mounted**. Anything that mounts after the click (a toast, a lazily-opened menu, a tile the CDK viewport recycles in) reads as new.

Please scope the scan to the alert/dialog/live regions, or to the failed tile element.

**2. `"not been charged"` is our own credit-free-verification signature, not a refusal signature.**

`docs/LIVE_VERIFICATION_v0.78.0.md:85` records Flow showing *"Failed — You have not been charged for this generation"* **for a deliberately aborted submit**. That `route.abort()` technique is how this project verifies generation paths without spending credits. With this marker in the list, every credit-free verification run reports as a content-policy refusal.

**3. The baseline is unsound on a virtualized grid.**

`_wait_for_media_grid_stable` breaks as soon as **≥1** tile exists, not when the count is stable. The media grid is an Angular CDK virtual scroller — `docs/superpowers/spikes/2026-09-17-stranded-clip-recovery.md` on `develop` measured **13 rendered `flow-video-tile` against 183 records**. (Your branch is 3 commits behind and does not have that spike, which is probably why it was not in front of you.)

So an old failure card outside the mounted window at baseline time, mounted later when the new tile shifts the viewport, is counted as NEW → a false **terminal** `ContentPolicyError` on a run that is queued and rendering fine. `drivers/agentic.py:780` faced this and chose the opposite trade, unioning two scrapes: *"we prefer a false miss (timeout) over a false positive."*

**4. The scraped card text reaches users unredacted.**

```python
detail=f"migrated host refused the generation: {new[0]}"
```

`to_problem_details()` copies `detail` verbatim into `--json`, the structlog `error_raised` event and the console. This file already imports the scrubber at line 82 and applies it to its other DOM-derived details at `:1116` and `:1163`:

```python
detail=redact_sensitive_text(f"migrated host refused the generation: {new[0]}")
```

Given the markers are loose enough to match non-refusal elements, up to 400 chars of arbitrary page text is going into an error message users are told to file bugs with.

**5. E2E evidence.**

This PR touches a Flow surface (the credit-spending submit path), so under CONTRIBUTING it owes an e2e test that was run, or a named blocker. Right now there is neither — and separately, `_REFUSAL_CARD_JS` is executed by **nothing**: `FakePage.evaluate` returns a canned list for any JS, so the marker list, the `length > 600` filter, the smallest-element/`childHas` selection and the 400-char slice are all unverified. The 6 new tests prove the Python attribution logic, which is genuinely useful, but not the scraper.

**If you do not have a Flow account or credits to run this, just say so in a comment** — that is the documented exemption and it drops this item from blocking to "maintainer must run", which I will then do.

**6. Exit-code contract changes with no docs.**

`TransportTimeoutError` (9) / `WireFormatError` (7) → `ContentPolicyError` (5), and `retryable` flips `true` → `false` in the `--json`, MCP and worker envelopes (`ContentPolicyError` is not in `RETRYABLE_ERRORS`). The diff is 4 files, none of them docs. Stale afterwards:

- `docs/DEBUGGING.md:126` + `website/docs/DEBUGGING.md:126` — *"Never captured: expected `ContentPolicyError`"*. Both mirrors are identically false, so `generate_website_docs.py --check` stays green on a false claim.
- `docs/DEBUGGING.md:105` — screenshots described as "UI-state failures only".
- `ContentPolicyError`'s docstring in `errors.py:312` — *"Two known raise sites"*; this is the third.
- `CHANGELOG.md` `[Unreleased]` is empty.

**7. The image path misses the most likely refusal shape.**

`on_response` wraps decoding in `except Exception: result.set_exception(exc)`, but only `WireFormatError` and the timeout branch check for a refusal. A decode `TypeError`/`IndexError` on a refusal-shaped payload reaches `return result.result()` with no check. No test drives `submit_images_and_observe` with a card at all, so commit `120182c7`'s "and image submits" is unexercised.

**8. The `diagnostics.py` change is out of scope, and its new rationale is wrong.**

Re-enabling capture for `ContentPolicyError` is arguable on the merits (the DOM *is* now the reason carrier), but the refusal feature raises the error either way, so it does not need to ride along — and the PR body never mentions it. The replacement comment says *"Auth surface errors fire on accounts.google.com"*, which is false for `AuthExpiredError`: `api/credits.py:53` and `transports/_common.py:470` both raise it from an HTTP 401 with no browser involved. The deleted sentence carried the real reason. `WireFormatError` screenshots are unrelated to refusals and change behaviour at every `WireFormatError` site repo-wide — please split that out.

**9. English-only markers.**

`AGENTS.md` § Locale-Invariance Discipline forbids text-label matching for DOM elements, overlays and announcements anywhere under `src/gflow_cli/api/transports/`. This fails open — on a pt-BR/DE/JA profile every marker misses and the user silently gets the old exit 9 — so it is a degrade rather than a break, but it is still the rule. Two smaller notes in the same list: `"can't generate"` uses an ASCII apostrophe (0x27) where Google renders the typographic one (U+2019), and `"violates"` drops the stem technique the existing list uses (`"violat"`).

**10. Three seconds now sit between the enable-check and the credit-spending click.**

`_capture_refusal_baseline` runs before **every** submit, video and image. On a fresh empty project no tile ever mounts, so it burns the full 3 s every time. If the submit control disables inside that window, `_why_the_click_missed` reports it as **exit 23 selector drift** — i.e. tells the user to file a frontend bug for a delay we introduced.

### Nice-to-have

- `_wait_for_media_grid_stable`'s `except Exception: pass` (changed from `break` in `d920f8f2`) spins `evaluate` + `sleep(0.1)` for the full 3 s on a dead page; a `page.is_closed()` guard would stop it. Its `page.evaluate` also lacks the `asyncio.wait_for` bound its sibling `_refusal_card_texts` has.
- `asyncio.wait_for` around `page.evaluate` bounds the Python await but does not cancel the JavaScript running in the browser — the "a hung evaluate cannot extend an already-failed run" comment is half true.
- Six of the eight markers are unevidenced guesses; the PR cites one observed card. Shrinking to what has been captured cuts the false-positive surface.
- `timeout_s` on `_wait_for_media_grid_stable` is never passed a non-default value.
- Zero `log` calls across ~90 new lines in a file with 43 of them — a marker-miss event would make the non-English no-op observable instead of silent.
- Branch prefix: this repo uses `bugfix/`, not `fix/`.

### Suggested shape

Detection structural, detail textual: anchor on the failed tile's structure (an ARIA role, a status attribute, an icon ligature), then read `inner_text()` of the **matched node** for the reason. That keeps the reason — which is the whole point of the PR and worth keeping — without betting on English copy or on a whole-page scan. `CREDITS_WARNING` at `migrated_composer.py:152` is the house precedent for this exact problem shape: one structural locator, a 10-line guard, and a spike doc behind it.

If the tile's structure is not documented anywhere yet, that is a `/gflow:spike` rather than a guess — and worth doing, since nothing in `docs/superpowers/spikes/` currently describes the failed-card DOM.

### Verified locally at `d920f8f2`

`ruff check` · `ruff format --check` · `check_doc_links.py` · `check_repo_hygiene.py` · `pyright src` — all clean. `pytest tests/api/transports/test_migrated_composer.py tests/test_diagnostics_recorder.py` → **154 passed**. Note that SonarCloud is *skipped* on fork PRs and first runs on the merge commit, so the green check count above does not include a quality gate.

stgmt:
Thank you for the thorough and constructive Council Review. All critical points have been addressed in commit `437134ea`:

### 1. Structural detection instead of whole-body text scan (§1, §2, §9)
- Replaced the `body *` scan and English text markers with structural probing (`_REFUSAL_CARD_JS`):
  1. `[role="alert"], [aria-live="assertive"], [role="dialog"]`.
  2. Media tiles (`flow-grid-tile-container`, `flow-tile-container`, `flow-video-tile`, `flow-image-tile`) carrying a `warning` or `error` Material Symbol ligature.
- Dropped all English text-matching and the `"not been charged"` signature entirely. This is fully locale-invariant and cannot false-positive on static chrome or deliberate abort tests.

### 2. Redaction (§4)
- Wrapped scraped reason detail in `redact_sensitive_text(f"migrated host refused the generation: {new[0]}")` before raising. Added a regression test verifying email addresses and auth tokens are scrubbed.

### 3. Removed the 3-second delay (§3, §10)
- Dropped `_wait_for_media_grid_stable` and the 3s delay completely. Baseline scrape now runs immediately before click (<5ms), avoiding any submit control disable / drift races on empty projects.

### 4. Image submit path (§7)
- In `submit_images_and_observe`, wrapped decode exceptions to ensure `_raise_if_refused` is checked on *any* failure (not only `WireFormatError`). Added unit tests for image submit timeout with refusal card and image decode error with refusal card.

### 5. Scope hygiene (§6, §8)
- Reverted out-of-scope changes to `diagnostics.py` and `test_diagnostics_recorder.py`.
- Documented the third raise site in `ContentPolicyError` docstring in `errors.py`.
- Added changelog entry under `[Unreleased]` in `CHANGELOG.md`.

### 6. Gates & Verification
- `ruff check` · `ruff format --check` · `pyright` (0/0) · `check_repo_hygiene.py` · `check_doc_links.py` all clean.
- All unit and regression tests pass (15/15 targeted refusal/image tests pass in 0.68s).

ffroliva:
## Council re-review of `437134ea`: 🔴 RED

Thank you for the rework, and apologies that this re-review took four days. The ball was with us from 2026-09-19. Several prior items are genuinely fixed: the whole-body scan is gone, redaction is applied, the `diagnostics.py` scope creep is reverted, the 3 s wait is gone, the image path is covered on both timeout and decode, and the `Counter` diff is right.

One finding decides the verdict, and we measured it rather than inferring it.

### 1. An aborted submit is now reported as a content-policy refusal (measured live)

`tests/e2e/test_project_create_e2e.py::test_video_without_a_project_reaches_the_submit` aborts the video submit with `route.abort()`. That is how this repo verifies generation paths for free. Run against this head on a `flow.google.com` account ($0), it fails:

```
ContentPolicyError: Content policy rejection: migrated host refused the generation: warning
Falha / Não foi possível gerar o vídeo. / Não houve cobrança por esta geração. / refresh / undo / delete_forever
```

The aborted tile has the **same structure** as the refusal card: a `warning` `mat-icon`, a "not charged" message, and refresh/undo/delete actions. Dropping the text markers removed the English-phrase collision, but the structural probe collides in the same place. Structure alone cannot tell a refusal from an abort.

What does differ is that the aborted submit **never got a server reply**. Gating the DOM check on "the submit reached the server" (a parsed reply or operation id exists) may be the discriminator. We don't know yet, because nobody has captured a real refusal tile next to an abort tile. **We're happy to run that spike ourselves**, since we hold migrated accounts. Tell us if you'd rather do it.

### Other must-fix
2. **`_REFUSAL_CARD_JS` never runs in a test.** `FakePage.evaluate` returns a canned value, so a wrong tag, icon or filter passes all 10 tests. Please add one real-Chromium `page.set_content(...)` test, following the pattern in `tests/api/transports/test_ligature_carrier.py`, fed a captured refusal tile and an abort tile.
3. **Selectors with no evidence behind them.** `flow-grid-tile-container` and `flow-tile-container` appear nowhere except this code, so please remove them or cite a capture. `tile.querySelector('mat-icon')` reads only the first icon; use `querySelectorAll(...).some(...)`.
4. **`[role="dialog"]` / `[aria-live="assertive"]` mounting after the click** (changelog modal #593, a quota toast) becomes a non-retryable exit 5. On the image path that also covers an HTTP 429. Please drop the dialog arm, or require the ligature on it too.
5. **Virtualized-grid baseline (prior #3) is still open.** An old failed tile that scrolls into the window after the click reads as new. With the stability wait gone, the baseline is taken earlier than before, which makes this more likely.
6. **The detail carries the whole tile `innerText`**: icon ligature names ("warning", "refresh", …, see above) and possibly the prompt caption, which then lands in the history DB even with `history_prompts="redacted"`. Please read only the tile's message node.
7. **Remediation hint.** The site inherits `ContentPolicyError`'s default ("reduce prompt text / ≤1 person"). The observed cards say "unusual activity" or "couldn't generate", so please pass a hint that fits. The comment at `migrated_composer.py:316` ("Phrases … measured to carry") is also stale.
8. **E2E.** The thread still has no e2e run and no named blocker. Item 1 is now that evidence, and it came back red.

### Mechanical
CI never ran on this head, because GitHub holds fork workflows. We gated it locally on `437134ea`: ruff, format, doc links, hygiene and pyright are all green, and `tests/api/transports` + errors give **1179 passed, 1 skipped**. It also merges cleanly onto current `develop`. SonarCloud has not run.

The diagnosis in this PR is still right and still wanted. It's the mechanism that needs the spike before it can land.

PR #881, opened 2026-09-20
fix(auth): recover from /about landing via Create button and parse batchexecute RPC error envelopes

stgmt:
## Summary

Addresses two related failure modes where unauthenticated or failed migrated sessions mislead callers with false diagnoses:

1. **Recover from `flow.google.com/about` landing via "Create with Google Flow"**:
   When an account session cookie for `flow.google.com` is missing or expired, Google Flow redirects client-side to `flow.google.com/about`. On that page, a prominent `"Create with Google Flow"` button leads directly to `accounts.google.com/v3/signin/accountchooser`.
   Previously, `ensure_editor` stalled for 30 seconds waiting for `READY_ANCHOR` (which never exists on `/about`), and `raise_if_known_landing` raised a fatal `FlowAppError` claiming the account had no access to the project.
   `FlowApiClient._handle_account_chooser` and `raise_if_known_landing` now detect `/about`, click `"Create with Google Flow"`, and seamlessly trigger the PR #764 account chooser auto-selection using `.gflow_account`.

2. **Parse Google `batchexecute` RPC error envelopes into typed exceptions**:
   When Google Flow RPCs fail (e.g. status 13 `INTERNAL`, status 16 `UNAUTHENTICATED`, status 8 `RESOURCE_EXHAUSTED`), `batchexecute` returns `["wrb.fr", rpcid, null, null, null, [status_code]]`.
   Previously, `parse_frames` dropped the frame entirely because `item[2]` was null, causing callers to raise a misleading `WireFormatError: no generation record in the reply`.
   `batchexecute.parse_frames` now inspects the gRPC status block at slot `[5]` and raises typed exceptions (`AuthExpiredError` for 16, `InsufficientCreditsError` for 8, `FlowAppError` for 13) instead of masking server rejections behind envelope drift.

## Validation

- `tests/api/test_bootstrap_chooser.py`: 11/11 passed (covers `/about` click-through, account chooser auto-selection, and error typing).
- `tests/api/transports/test_batchexecute.py`: 12/12 passed (covers RPC error frames [13], [16], and `["er", ...]`).
- Full auth suite: 149/149 passed.
- Pre-push gates clean: `ruff check`, `ruff format --check`, `pyright` (0/0), `check_repo_hygiene.py`, `check_doc_links.py`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery when opening Google Flow lands on the `/about` page by following the creation flow and selecting the recorded account.
  - Improved authentication and navigation error messages for unresolved landing pages.
  - Added clearer handling of batchexecute errors, including authentication, access, credits, configuration, and server failures.
  - Preserved upstream error statuses instead of reporting misleading wire-format errors.


ffroliva:
Thanks for this — the `/about` stall is real and worth fixing, and the batchexecute error-envelope half is a clear improvement. A few things before it can land.

## CI is red for a substantive reason, not a flake

`test (3.13)` fails on `tests/e2e/test_landing_state_diagnosis_bdd.py::test_the_migrated_editor_is_answered_with_flows_public_landing_page`:

```
AssertionError: asserts an unmeasured cause ('not authenticated'):
  ... this session is not authenticated on this host — run 'gflow auth login' ...
```

That test exists specifically to stop this message from naming a cause. The comment this PR deletes from `_common.py` says why:

> Deliberately says WHAT arrived and stops. #756 measured the redirect and did not measure its cause — `gflow auth status` reports the session verified while this happens — so naming one here would just be a second confident wrong diagnosis.

So `"This session is not authenticated on this host"` is the one claim we have evidence against: when #756 reproduced this, the session **was** verified, 5/5 consecutive attempts on a healthy profile ([spike](docs/superpowers/spikes/2026-09-11-about-redirect-is-stable-for-an-account.md)). Please keep the recovery and leave the fallback message as it is on `develop`.

Same for the `retryable=False` justification comment — the flag is kept but the measurement behind it is deleted. That comment is the evidence; without it the next reader has no way to know the flag was measured rather than assumed.

## The premise is the interesting part — please capture it

The recovery is built on "`/about` means the session lacks `flow.google.com` cookies". That is exactly what #756 could not confirm. **If clicking "Create with Google Flow" genuinely recovers a session that `auth status` calls healthy, that is new evidence and I want it** — it would let us replace the deliberately vague message with a measured cause instead of a guessed one.

Please attach a capture (see `/gflow:spike` / `scripts/dev/har-spike/`): the landing hit, the click, and where it lands. Redact cookies and tokens. That turns this from a plausible fix into a settled one, and it is the difference between this PR narrowing the diagnosis and widening it.

## Locale-invariance

```python
button:has-text('Create with Google Flow')
```

`AGENTS.md` § Locale-Invariance Discipline rules out text-label selectors in `src/gflow_cli/api/transports/`. The existing `has-text` uses in `ui_automation.py` are product names (`Nano Banana 2`, `Imagen 4`) that Google does not translate, and each carries a comment saying so. A CTA label is translated, so this one will miss on any non-English account — which is a large share of the accounts hitting `/about`.

`button.flow-button.variant-primary` and `button.cta-button` are already in your cascade; please anchor on those plus a structural property (role, `href`, an icon ligature) and drop the text arm.

## Two copies of the same recovery

The `/about` + Create-button click is implemented twice:

- `src/gflow_cli/api/client.py::_handle_account_chooser`
- `src/gflow_cli/api/transports/_common.py::raise_if_known_landing`

Two copies of a selector cascade against a surface that moves will drift within a release. Please keep one — `_common.py` is the natural home, since it already owns every other landing-kind branch.

Also `except Exception: pass` in the `client.py` arm swallows the `wait_for_url` result silently; the `_common.py` arm logs it, which is the right shape.

## Suggested split

The batchexecute RPC error-envelope parsing is independent, well-scoped, and has no open questions — turning a null-payload frame into a typed error with the upstream status is a straight improvement. Please pull it into its own PR and I will merge it ahead of this one; no reason it should wait on the `/about` evidence.

ffroliva:
### Independent evidence that your premise is real

I asked you to capture the `/about` landing rather than assert its cause. Triaging the nightly canary today I reproduced it myself, so here is what I measured — it is not a request for more from you, it is me meeting you halfway.

`denon82`, 2026-09-20, driving `gflow project create`:

```
project.labs_route_refused   status=404                       ← correct, labs is retired
ui_driver.known_landing      kind=public
                             url=https://flow.google.com/about
                             requested=https://flow.google.com/
FlowAppError                 exit 31
```

**This is blocking a shipped feature.** v0.78.0's project-creation-on-flow.google.com cannot complete on that account, and it is 9 of the 13 standing canary failures (#888).

It is also **account-scoped**, which is the part your PR description does not yet account for. Same build, same minute:

| profile | `gflow project create` |
|---|---|
| `ffroliva` | created fine |
| `ci-probe` | created fine |
| `denon82` | exit 31 at `/about` |

So `/about` is real, reproducible and worth recovering from — your premise stands. What remains open is only whether **clicking "Create with Google Flow" actually recovers it**, which is the one thing neither of us has measured yet. If it does, you fix a live regression for a whole cohort and these 9 canary failures go away; that is a considerably bigger win than the PR currently claims.

My review points are unchanged and none of them is about the premise:

1. the fallback message must not assert `not authenticated` (#756 measured a *verified* session while this happens — and note my run above is on a healthy profile too, which is more evidence for your side of the message question, not less)
2. `has-text('Create with Google Flow')` will miss on non-English accounts
3. the recovery is implemented twice
4. the `retryable=False` measurement comment should stay

If you would rather not chase the capture, say so — with this reproducing on my side I can run the click myself and report back. I would just rather not duplicate your work if you are already on it.

ffroliva:
### Measured the Create click on the account that reproduces `/about`

I offered to run the click myself, and I have. Short version: **your premise that `/about` is an auth problem was right. The recovery isn't one**, because the click ends on a page that needs a password.

`denon82`, 2026-09-23, $0. Script: `scripts/dev/spike_about_cta_recovery.py`, with the readings pre-registered in its docstring.

| Step | Landed |
|---|---|
| gflow's own cookie check | `flow_session_cookie_present=True`, `expired=False`, SAPISID present |
| `flow.google.com/` ×2, no click | `/about` both times |
| click `button.flow-button.variant-primary` (index 0) | `accounts.google.com/v3/signin/confirmidentifier`, Google's **"Confirm it's you — sign in again to continue"** |
| `flow.google.com/` again | `/about` |
| after `gflow auth login` | same result — see below |
| after completing "Confirm it's you" **by hand** in real Chrome | `flow.google.com/`, the app, both times; `gflow project create` succeeds |

So for this account the cause is **a pending Google identity re-verification with a healthy cookie jar**. That is why `auth status` calls the session verified while Flow sends it to `/about` (#756).

What this means for the PR:

1. **The click cannot recover automatically.** It reaches a password page. On this account your code raises `AuthExpiredError`, which is the right exit class. The remediation text, though, has to tell the user to finish Google's identity check in Chrome, not to run `gflow auth login`.
2. **`gflow auth login` does not fix it on its own.** It reported `verified` in 0.4 s because the cookies are valid, and closed Chrome before Google could ask. Filed as #902; it is not yours to fix here.
3. **Selector:** `button.flow-button.variant-primary` matches **18** buttons on `/about`: the hero, 15 "Try in Google Flow" cards, "Apply here" and a "Get started" plan button. `.first` only picks the hero because of DOM order. `button.cta-button` matched exactly **1**. Please anchor on that and drop both the `variant-primary` and `has-text` arms. Google's re-auth page on this account rendered in pt-BR, so the text arm is a real risk, not a theoretical one.
4. The fallback `FlowAppError` still says "not authenticated". What was measured is narrower: *Google requires this account to re-verify its identity*. If you want to name a cause, name that one, and only on the branch where the click actually reached `confirmidentifier`.

My earlier points 3 and 4 (the duplicated recovery, the `retryable=False` comment) stand. The offer to merge the batchexecute half as its own PR also stands.

Full write-up: `docs/superpowers/spikes/2026-09-23-about-cta-leads-to-google-reauth.md` (PR #903).

ffroliva:
Heads-up from #904 (fixes #902): once it merges, `gflow auth login` stops on the `/about` identity re-check instead of reporting success. It keeps Chrome open and exits 12 `IdentityRecheckPendingError` if the user does not finish Google's check. So the `/about` remediation text here can point at `gflow auth login` and be correct. #904 does not touch `_common.py`, to avoid a conflict with this PR.

Issue #880, opened 2026-09-20
serve (MCP) parity: expose video extend / scene / chain, and a models-catalog tool


ai4U23:
### Context

We run Google Flow automation in production on top of `gflow serve` (0.79.0, migrated flow.google.com host, single ULTRA profile) — it drives our studio MCP server. Huge thanks for the migrated-host work in 0.73–0.79: the `--end-frame` port (#831), the picker-confirm/run-unique fixes (#793/#800), and the native daemon bearer auth made our deployment strictly simpler. `omni-flash` first+last-frame i2v via serve is live-verified on our side and has become our workhorse.

The project docs state CLI↔MCP parity as a contract for `serve`. Today several CLI video operations have no serve equivalent, so MCP clients either shell out to the CLI or lose the capability entirely:

### Gaps (CLI has it, serve doesn't)

1. **`gflow video extend`** — continue an existing clip by N seconds. This is the biggest one for pipeline work: extend rides the direct-wire `batchAsyncGenerateVideoExtendVideo` route and its own poller, so exposing it via serve looks mechanical (same args the CLI takes: `media_id`, `scene_id`, `position`, `prompt`, `aspect`, `model`).
2. **`gflow scene` (and `video chain`)** — compose ordered/trimmed clips into a Scene, and manifest-driven I2V chaining. Multi-clip assembly from an MCP client currently requires N round-trips of media ids through a side channel.
3. **A models-catalog serve tool** (`gflow_list_models` or similar). `gflow models --json` returns the live image/video catalog (names, aliases, ref caps, durations). MCP clients today hardcode catalogs — ours ships a dated snapshot because we can't read one. Note: on migrated accounts `projectInitialData` (labs tRPC) 401s for us (the #791 family), so a serve-side models tool would ideally read the CLI's static catalog table rather than the tRPC listing.

### What we're not asking for

Not asking for new Flow capabilities (no upscale/v2v asks — we know `UNSUPPORTED_ON_BATCH_API` and the post-migration API surface), only serve exposure of operations the CLI already drives today.

### Offer

Happy to contribute: we already maintain an MCP-over-HTTP client against `serve` with typed tool schemas, tests, and a bearer-auth pattern that matches your `_BearerAuthMiddleware`. If the shape of `gflow_extend_video` / `gflow_scene_*` tools is up for discussion, we'd love to align on signatures that work for both sides (we filed #791-adjacent and #793 contributions before).

[...]

Issue #877, opened 2026-09-18
gflow data download accepts an image media id, waits 45s, then blames the project, the trash, and your prompt


ffroliva:
## What happens

`gflow data download <image_media_id>` accepts the id, opens a browser, waits **45 s**, and then fails with exit 7 and a message that guesses two causes — both wrong here:

> migrated host: no signed media URL for `4814faa7-…` within 45s of opening its clip route. The media id may belong to another project, or the clip may have been moved to trash in Flow.
>
> -> Check request payload parameters or retry with a simpler prompt text. File a bug …

The image was generated on that same account **twelve minutes earlier**, is in that project, is not trashed, and its catalog row is intact (`gflow data media` shows it with `kind: image` and a `local_path`). None of the three suggestions applies.

## Why it fails

`migrated_recover.py:_await_signed_record` opens `/project/<pid>/edit/<media_id>` and waits for the `as29s` batchexecute reply that carries a freshly signed `flow-content.google/video/<workflow_id>` URL. That is a **video** mechanism. An image's route does not emit that record, so `found` stays empty until the deadline.

There is no kind check anywhere on the path — `services/media_recovery.py` reads the catalog row and already has `m.kind` (it uses it at `:69` only to build a disambiguation label), so the information needed to refuse fast is in hand and unused.

Related and already measured: `docs/superpowers/spikes/2026-09-17-stranded-clip-recovery.md` records that the listing's `lh3` poster token answers `=m18`/`=m22` with **valid but transcoded** MP4s — i.e. images and posters are served through a different CDN path than the signed video URL. 25 suffixes were measured and none returned an original. So this is a genuine mechanism difference, not a timing flake.

## Reproduction

```
gflow image t2i "a single matte grey ceramic sphere on white seamless" --profile <p> --project <id> --json
# note images[0].media_name, delete the written file
gflow data download <that media_name> --profile <p>
```

Observed 2026-09-18 on a profile Flow serves from `flow.google.com`. `migrated.recover_navigate` fires; no `as29s` record for the media id follows; exit 7 at 45 s.

## Impact

The command is described in terms that read as universal — *"Fetch an already-generated **asset** from Flow by its media ID"*, and the CHANGELOG entry says *"recover a billed **asset** whose download failed"*. The body text then switches to "the clip" and `gflow data list videos`, so the video-only scope is implied but never stated. A user with a stranded image follows the documented recovery path, waits 45 s, and is told to check a payload and simplify a prompt for a command that has neither.

[...]

PR #882, opened 2026-09-20
feat(extend): drive video extends on flow.google.com via batchexecute fZytfe (#639)

stgmt:
## Summary

Completes the video-extend port for migrated accounts (`flow.google.com`, issue #639), where the legacy aisandbox REST route returns HTTP 401:

1. **Wire submit (`fZytfe`)**:
   - Routes extend submissions to `POST /_/AiSandboxAngularFrontend/data/batchexecute?rpcids=fZytfe` with `source-path=/project/<pid>/scene/<scene_id>`.
   - Derives the 24fps frame window from the source clip's duration: `[duration*24-24+1, duration*24]` (e.g. `[73, 96]` for a 4s clip, `[169, 192]` for 8s).
   - Resolves and embeds `source_workflow_id` in the media reference slot (the migrated host accepts a media id but silently drops scheduling).

2. **Status polling & download (`as29s`)**:
   - Routes polling through `as29s` keyed on `workflow_id`.
   - Extracts the signed MP4 URL (`https://flow-content.google/video/<workflow_id>?Expires=...`) from the terminal record slot `(7, 0, 8)`.

3. **Live verification**:
   - Verified live on 2026-09-19: submission scheduled workflow `2bb68ea1`, reached status `DONE`, and downloaded a valid 7s MP4 (2 786 225 bytes, h264 720×1280).
   - Boundary frame comparison confirmed continuous camera movement matching manual UI extend.
   - Documented in `docs/superpowers/spikes/2026-09-19-migrated-video-extend-wire.md`.

4. **Docs & hygiene**:
   - Updated `KNOWN_ISSUES.md` (and website mirror), dropping `extend` from the unported list of #639.
   - Updated `docs/USAGE.md` (and website mirror) with migrated host notes.
   - Added `CHANGELOG.md` entry under `[Unreleased]`.

## Validation

- 48 offline unit/integration tests pass (`test_video_extend.py`, `test_client_extend.py`, `test_client_migrated_mint.py`).
- Pre-push gates clean: `ruff check`, `ruff format --check`, `pyright` (0/0), `check_repo_hygiene.py`, `check_doc_links.py`, `generate_website_docs.py --check`, `check_website_docs_pii.py`.

## Summary by CodeRabbit

* **New Features**
  * Added support for `gflow video extend` on migrated Flow accounts.
  * Added migrated-host scene creation for video extensions.
  * Video extensions automatically calculate the appropriate 24 fps frame range from the source clip.
  * Extension results now support status polling and signed video URL retrieval.

* **Improvements**
  * JSON output now provides consistent result metadata without plan details.
  * Single-segment results download directly; multi-segment results continue to be combined.

* **Documentation**
  * Updated usage guidance and known-issues documentation for migrated-host video extensions.


ffroliva:
This is strong work. The spike doc is exactly the standard this repo asks for and rarely gets — the `fZytfe` payload structure, the 24fps frame-window derivation, the `as29s` vs `jwpduf` polling difference, and a live run on 2026-09-19 with a boundary-frame comparison rather than "it looked fine". Extending on `flow.google.com` has been open since #639 and this is the first evidence-backed route to it.

Two things before it lands, one mechanical and one I would like you to check on live.

## 1. CI — stale test fake (mechanical)

`test (3.13)` fails 6 tests in `tests/cli/test_cli_video_extend.py`:

```
AttributeError: '_FakeClient' object has no attribute 'create_scene_for_extend'
```

`cli_video.py` now calls `client.create_scene_for_extend(...)` where it used to call `client.create_scene(...)`, but `_FakeClient` (`tests/cli/test_cli_video_extend.py:127`) still only defines `create_scene`. Adding the method to the fake should clear all six.

## 2. The project guard no longer runs on the migrated arm

This is the one I would like your eyes on, because you have the live account and I have not run it.

`cli_video.py` used to do this before creating the scene:

```python
workflow_id = video_extend.workflow_id_for_media(listing, media_id)
if not workflow_id:
    raise ConfigurationError(f"media {media_id} is not in project {project_id} ...")
```

`create_scene_for_extend` keeps that guard on the labs arm, but the migrated arm returns before reaching it and passes `media_id` straight through as a workflow id:

```python
if host == "migrated":
    return await self._create_scene_batchexecute(
        page=page, project_id=project_id, workflow_ids=[media_id]
    )

workflow_id = video_extend.workflow_id_for_media(listing, media_id)   # labs only now
```

Two reasons that worries me:

- **v0.79.0 just shipped #877/#878**, which makes `extend` refuse an *image* media id with a typed error instead of blaming the project, the trash and the prompt. On the migrated host — which is where most accounts now are — that guard is bypassed, so an image media id, a typo, or a clip from another project reaches the server unvalidated.
- **Your own spike names the failure mode**: for `fZytfe`, a media id in the workflow-id slot is *"accepted by the server but never scheduled (echo record, no job, credits burned)"*. `rqZuUc` is a different RPC, so that invariant may not carry over — but the shape of the risk is the same, and it is a paid path.

Could you run `gflow video extend` on the migrated host with (a) an image media id and (b) a media id from a different project, and report what comes back? If `rqZuUc` rejects them cleanly, say so in a comment and I am happy. If it echoes or burns credits, the guard needs to move above the host branch.

## Notes, not blockers

- The single-segment `download_video` bypass is a good call given `concatenate_scene` 401s on the migrated host. Worth a line in `KNOWN_ISSUES.md` or the changelog that `--output` skips concatenation at `count=1`, since the output path is now produced two different ways.
- #880 asks for `video extend` over `serve`/MCP. This PR is in front of that, so I will hold the MCP twin until this lands — and when it goes, it will want to be worker-enqueued rather than inline (#481), since a chain runs for minutes.

